PR #19, opened 2015-07-16
fixes #8 migrating section 5 notational conventions

### LINKED ISSUE #8: [Migrating Ecma-402 to HTML] Section 5 - Notational Conventions

caridy:
first draft of the migration...

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -33,7 +33,7 @@
 <h3 class="orange">3<sup>rd</sup> Edition DRAFT</h3>
 
 
-<div><h2>Table of Contents</h2><ol class="toc"><li><a href="#introduction"><span class="secnum"></span> Introduction</a></li><li><a href="#scope"><span class="secnum">1</span> Scope</a></li><li><a href="#conformance"><span class="secnum">2</span> Conformance</a></li><li><a href="#normative-references"><span class="secnum">3</span> Normative References</a></li><li><a href="#overview"><span class="secnum">4</span> Overview</a><ol class="toc"><li><a href="#sec-internationalization-localization-globalization"><span class="secnum">4.1</span> Internationalization, Localization, and Globalization</a></li><li><a href="#sec-api-overview"><span class="secnum">4.2</span> API Overview</a></li><li><a href="#sec-implementation-dependencies"><span class="secnum">4.3</span> Implementation Dependencies</a></li></ol></li></ol></div><emu-intro id="introduction">
+<div><h2>Table of Contents</h2><ol class="toc"><li><a href="#introduction"><span class="secnum"></span> Introduction</a></li><li><a href="#scope"><span class="secnum">1</span> Scope</a></li><li><a href="#conformance"><span class="secnum">2</span> Conformance</a></li><li><a href="#normative-references"><span class="secnum">3</span> Normative References</a></li><li><a href="#overview"><span class="secnum">4</span> Overview</a><ol class="toc"><li><a href="#sec-internationalization-localization-globalization"><span class="secnum">4.1</span> Internationalization, Localization, and Globalization</a></li><li><a href="#sec-api-overview"><span class="secnum">4.2</span> API Overview</a></li><li><a href="#sec-implementation-dependencies"><span class="secnum">4.3</span> Implementation Dependencies</a></li></ol></li><li><a href="#conventions"><span class="secnum">5</span> Notational Conventions</a><ol class="toc"><li><a href="#sec-well-known-intrinsic-objects"><span class="secnum">5.1</span> Well-Known Intrinsic Objects</a></li></ol></li></ol></div><emu-intro id="introduction">
   <h1><span class="secnum"></span>Introduction</h1>
   <p>
     The ECMAScript 2016 Internationalization API Specification (ECMA-402 3<sup>rd</sup> Edition) provides key language-sensitive functionality as a complement to the ECMAScript 2016 Language Specification (ECMA-262 7<sup>th</sup> Edition or successor). Its functionality has been selected from that of well-established internationalization APIs such as those of the Internationalization Components for Unicode (ICU) library, of the .NET framework, or of the Java platform.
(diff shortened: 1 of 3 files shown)


spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
+<emu-clause id="conventions">
+  <h1>Notational Conventions</h1>
+
+  <p>
+    This standard uses a subset of the notational conventions of the ECMAScript 2015 Language Specification (ECMA-262 6th Edition), as ES2015:
+  </p>
+
+  <ul>
+    <li>Object Internal Methods and Internal Slots, as described in ES2015, 6.1.7.2.</li>
```
caridy: @rwaldron we might need a way to link to ES2015 for all those sections that we reference here. I wonder if we have a resource list that can be generated but 262 that we can import to enable all those links :), or maybe that's just too much. /cc @bterlson

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      As described in the ECMAScript Language Specification, algorithms are used to precisely specify the required semantics of ECMAScript constructs, but are not intended to imply the use of any specific implementation technique. Internal slots are used to define the semantics of object values, but are not part of the API. They are defined purely for expository purposes. An implementation of the API must behave as if it produced and operated upon internal slots in the manner described here.
+    </p>
+  </emu-note>
+
+  <p>
+    As an extension to the Record Specification Type, the notation "[[name]]" denotes a field whose name is given by the variable name, which must have a String value. For example, if a variable s has the value "a", then [[s]] denotes the field [[a]].
+  </p>
+
+  <p>
+    For ECMAScript objects, this standard may use variable-named internal slots: The notation "[[name]]" denotes an internal slot whose name is given by the variable name, which must have a String value. For example, if a variable s has the value "a", then [[s]] denotes the [[a]] internal slot.
```
caridy: the original doc was referencing things like `[[<s>]]`, that sort of breaks this. I rename them to just `[[s]]`, I think going forward we might want to use that notation, not sure.

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+
+    <p>
+      The following table extends the Well-Known Intrinsic Objects table defined in ES2015, 6.1.7.4.
+    </p>
+
+    <h2 id="table-7">Table 7 — Well-known Intrinsic Objects (Extensions)</h2>
+    <table>
+      <thead>
+        <tr>
+          <th>Intrinsic Name</th>
```
caridy: In the loader spec we reference this as "Internal Slot" and entries in that column will be reference as `[[Something]]`. Can we agree on the right way to define this? /cc @bterlson

spec/source/conventions.html
rwaldron: >  I rename them to just [[s]], I think going forward we might want to use that notation, not sure.

No, that's wrong. The `[[<s>]]` notation has a specific meaning that will be completely broken if these are just changed to `[[s]]`. 

image

Is it breaking because the literal `[[<s>]]` is interpreted as though `<s>` is an HTML tag? If that's the case, then just write it out like `[[&lt;s&gt;]]` for these normative explanations. Then in the algorithms, escape them: `[[\<s\>]]`

spec/index.html
```diff
@@ -303,4 +303,112 @@ <h3 class="orange">3<sup>rd</sup> Edition DRAFT</h3>
...
+
+  <p>
+    This standard uses a subset of the notational conventions of the ECMAScript 2015 Language Specification (ECMA-262 6th Edition), as ES2015:
+  </p>
+
+  <ul>
+    <li>Object Internal Methods and Internal Slots, as described in ES2015, 6.1.7.2.</li>
+    <li>Algorithm conventions, including the use of abstract operations, as described in ES2015, 7.1, 7.2, 7.3.</li>
+    <li>Internal Slots, as described in ES2015, 9.1.</li>
+    <li>The List and Record Specification Type, as described in ES2015, 6.2.1.</li>
```
rwaldron: ES2015 => 2016

spec/index.html
```diff
@@ -303,4 +303,112 @@ <h3 class="orange">3<sup>rd</sup> Edition DRAFT</h3>
   </emu-clause>
 </emu-clause>
 
+<emu-clause id="conventions">
+  <h1><span class="secnum">5</span>Notational Conventions</h1>
+
+  <p>
+    This standard uses a subset of the notational conventions of the ECMAScript 2015 Language Specification (ECMA-262 6th Edition), as ES2015:
```
rwaldron: ECMA-262 6th Edition => ECMA-262 7th Edition

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
+<emu-clause id="conventions">
+  <h1>Notational Conventions</h1>
+
+  <p>
+    This standard uses a subset of the notational conventions of the ECMAScript 2015 Language Specification (ECMA-262 6th Edition), as ES2015:
```
rwaldron: ECMA-262 6th Edition => ECMA-262 7th Edition

rwaldron: Or maybe not? Maybe we should produce a version of the 2nd edition, tag it, and then move onto updating?

spec/source/conventions.html
rwaldron: An Intrinsic Name is _not_ the same as an Internal Slot.

rwaldron: http://www.ecma-international.org/ecma-262/6.0/#sec-well-known-intrinsic-objects

> Well-known intrinsics are built-in objects that are explicitly referenced by the algorithms of this specification and which usually have Realm specific identities. Unless otherwise specified each intrinsic object actually corresponds to a set of similar objects, one per Realm.

> Within this specification a reference such as %name% means the intrinsic object, associated with the current Realm, corresponding to the name. Determination of the current Realm and its intrinsics is described in 8.3. The well-known intrinsics are listed in Table 7.

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+  </ul>
+
+  <emu-note>
+    <p>
+      As described in the ECMAScript Language Specification, algorithms are used to precisely specify the required semantics of ECMAScript constructs, but are not intended to imply the use of any specific implementation technique. Internal slots are used to define the semantics of object values, but are not part of the API. They are defined purely for expository purposes. An implementation of the API must behave as if it produced and operated upon internal slots in the manner described here.
+    </p>
+  </emu-note>
+
+  <p>
+    As an extension to the Record Specification Type, the notation "[[name]]" denotes a field whose name is given by the variable name, which must have a String value. For example, if a variable s has the value "a", then [[s]] denotes the field [[a]].
```
rwaldron: - "[[name]]" => "[[\<_name\>_]]"
- "the variable name" => "the variable _name_"
- "if a variable s has the value "a", then [[s]]" => "if a variable _s_ has the value "a", then [[\<_s\>_]]"

Variables are always written in italics

spec/source/conventions.html
rwaldron: The changes to formatting need to be made in this paragraph

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+        <tr>
+          <th>Intrinsic Name</th>
+          <th>Global Name</th>
+          <th>ECMAScript Language Association</th>
+        </tr>
+      </thead>
+      <tr>
+        <td>%Date_now%</td>
+        <td>"Date.now"</td>
+        <td>The initial value of the "now" data property of the intrinsic %Date% (ES2015, 20.3.3.1)</td>
```
rwaldron: - "now" => monospaced

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%DateTimeFormatPrototype%</td>
+        <td>"Intl.DateTimeFormat.prototype"</td>
+        <td>The initial value of the "prototype" data property of the intrinsic %DateTimeFormat% (11.2.1).</td>
+      </tr>
+      <tr>
+        <td>%StringProto_split%</td>
+        <td>"String.prototype.split"</td>
+        <td>The initial value of the "split" data property of the intrinsic %StringPrototype% (ES2015, 21.1.3.17)</td>
```
rwaldron: - "split" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%Collator%</td>
+        <td>"Intl.Collator"</td>
+        <td>The Intl.Collator constructor (10.1)</td>
+      </tr>
+      <tr>
+        <td>%CollatorPrototype%</td>
+        <td>"Intl.Collator.prototype"</td>
+        <td>The initial value of the "prototype" data property of the intrinsic %Collator% (10.2.1).</td>
```
rwaldron: - "prototype" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%NumberFormat%</td>
+        <td>"Intl.NumberFormat"</td>
+        <td>The Intl.NumberFormat constructor (11.1)</td>
+      </tr>
+      <tr>
+        <td>%NumberFormatPrototype%</td>
+        <td>"Intl.NumberFormat.prototype"</td>
+        <td>The initial value of the "prototype" data property of the intrinsic %NumberFormat% (11.2.1).</td>
```
rwaldron: - "prototype" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%DateTimeFormat%</td>
+        <td>"Intl.DateTimeFormat"</td>
+        <td>The Intl.DateTimeFormat constructor (12.1).</td>
+      </tr>
+      <tr>
+        <td>%DateTimeFormatPrototype%</td>
+        <td>"Intl.DateTimeFormat.prototype"</td>
+        <td>The initial value of the "prototype" data property of the intrinsic %DateTimeFormat% (11.2.1).</td>
```
rwaldron: - "prototype" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%StringProto_split%</td>
+        <td>"String.prototype.split"</td>
+        <td>The initial value of the "split" data property of the intrinsic %StringPrototype% (ES2015, 21.1.3.17)</td>
+      </tr>
+      <tr>
+        <td>%StringProto_indexOf%</td>
+        <td>"String.prototype.indexOf"</td>
+        <td>The initial value of the "indexOf" data property of the intrinsic %StringPrototype% (ES2015, 21.1.3.8)</td>
```
rwaldron: - "indexOf" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%StringProto_includes%</td>
+        <td>"String.prototype.includes"</td>
+        <td>The initial value of the "includes" data property of the intrinsic %StringPrototype% (ES2015, 21.1.3.7)</td>
```
rwaldron: - "includes" => code/monospaced font

spec/source/conventions.html
```diff
@@ -0,0 +1,107 @@
...
+      </tr>
+      <tr>
+        <td>%ArrayProto_indexOf%</td>
+        <td>"Array.prototype.indexOf"</td>
+        <td>The initial value of the "indexOf" data property of the intrinsic %StringPrototype% (ES2015, 22.1.3.11)</td>
```
rwaldron: - "indexOf" => code/monospaced font

spec/source/conventions.html
caridy: I think so!

spec/source/conventions.html
caridy: `[[<something>]]` breaks ecmarkup for some reason, I think it is interpreting it as a new tag. we might use `&entities;` instead to support that though.

spec/source/conventions.html
rwaldron: Cool, then we'll have to undo the ones I've already changed, but that should be trivial

spec/source/conventions.html
rwaldron: Are you thinking something like my explanation here: https://github.com/tc39/ecma402/pull/19/files#r35254684 ?

caridy: correct. I will make the changes then.

spec/source/conventions.html
caridy: ok.

caridy:
Ok @rwaldron, I took some liberties on the first past to improve some formatting, but I have rolled them back to match what we have today. We can improve it after we finish this exercise anyway.

PR #41, opened 2015-09-29
Migrate to <emu-import href="...">

rwaldron:
@caridy newer versions of ecmarkup have replaced `<link ...>` with `<emu-import ...>`. This patch updates the spec source.


caridy:
+1

caridy:
feel free to merge it, or I can take care of it later today along with few other updates I have line up to align better with the new version of ecmarkup.

Issue #48, opened 2015-10-27
Tracking the new algorithm convention for return-if-abrupt


rwaldron:
https://github.com/tc39/ecma262/pull/129

bterlson:
I have a node script to do the conversion if you're interested.

rwaldron:
Yep, I think @caridy and I would prefer mechanically updating vs manual

caridy:
:) will ping on thru to get it.

bterlson:
It's still being edited now, so I'll give you the final version once tc39/ecma262#129 is accepted.

caridy:
fixed in various commits, we no longer use return-if-abrupt form in 402, we use either `?` or `!`

PR #50, opened 2015-11-05
editorial fixes and alignment with ES2016 conventions

caridy:
## Editorial fixes

- then and else corrections
- algos with numeric index in the steps are now set to 1.
- few missing cross-links
- proper use of `*this* value`

## Aligning with 262:

- ReturnIfAbrupt -> ?
- inline styles and conventions

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -49,46 +49,39 @@
 
       <emu-alg>
         1. If collator has an [[initializedIntlObject]] internal slot with value *true*, throw a *TypeError* exception.
-        2. Set collator.[[initializedIntlObject]] to *true*.
-        3. Let _requestedLocales_ be CanonicalizeLocaleList(_locales_).
-        4. ReturnIfAbrupt(_requestedLocales_).
-        5. If _options_ is *undefined*, then
+        1. Set collator.[[initializedIntlObject]] to *true*.
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
+        1. If _options_ is *undefined*, then
           1. Let _options_ be ObjectCreate(*%ObjectPrototype%*).
-        6. Else
-          1. Let _options_ be ToObject(_options_).
-          1. ReturnIfAbrupt(_options_).
-        7. Let _u_ be GetOption(_options_, *"usage"*, *"string"*, «*"sort"*, *"search"*», *"sort"*).
-        8. ReturnIfAbrupt(_u_).
-        9. Set _collator_.[[usage]] to _u_.
-        10. If _u_ is *"sort"*, then
-          1. Let _localeData_ be the value of *%Collator%*.[[sortLocaleData]];
-        11. Else
+        1. Else
+          1. Let _options_ be ? ToObject(_options_).
+        1. Let _u_ be ? GetOption(_options_, *"usage"*, *"string"*, «*"sort"*, *"search"*», *"sort"*).
+        1. Set _collator_.[[usage]] to _u_.
+        1. If _u_ is *"sort"*, then
+          1. Let _localeData_ be the value of *%Collator%*.[[sortLocaleData]].
+        1. Else,
           1. Let _localeData_ be the value of *%Collator%*.[[searchLocaleData]].
-        12. Let _opt_ be a new Record.
-        13. Let _matcher_ be GetOption(_options_, *"localeMatcher"*, *"string"*, «*"lookup"*, *"best fit"*», *"best fit"*).
-        14. ReturnIfAbrupt(_matcher_).
-        15. Set _opt_.[[localeMatcher]] to _matcher_.
-        16. For each row in <emu-xref href="#table-1">Table 1</emu-xref>, except the header row, do:
+        1. Let _opt_ be a new Record.
+        1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, «*"lookup"*, *"best fit"*», *"best fit"*).
+        1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. For each row in <emu-xref href="#table-1">Table 1</emu-xref>, except the header row, do:
           1. Let _key_ be the name given in the Key column of the row.
           1. Let _prop_ be the name given in the Property column of the row.
           1. Let _type_ be the string given in the Type column of the row.
           1. Let _list_ be a List containing the Strings given in the Values column of the row, or *undefined* if no strings are given.
-          1. Let _value_ be GetOption(_options_, _prop_, _type_, _list_, *undefined*).
-          1. ReturnIfAbrupt(_value_).
+          1. Let _value_ be ? GetOption(_options_, _prop_, _type_, _list_, *undefined*).
           1. If the string given in the Type column of the row is *"boolean"* and value is not *undefined*, then
-            1. Let _value_ be ToString(_value_).
-            1. ReturnIfAbrupt(_value_).
+            1. Let _value_ be ? ToString(_value_).
           1. Set _opt_.[[&lt;_key_&gt;]] to _value_.
-        17. Let _relevantExtensionKeys_ be the value of *%Collator%*.[[relevantExtensionKeys]].
-        18. Let _r_ be ResolveLocale(*%Collator%*.[[availableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
-        19. Set _collator_.[[locale]] to the value of _r_.[[locale]].
-        20. Let _k_ be 0.
-        21. Let _lenValue_ be Get(_relevantExtensionKeys_, *"length"*).
-        22. Let _len_ be ToLength(_lenValue_).
-        23. Repeat while _k_ < _len_:
+        1. Let _relevantExtensionKeys_ be the value of *%Collator%*.[[relevantExtensionKeys]].
+        1. Let _r_ be ResolveLocale(*%Collator%*.[[availableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
+        1. Set _collator_.[[locale]] to the value of _r_.[[locale]].
+        1. Let _k_ be 0.
+        1. Let _lenValue_ be Get(_relevantExtensionKeys_, `"length"`).
+        1. Let _len_ be ToLength(_lenValue_).
+        1. Repeat while _k_ < _len_:
           1. Let _Pk_ be ToString(_k_).
-          1. Let _key_ be Get(_relevantExtensionKeys_, _Pk_).
-          1. ReturnIfAbrupt(_key_).
+          1. Let _key_ be ? Get(_relevantExtensionKeys_, _Pk_).
           1. If _key_ is *"co"*, then
             1. Let _property_ be *"collation"*.
             1. Let _value_ be the value of _r_.[[co]].
@@ -99,21 +92,20 @@
             1. If the name given in the Type column of the row is *"boolean"*, let _value_ be the result of comparing value with *"true"*.
           1. Set _collator_.[[&lt;_property_&gt;]] to _value_.
           1. Increase _k_ by 1.
-        24. Let _s_ be GetOption(_options_, *"sensitivity"*, *"string"*, «*"base"*, *"accent"*, *"case"*, *"variant"*», *undefined*).
-        25. ReturnIfAbrupt(_s_).
-        26. If _s_ is *undefined*, then
-          1. If _u_ is *"sort"*, then let _s_ be *"variant"*.
-          1. Else
+        1. Let _s_ be ? GetOption(_options_, *"sensitivity"*, *"string"*, «*"base"*, *"accent"*, *"case"*, *"variant"*», *undefined*).
+        1. If _s_ is *undefined*, then
+          1. If _u_ is *"sort"*, then
+            1. Let _s_ be *"variant"*.
+          1. Else,
             1. Let _dataLocale_ be the value of _r_.[[dataLocale]].
             1. Let _dataLocaleData_ be Get(_localeData_, _dataLocale_).
             1. Let _s_ be Get(_dataLocaleData_, *"sensitivity"*).
-        27. Set _collator_.[[sensitivity]] to _s_.
-        28. Let _ip_ be GetOption(_options_, *"ignorePunctuation"*, *"boolean"*, *undefined*, *false*).
-        29. ReturnIfAbrupt(_ip_).
-        30. Set _collator_.[[ignorePunctuation]] to _ip_.
-        31. Set _collator_.[[boundCompare]] to *undefined*.
-        32. Set _collator_.[[initializedCollator]] to *true*.
-        33. Return _collator_.
+        1. Set _collator_.[[sensitivity]] to _s_.
+        1. Let _ip_ be ? GetOption(_options_, *"ignorePunctuation"*, *"boolean"*, *undefined*, *false*).
+        1. Set _collator_.[[ignorePunctuation]] to _ip_.
+        1. Set _collator_.[[boundCompare]] to *undefined*.
+        1. Set _collator_.[[initializedCollator]] to *true*.
+        1. Return _collator_.
       </emu-alg>
 
     </emu-clause>
@@ -127,9 +119,8 @@
 
       <emu-alg>
         1. If _NewTarget_ is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be _NewTarget_.
-        2. Let _collator_ be OrdinaryCreateFromConstructor(_newTarget_, *%CollatorPrototype%*).
-        3. ReturnIfAbrupt(_collator_).
-        4. Return InitializeCollator(_collator_, _locales_, _options_).
+        1. Let _collator_ be ? OrdinaryCreateFromConstructor(_newTarget_, *%CollatorPrototype%*).
+        1. Return InitializeCollator(_collator_, _locales_, _options_).
       </emu-alg>
     </emu-clause>
   </emu-clause>
@@ -138,7 +129,7 @@
     <h1>Properties of the Intl.Collator Constructor</h1>
 
     <p>
-      Besides the internal slots and the *length* property (whose value is 0), the Intl.Collator constructor has the following properties:
+      Besides the internal slots and the `length` property (whose value is 0), the Intl.Collator constructor has the following properties:
     </p>
 
     <emu-clause id="sec-Intl.Collator.prototype">
@@ -161,13 +152,12 @@
       </p>
 
       <emu-alg>
-        1. Let _requestedLocales_ be CanonicalizeLocaleList(_locales_).
-        2. ReturnIfAbrupt(_requestedLocales_).
-        3. Return SupportedLocales(*%Collator%*.[[availableLocales]], _requestedLocales_, _options_).
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
+        1. Return SupportedLocales(*%Collator%*.[[availableLocales]], _requestedLocales_, _options_).
       </emu-alg>
 
       <p>
-        The value of the *length* property of the *supportedLocalesOf* method is 1.
+        The value of the `length` property of the *supportedLocalesOf* method is 1.
       </p>
     </emu-clause>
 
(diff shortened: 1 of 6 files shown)


caridy:
fixes https://github.com/tc39/ecma402/issues/48

spec/collator.html
```diff
@@ -49,46 +49,39 @@
...
-        10. If _u_ is *"sort"*, then
-          1. Let _localeData_ be the value of *%Collator%*.[[sortLocaleData]];
-        11. Else
+        1. Else
+          1. Let _options_ be ? ToObject(_options_).
+        1. Let _u_ be ? GetOption(_options_, *"usage"*, *"string"*, «*"sort"*, *"search"*», *"sort"*).
+        1. Set _collator_.[[usage]] to _u_.
+        1. If _u_ is *"sort"*, then
+          1. Let _localeData_ be the value of *%Collator%*.[[sortLocaleData]].
+        1. Else,
```
zbraniecki: `Else,` or `Else`? Lines 62 and 56 are inconsistent

spec/collator.html
```diff
@@ -49,46 +49,39 @@
...
-        19. Set _collator_.[[locale]] to the value of _r_.[[locale]].
-        20. Let _k_ be 0.
-        21. Let _lenValue_ be Get(_relevantExtensionKeys_, *"length"*).
-        22. Let _len_ be ToLength(_lenValue_).
-        23. Repeat while _k_ < _len_:
+        1. Let _relevantExtensionKeys_ be the value of *%Collator%*.[[relevantExtensionKeys]].
+        1. Let _r_ be ResolveLocale(*%Collator%*.[[availableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
+        1. Set _collator_.[[locale]] to the value of _r_.[[locale]].
+        1. Let _k_ be 0.
+        1. Let _lenValue_ be Get(_relevantExtensionKeys_, `"length"`).
```
zbraniecki: Why do you use '\`' instead of '*' here, but not for other strings? (line 65)

zbraniecki:
r+

spec/collator.html
rwaldron: Right. It looks like @bterlson is using these two conventions: 

- comma: 

  ```
  1. Else,
    1. When there is
    2. More than one line
  ```
- no comma:

  ```
  1. Else if there is only one line
  ```

spec/collator.html
```diff
@@ -49,46 +49,39 @@
 
       <emu-alg>
         1. If collator has an [[initializedIntlObject]] internal slot with value *true*, throw a *TypeError* exception.
```
rwaldron: Unrelated to this PR, "collator" should be in italics (throughout this entire algorithm)

spec/collator.html
caridy: yeah, I plan to do another pass on the `<code>` representation in all algos to match 262, specifically, any string token should be described as:
```
`"something"`
```

more details here: http://bterlson.github.io/ecmarkup/#editorial-conventions

@bterlson is that assessment correct? I don't see any "double-quoted string" in algos in 262 without the wrapping code call. should we update the editorial conventions?

spec/collator.html
caridy: good catch, will fix it in another commit. (I suspect that there are few more like this one, and I would love to have a emu validator to catch them all). @bterlson can we get a linter? jajajaja

spec/collator.html
rwaldron: Likely out of habit.

spec/collator.html
```diff
@@ -49,46 +49,39 @@
...
           1. Let _type_ be the string given in the Type column of the row.
           1. Let _list_ be a List containing the Strings given in the Values column of the row, or *undefined* if no strings are given.
-          1. Let _value_ be GetOption(_options_, _prop_, _type_, _list_, *undefined*).
-          1. ReturnIfAbrupt(_value_).
+          1. Let _value_ be ? GetOption(_options_, _prop_, _type_, _list_, *undefined*).
           1. If the string given in the Type column of the row is *"boolean"* and value is not *undefined*, then
-            1. Let _value_ be ToString(_value_).
-            1. ReturnIfAbrupt(_value_).
+            1. Let _value_ be ? ToString(_value_).
           1. Set _opt_.[[&lt;_key_&gt;]] to _value_.
```
rwaldron: @caridy Should we pick some weird unicode symbol to replace this mechanism in the 3rd edition? (Also unrelated to this PR)

spec/collator.html
caridy: yeah, that's what I'm following.

spec/collator.html
```diff
@@ -138,7 +129,7 @@
     <h1>Properties of the Intl.Collator Constructor</h1>
 
     <p>
-      Besides the internal slots and the *length* property (whose value is 0), the Intl.Collator constructor has the following properties:
+      Besides the internal slots and the `length` property (whose value is 0), the Intl.Collator constructor has the following properties:
```
rwaldron: Intentional?

spec/collator.html
```diff
@@ -161,13 +152,12 @@
...
-        1. Let _requestedLocales_ be CanonicalizeLocaleList(_locales_).
-        2. ReturnIfAbrupt(_requestedLocales_).
-        3. Return SupportedLocales(*%Collator%*.[[availableLocales]], _requestedLocales_, _options_).
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
+        1. Return SupportedLocales(*%Collator%*.[[availableLocales]], _requestedLocales_, _options_).
       </emu-alg>
 
       <p>
-        The value of the *length* property of the *supportedLocalesOf* method is 1.
+        The value of the `length` property of the *supportedLocalesOf* method is 1.
```
rwaldron: ?

spec/collator.html
```diff
@@ -187,7 +177,7 @@
       </p>
 
       <ul>
-        <li>The first element of [[sortLocaleData]][locale].co and [[searchLocaleData]][locale].co must be null for all locale values.</li>
+        <li>The first element of [[sortLocaleData]][locale].co and [[searchLocaleData]][locale].co must be *null* for all locale values.</li>
```
rwaldron: The word null must be formatted bold, not as code (IIRC, "*" make code formatting?)

spec/collator.html
```diff
@@ -237,13 +227,15 @@
...
 
       <emu-alg>
-        1. Let _collator_ be this Collator object.
-        2. If _collator_.[[boundCompare]] is *undefined*, then
-          1. Let _F_ be a new built-in function object as defined in 11.3.4.
-          1. The value of _F_'s length property is 2.
-          1. Let _bc_ be BoundFunctionCreate(_F_, «*this* value»).
+        1. Let _collator_ be *this* value.
+        1. If Type(_collator_) is not Object, throw a *TypeError* exception.
+        1. If _collator_ does not have a [[boundCompare]] internal slot, throw a *TypeError* exception.
```
rwaldron: I like that this is more explicit now

spec/collator.html
```diff
@@ -265,14 +257,12 @@
...
         1. Let _collator_ be the *this* value.
-        2. Assert: Type(collator) is Object and collator has an [[initializedCollator]] internal slot whose value is *true*.
-        3. If _x_ is not provided, let _x_ be undefined.
-        4. If _y_ is not provided, let _y_ be undefined.
-        5. Let _X_ be ToString(_x_).
-        6. ReturnIfAbrupt(_X_).
-        7. Let _Y_ be ToString(_y_).
-        8. ReturnIfAbrupt(_Y_).
-        9. Return CompareStrings(_collator_, _X_, _Y_).
+        1. Assert: Type(collator) is Object and collator has an [[initializedCollator]] internal slot whose value is *true*.
```
rwaldron: Unrelated to this PR: `collator` => `_collator_`

spec/datetimeformat.html
```diff
@@ -102,16 +97,14 @@
         1. Let _opt_ be a new Record.
         1. For each row of <emu-xref href="#table-3">Table 3</emu-xref>, except the header row, do:
           1. Let _prop_ be the name given in the Property column of the row.
-          1. Let _value_ be GetOption(_options_, _prop_, *"string"*, «the strings given in the Values column of the row», *undefined*).
-          1. ReturnIfAbrupt(_value_).
+          1. Let _value_ be ? GetOption(_options_, _prop_, *"string"*, «the strings given in the Values column of the row», *undefined*).
```
rwaldron: The word undefined must be formatted bold, not as code

spec/datetimeformat.html
```diff
@@ -119,8 +112,7 @@
           1. Let _p_ be Get(_bestFormat_, _prop_).
           1. If _p_ not *undefined*, then
             1. Set _dateTimeFormat_.[[<_prop_>]] to _p_.
-        1. Let _hr12_ be GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
-        1. ReturnIfAbrupt(_hr12_).
+        1. Let _hr12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
```
rwaldron: Same as above, re: undefined

spec/datetimeformat.html
```diff
@@ -144,29 +136,26 @@
       </p>
 
       <emu-alg aoid="ToDateTimeOptions">
-        1. If _options_ is *undefined*, then let _options_ be *null*, else let _options_ be ToObject(_options_).
-        1. ReturnIfAbrupt(_options_).
+        1. If _options_ is *undefined*, let _options_ be *null*; otherwise let _options_ be ? ToObject(_options_).
```
rwaldron: Same as above, re: undefined

spec/datetimeformat.html
```diff
@@ -144,29 +136,26 @@
...
         1. Let _options_ be ObjectCreate(_options_).
         1. Let _needDefaults_ be *true*.
         1. If _required_ is *"date"* or *"any"*,
           1. For each of the property names *"weekday"*, *"year"*, *"month"*, *"day"*:
             1. Let _prop_ be the property name.
-            1. Let _value_ be Get(_options_, _prop_).
-            1. ReturnIfAbrupt(_value_).
-            1. If _value_ is not *undefined*, then let _needDefaults_ be *false*.
+            1. Let _value_ be ? Get(_options_, _prop_).
+            1. If _value_ is not *undefined*, let _needDefaults_ be *false*.
```
rwaldron: Same as above, re: undefined

spec/datetimeformat.html
```diff
@@ -144,29 +136,26 @@
...
         1. If _needDefaults_ is *true* and _defaults_ is either *"date"* or *"all"*, then
           1. For each of the property names *"year"*, *"month"*, *"day"*:
-            1. Let _status_ be CreateDatePropertyOrThrow(_options_, _prop_, *"numeric"*).
+            1. Let _status_ be the result of calling CreateDatePropertyOrThrow(_options_, _prop_, *"numeric"*).
```
rwaldron: This language pattern exists nowhere else in this spec and only twice in Ecma-262, please revert

spec/datetimeformat.html
```diff
@@ -144,29 +136,26 @@
...
             1. ReturnIfAbrupt(_status_).
         1. If _needDefaults_ is *true* and _defaults_ is either *"time"* or *"all"*, then
           1. For each of the property names *"hour"*, *"minute"*, *"second"*:
-            1. Let _status_ be CreateDatePropertyOrThrow(_options_, _prop_, *"numeric"*).
+            1. Let _status_ be the result of calling CreateDatePropertyOrThrow(_options_, _prop_, *"numeric"*).
```
rwaldron: Same

spec/datetimeformat.html
```diff
@@ -236,7 +224,7 @@
     <h1>Properties of the Intl.DateTimeFormat Constructor</h1>
 
     <p>
-      Besides the internal slots and the *length* property (whose value is 0), the Intl.DateTimeFormat constructor has the following properties:
+      Besides the internal slots and the `length` property (whose value is 0), the Intl.DateTimeFormat constructor has the following properties:
```
rwaldron: Revert

spec/datetimeformat.html
```diff
@@ -259,13 +247,12 @@
...
         1. Let _availableLocales_ be the value of *%DateTimeFormat%*.[[availableLocales]].
-        1. Let _requestedLocales_ be CanonicalizeLocaleList(_locales_).
-        1. ReturnIfAbrupt(_requestedLocales_).
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Return _supportedLocales_ be SupportedLocales(_availableLocales_, _requestedLocales_, _options_).
       </emu-alg>
 
       <p>
-        The value of the *length* property of the *supportedLocalesOf* method is 1.
+        The value of the `length` property of the *supportedLocalesOf* method is 1.
```
rwaldron: Revert

spec/datetimeformat.html
```diff
@@ -363,12 +350,13 @@
...
-        1. ReturnIfAbrupt(_dtf_).
-        1. If the [[boundFormat]] internal slot of this DateTimeFormat object is *undefined*, then
+        1. Let _dtf_ be *this* value.
+        1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
+        1. If _dtf_ does not have a [[boundFormat]] internal slot, throw a *TypeError* exception.
+        1. If the [[boundFormat]] internal slot of _dtf_ is *undefined*, then
           1. Let _F_ be a new built-in function object as defined in <emu-xref href="#sec-datetime-format-functions"></emu-xref>.
-          1. The value of _F_’s *length* property is 1.
-          1. Let _bf_ be BoundFunctionCreate(F, «*this* value»).
+          1. The value of _F_’s `length` property is 1.
```
rwaldron: This change keeps appearing? I looked through the ecma-262 commits and don't see a new syntax for backticks.

spec/datetimeformat.html
```diff
@@ -428,13 +413,13 @@
               1. Let _fv_ be FormatNumber(_nf_, _v_).
             1. Else if f is *"2-digit"*, then
               1. Let _fv_ be FormatNumber(_nf2_, _v_).
-              1. If the length of _fv_ is greater than 2, let _fv_ be the substring of _fv_ containing the last two characters.
+              1. If the `length` property of _fv_ is greater than 2, let _fv_ be the substring of _fv_ containing the last two characters.
```
rwaldron: Should be *

spec/locale-sensitive-functions.html
```diff
@@ -25,17 +25,14 @@
...
-        3. ReturnIfAbrupt(_S_).
-        4. Let _That_ be ToString(_that_).
-        5. ReturnIfAbrupt(_That_).
-        6. Let _collator_ be Construct(*%Collator%*, «*locales*, *options*»).
-        7. ReturnIfAbrupt(_collator_).
-        8. Return CompareStrings(_collator_, _S_, _That_)
+        1. Let _S_ be ? ToString(_O_).
+        1. Let _That_ be ? ToString(_that_).
+        1. Let _collator_ be ? Construct(*%Collator%*, «*locales*, *options*»).
+        1. Return CompareStrings(_collator_, _S_, _That_)
```
rwaldron: Soooo much nicer

spec/locale-sensitive-functions.html
```diff
@@ -87,7 +82,7 @@
       </p>
 
       <p>
-        The value of the *length* property of the *toLocaleLowerCase* method is 0.
+        The value of the `length` property of the *toLocaleLowerCase* method is 0.
```
rwaldron: Revert

spec/locale-sensitive-functions.html
```diff
@@ -116,7 +111,7 @@
       </p>
 
       <p>
-        The value of the *length* property of the *toLocaleUpperCase* method is 0.
+        The value of the `length` property of the *toLocaleUpperCase* method is 0.
```
rwaldron: Revert

spec/locale-sensitive-functions.html
```diff
@@ -267,32 +251,26 @@
...
-            1. ReturnIfAbrupt(_R_).
+          1. Else,
+            1. Let _R_ be ? ToString(Invoke(_nextElement_, *"toLocaleString"*, «*locales*, *options*»)).
           1. Let _R_ be a String value produced by concatenating _S_ and _R_.
           1. Increase _k_ by 1.
-        13. Return _R_.
+        1. Return _R_.
       </emu-alg>
 
       <emu-note>
```
rwaldron: Just revert any incorrectly changed formatting (so I don't have to mark them all)

rwaldron:
Ping me when formatting issues are fixed—thanks!

spec/collator.html
bterlson: I personally think *"something"* makes the most sense but the current convention is indeed `"something"`.

spec/datetimeformat.html
caridy:
```
1. Let _f_ be Something().
```

it is a call to an abstract operation with an implicit step that sets `f` to the Completion Record's `[[Value]]`, while:

```
1. Let _f_ be the result of calling | evaluating Something().
```

is a call to an abstract operation that return Completion Records explicitly.

I see a lot of spec text that are using the `Let _status_ ... ` vs `Let _f_ ... ` indistinctly, because it was very hard to deal with the value vs status of the operation to abrupt of use the value, but now that we have ?, it is really simply. I think we can now enforce that when someone is using `ReturnIfAbrupt()` explicitly, it will have to be on the result of calling/evaluating step.

That's my interpretation of this topic:
http://tc39.github.io/ecma262/#sec-implicit-completion-values

/cc @bterlson can you chime in here?

bterlson: There really isn't a difference between `let _f_ be Call()` and `Let _f_ be the result of calling`. The latter is an outmoded more verbose convention that can be replaced with the former any time its spotted. You are correct that there's an oddity with completion records - the oddity is that sometimes algorithms treat a completion record as if it were a value. This is ok per 6.2.2.2 last paragraph.

I completely agree that depending on the 6.2.2.2 implicitness is bad and now that it's easy to be explicit, we should be. Proposals for making this easier are welcome! (Eg. I had originally proposed ! as a short-hand for asserting an operation is not an abrupt completion and replacing the completion record with its value, for example).

spec/collator.html
caridy: in 262, all occurrences of null are using `*null*`. /cc @bterlson

spec/datetimeformat.html
caridy: in 262, it is `*undefined*` everywhere. after a discussion with @bterlson, his recommendation for code is:

"the thing inside ` is ecmascript source text, and not a literal, e.g.: `Function.prototype`".

spec/datetimeformat.html
caridy: Ok, I think I can simplify these lines for now.

As a side note for @bterlson, for abtract operations that return nothing like `CreateDataPropertyOrThrow`, should we just do:

```
1. ? CreateDataPropertyOrThrow(...).
```

for now I will roll with:

```
1. Let _status_ be ? CreateDataPropertyOrThrow(...).
```

spec/collator.html
bterlson: \* is an ecmascript value (renders in bold). \` is for code :)

caridy:
@rwaldron ping!

spec/collator.html
rwaldron: The inconsistency is on line 56: 

```
56 +        1. Else
57 +          1. Let _options_ be ? ToObject(_options_).
```

spec/collator.html
rwaldron: I have no explanation for why I ever thought that was the other way around.

spec/datetimeformat.html
rwaldron: Yes, I was confused looking at the source here because there are probably 100's of places where literal strings or intrinsics should've been formatted as code, but have only `*"foo"*` or `*foo*` and I guess I thought those were correct, but was obviously misunderstanding what I was reading. 

Compare: 



to:

rwaldron: Looks like intrinsics should not have any special formatting (according to 262)... but they end up looking super crappy.

spec/collator.html
caridy: ah!, will fix it.

caridy: fixed.

spec/collator.html
caridy: opened an issue to discuss the details: https://github.com/tc39/ecma402/issues/56

spec/collator.html
caridy: good

caridy:
I will merge this, and we will continue iterating on other editorial fixes very soon.

Issue #30, opened 2015-09-16
Extend Formatters to allow for token formatting


zbraniecki:
UPDATE:
- [Proposal](https://github.com/zbraniecki/proposal-intl-formatToParts)
- [Latest slides - Nov 2015](https://github.com/tc39/ecma402/files/41759/formatToParts.pdf)

---

DateTimeFormat and NumberFormat are design to provide an opaque string as an output that is not meant to be manipulated by the consumer.
Unfortunately, that is impossible if one needs to format the output and that currently forces very dirty solutions that may break internationalization.

Example use case, is formatting minutes token to be smaller or different color than hours in hour:minute string.

``` javascript
var formatter = Intl.DateTimeFormat(navigator.languages, {
  hour: 'numeric',
  minute: 'numeric'
});
var string = formatter.format(new Date());
element.innerHTML = string; // we want to display hour token with bold font.
```

Modern user interfaces do that all the time - examples are Material Design in Android, modern Windows Mobile UI and PS4.

My proposal is to allow for token format strings to be available as option for the formatter. It could look like this:

``` javascript
var formatter = Intl.DateTimeFormat(navigator.languages, {
  hour: 'numeric',
  minute: 'numeric',
  format: {
    hour: '<strong>$&</strong>' // similar to String.replace(hour, '<strong>$&</strong>')
  }
});
var string = formatter.format(new Date()); // return '<strong>14</strong>:12' in pl
```

For non HTML outputs, the formatting may involve console formatting characters for terminal UI apps.

caridy:
I'm not sure about this at first glance, but it seems that we have some similarities between what you describe here and what a tagged template string does (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings#Tagged_template_strings), which ultimately give you more control over the formatting of the final output. Maybe we should think in those terms instead of introducing yet another template form.

Aside from that, having the custom `format` in the constructor prevent you from reusing the formatter in multiple places if the custom `format` is different.

zbraniecki:
I agree, I didn't spend too much time thinking about how the API should look like, and I'd be happy to reuse template strings as long as it allows for token formatting in datetime string construction.

You're also correct about format in constructor, but I would say that it may be a good tradeoff. Formatters should be consistent and if a customer needs to produce two different strings according to different formattings, he may need two different formatters.
But if it's easier to add a format template string as an argument for `formatter.format`, I'm totally fine with that.

srl295:
[...]

PR #64, opened 2016-01-14
[4rd Edition] DateTimeFormat.formatToParts()

zbraniecki:
Issue: https://github.com/tc39/ecma402/issues/30

## Features

* `CreateDateTimeParts` is a new abstract that powers `format()` and `formatToParts()` for `dataTimeFormat` instances.
* Moving all abstracts to the new Abstract section.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -223,16 +223,20 @@
         1. Return FormatDateTime(_dtf_, _x_).
       </emu-alg>
 
+      <emu-note>
+        The function returned by [[Get]] is bound to this DateTimeFormat object so that it can be passed directly to Array.prototype.map or other functions.
+      </emu-note>
+
       <p>
         The *length* property of a DateTime format function is 1.
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-formatdatetime" aoid="FormatDateTime">
-      <h1>FormatDateTime (dateTimeFormat, x)</h1>
+    <emu-clause id="sec-partitiondatetimepattern" aoid="PartitionDateTimePattern">
+      <h1>PartitionDateTimePattern (dateTimeFormat, x)</h1>
 
       <p>
-        When the FormatDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), it returns a String value representing _x_ (interpreted as a time value as specified in ES2017, <emu-xref href="#sec-time-values-and-time-range"></emu-xref>) according to the effective locale and the formatting options of _dateTimeFormat_. This abstract operation functions as follows:
+        The *PartitionDateTimePattern* abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), interprets _x_ as a time value as specified in ES2015, 20.3.1.1, and creates the corresponding parts according to the effective locale and the formatting options of _dateTimeFormat_. The following steps are taken:
       </p>
 
       <emu-alg>
@@ -247,17 +251,27 @@
         1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"useGrouping"`, *false*).
         1. Let _nf2_ be ? Construct(%NumberFormat%, « _nfLocale_, _nf2Options_ »).
         1. Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[calendar]], _dateTimeFormat_.[[timeZone]]).
-        1. Let _result_ be _dateTimeFormat_.[[pattern]].
-        1. For each row of <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, do:
-          1. If _dateTimeFormat_ has an internal slot with the name given in the Property column of the row, then
-            1. Let _p_ be the name given in the Property column of the row.
+        1. Let _pattern_ be _dateTimeFormat_.[[pattern]].
+        1. Let _result_ be a new empty List.
+        1. Let _beginIndex_ be Call(*%StringProto_indexOf%*, _pattern_, "{", *0*).
+        1. Let _endIndex_ be 0.
+        1. Let _nextIndex_ be 0.
+        1. Let _length_ be the number of code units in _pattern_.
+        1. Repeat while _beginIndex_ is an integer index into _pattern_:
+          1. Set _endIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _beginIndex_)
+          1. If _endIndex_ = -1, throw new Error exception.
+          1. If _beginIndex_ is greater than _nextIndex_, then:
+            1. Let _literal_ be a substring of _pattern_ from position _nextIndex_, inclusive, to position _beginIndex_, exclusive.
+            1. Add new part record { [[type]]: *"literal"*, [[value]]: _literal_ } as a new element of the list _result_.
+          1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
+          1. If _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then:
+
             1. Let _f_ be _dateTimeFormat_.[[<_p_>]].
             1. Let _v_ be _tm_.[[<_p_>]].
             1. If _p_ is *"year"* and _v_ ≤ 0, let _v_ be 1 - _v_.
             1. If _p_ is *"month"*, increase _v_ by 1.
             1. If _p_ is *"hour"* and _dateTimeFormat_.[[hour12]] is *true*, then
               1. Let _v_ be _v_ modulo 12.
-              1. If _v_ is equal to _tm_.[[<_p_>]], let _pm_ be *false*; else let _pm_ be *true*.
               1. If _v_ is 0 and _dateTimeFormat_.[[hourNo0]] is *true*, let _v_ be 12.
             1. If _f_ is *"numeric"*, then
               1. Let _fv_ be FormatNumber(_nf_, _v_).
@@ -265,19 +279,71 @@
               1. Let _fv_ be FormatNumber(_nf2_, _v_).
               1. If the *length* property of _fv_ is greater than 2, let _fv_ be the substring of _fv_ containing the last two characters.
             1. Else if _f_ is *"narrow"*, *"short"*, or *"long"*, then let _fv_ be a String value representing _f_ in the desired form; the String value depends upon the implementation and the effective locale and calendar of _dateTimeFormat_. If _p_ is *"month"*, then the String value may also depend on whether _dateTimeFormat_ has a [[day]] internal slot. If _p_ is *"timeZoneName"*, then the String value may also depend on the value of the [[inDST]] field of _tm_. If _p_ is *"era"*, then the String value may also depend on whether _dateTimeFormat_ has a [[era]] internal slot and if the implementation does not have a localized representation of _f_, then use _f_ itself.
-            1. Replace the substring of _result_ that consists of *"{"*, _p_, and *"}"*, with _fv_.
-        1. If _dateTimeFormat_.[[hour12]] is *true*, then
-          1. If _pm_ is *true*, then
-            1. Let _fv_ be an implementation and locale dependent String value representing "post meridiem".
-          1. Else,
-            1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
-          1. Replace the substring of _result_ that consists of *"{ampm}"*, with _fv_.
+            1. Add new part record { [[type]]: _p_, [[value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal "ampm", then:
+            1. Let _v_ be the value of _tm_.[[hour]].
+            1. If _v_ is greater than 11, then:
+              1. Let _fv_ be an implementation and locale dependent String value representing "post meridiem";
+            1. Else,
+              1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
+            1. Add new part record { [[type]]: *"dayPeriod"*, [[value]]: _fv_ } as a new element of the list _result_.
+            1. Let _literal_ be the substring of _pattern_ from position _beginIndex_, inclusive, to position _endIndex_, inclusive.
+            1. Add new part record { [[type]]: *"literal"*, [[value]]: _literal_ } as a new element of the list _result_.
+          1. Set _nextIndex_ to _endIndex_ + 1.
+          1. Set _beginIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _nextIndex_)
+        1. If _nextIndex_ is less than _length_, then:
+          1. Let _literal_ be the substring of _pattern_ from position _nextIndex_, exclusive, to position _length_, exclusive.
+          1. Add new part record { [[type]]: *"literal"*, [[value]]: _literal_ } as a new element of the list _result_.
         1. Return _result_.
       </emu-alg>
 
       <emu-note>
         It is recommended that implementations use the locale and calendar dependent strings provided by the Common Locale Data Repository (available at <a href="http://cldr.unicode.org/">http://cldr.unicode.org/</a>), and use CLDR "abbreviated" strings for DateTimeFormat "short" strings, and CLDR "wide" strings for DateTimeFormat "long" strings.
       </emu-note>
+
+      <emu-note>
+        It is recommended that implementations use the time zone information of the IANA Time Zone Database.
+      </emu-note>
+
+    </emu-clause>
+
+    <emu-clause id="sec-formatdatetime" aoid="FormatDateTime">
+      <h1>FormatDateTime(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be ? PartitionDateTimePattern(_dateTimeFormat_, _x_).
+        1. Let _result_ be an empty String.
+        1. For each _part_ in _parts_, do:
+          1. Set _result_ to a String value produced by concatenating _result_ and _part_.[[value]].
+        1. Return _result_.
+      </emu-alg>
+
+    </emu-clause>
+
+    <emu-clause id="sec-formatdatetimetoparts" aoid="FormatDateTimeToparts">
+      <h1>FormatDateTimeToParts(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatDateTimeToParts abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be ? PartitionDateTimePattern(_dateTimeFormat_, _x_).
+        1. Let _result_ be ArrayCreate(0).
+        1. Let _n_ be 0.
+        1. For each _part_ in _parts_, do:
+          1. Let _O_ be ObjectCreate(%ObjectPrototype%).
+          1. Perform ? CreateDataPropertyOrThrow(_O_, "type", _part_.[[type]]).
+          1. Perform ? CreateDataPropertyOrThrow(_O_, "value", _part_.[[value]]).
+          1. Perform ? CreateDataProperty(_result_, ? ToString(_n_), _O_).
+          1. Increment _n_ by 1.
+        1. Return _result_.
+      </emu-alg>
+
     </emu-clause>
 
     <emu-clause id="sec-tolocaltime" aoid="ToLocalTime">
@@ -455,6 +521,25 @@
           1. Perform ! DefinePropertyOrThrow(_bf_, `"length"`, PropertyDescriptor {[[Value]]: 1, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
           1. Set _dtf_.[[boundFormat]] to _bf_.
         1. Return _dtf_.[[boundFormat]].
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-Intl.DateTimeFormat.prototype.formatToParts">
+      <h1>Intl.DateTimeFormat.prototype.formatToParts ([ date ])</h1>
+
+      <p>
+        When the *Intl.DateTimeFormat.prototype.formatToParts* is called with an optional argument _date_, the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Let _dtf_ be *this* value.
+        1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
+        1. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot, throw a *TypeError* exception.
+        1. If _date_ is not provided or is *undefined*, then
+          1. Let _x_ be *%Date_now%*().
+        1. Else,
+          1. Let _x_ be ? ToNumber(_date_).
+        1. Return ? FormatDateTimeToParts(_dtf_, _x_).
       </emu-alg>
     </emu-clause>
 


spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
     </emu-clause>
 
-    <emu-clause id="sec-formatdatetime" aoid="FormatDateTime">
-      <h1>FormatDateTime (dateTimeFormat, x)</h1>
+    <emu-clause id="sec-datetime-formatToParts-functions">
+      <h1>DateTime FormatToParts Functions</h1>
 
       <p>
-        When the FormatDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), it returns a String value representing _x_ (interpreted as a time value as specified in ES2015, 20.3.1.1) according to the effective locale and the formatting options of _dateTimeFormat_. This abstract operation functions as follows:
+        A DateTime formatToParts function is an anonymous function that optionally takes an argument _date_. It performs the following steps:
```
caridy: When a DateTime formatToParts function is called with optional argument _date_, the following steps are taken:

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
+        1. Else,
+          1. Let _x_ be ? ToNumber(_date_).
+        1. Return FormatToPartsDateTime(_dtf_, _x_).
+      </emu-alg>
+
+      <emu-note>
+        The function returned by [[Get]] is bound to this DateTimeFormat object so that it can be passed directly to Array.prototype.map or other functions.
+      </emu-note>
+
+      <p>
```
caridy: remove this paragraph.

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
+        1. Let _dtf_ be the *this* value.
+        1. Assert: Type(_dtf_) is Object and _dtf_ has an [[initializedDateTimeFormat]] internal slot whose value is *true*.
+        1. If _date_ is not provided or is *undefined*, then
+          1. Let _x_ be *%Date_now%*().
+        1. Else,
+          1. Let _x_ be ? ToNumber(_date_).
+        1. Return FormatToPartsDateTime(_dtf_, _x_).
+      </emu-alg>
+
+      <emu-note>
```
caridy: remove, this is not an accessor.

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
       <emu-alg>
+        1. Let _dtf_ be the *this* value.
+        1. Assert: Type(_dtf_) is Object and _dtf_ has an [[initializedDateTimeFormat]] internal slot whose value is *true*.
+        1. If _date_ is not provided or is *undefined*, then
+          1. Let _x_ be *%Date_now%*().
+        1. Else,
+          1. Let _x_ be ? ToNumber(_date_).
+        1. Return FormatToPartsDateTime(_dtf_, _x_).
+      </emu-alg>
+
```
caridy: Missing: "The *length* property of a DateTime formatToParts function is 1."

spec/datetimeformat.html
```diff
@@ -452,6 +556,26 @@
...
+      <h1>get Intl.DateTimeFormat.prototype.formatToParts</h1>
+
+      <p>
+        Intl.DateTimeFormat.prototype.formatToParts is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _dtf_ be *this* value.
+        1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
+        1. If _dtf_ does not have a [[boundFormatToParts]] internal slot, throw a *TypeError* exception.
```
caridy: ... not have a [[initializedDateTimeFormat]] internal slot.

this is because NumberFormat instances will have the same internal slot.

spec/datetimeformat.html
```diff
@@ -452,6 +556,26 @@
...
+        1. If the [[boundFormatToParts]] internal slot of _dtf_ is *undefined*, then
+          1. Let _F_ be a new built-in function object as defined in DateTime Format Functions (<emu-xref href="#sec-datetime-formatToParts-functions"></emu-xref>).
```
caridy: ... defined in DateTime FormatToParts Functions...

spec/datetimeformat.html
```diff
@@ -452,6 +556,26 @@
...
+          1. The value of _F_’s *length* property is 1.
```
caridy: remove this step, not needed since it is signaled in the functions abstract.

caridy: move this lined below the creation of the bound function, and use this syntax:

```
1. Perform ! DefinePropertyOrThrow(_bf_, *"length"*, PropertyDescriptor {[[Value]]: 1, [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: true}).
```

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+
+      <h1>FormatToPartDateTime(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatToPartDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be CreateDateTimeParts(_dateTimneFormat_, _x_).
+        1. Let _result_ be ArrayCreate(0).
```
caridy: this abstract should probably create just a List, and append objects to it before calling CreateArrayFromList() in the return statement.

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+
+      <h1>FormatDateTime(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be CreateDateTimeParts(_dateTimneFormat_, _x_).
+        1. Let _result_ be an empty string.
```
caridy: ... empty String.

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+    <emu-clause id="alg-FormatDateTime" aoid="DateTimeFormat">
+
+      <h1>FormatDateTime(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be CreateDateTimeParts(_dateTimneFormat_, _x_).
```
caridy: ... be ? CreateDateTimeParts...

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
+
+      <h1>CreateDateTimeParts(dateTimeFormat, x)</h1>
+
+      <p>
+        The CreateDateTimeParts abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg aoid="CreateDateTimeParts">
         1. If _x_ is not a finite Number, throw a *RangeError* exception.
         1. Let _locale_ be the value of _dateTimeFormat_.[[locale]].
```
caridy: we should probably assert() that _dateTimeFormat_ is what it should be before this step.

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+          1. Set _beginIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _index_)
+        1. If _endIndex_ is less than _S_, then:
+          1. Let _separator_ to be a new Record.
+          1. Let _separatorValue_ be the substring of _pattern_ from position _endIndex_, exclusive, to position _S_, exclusive.
+          1. Add new part record { [[type]]: "separator", [[value]]: _separatorValue_ } as a new element of the list _result_.
         1. Return _result_.
       </emu-alg>
 
       <emu-note>
+        CreateDateTimeParts abstract operation interpretes _x_ as a time value as specified in ES2015, 20.3.1.1; and create the corresponding parts according to the effective locale and the formatting options of _dateTimeFormat_.
```
caridy: we should probably have a table describing each different type produced by this abstract operation.

caridy:
Also, we should resolve the conflicts with master before producing a new HTML for the formal review from @littledan

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
-          1. If _pm_ is *true*, then
-            1. Let _fv_ be an implementation and locale dependent String value representing "post meridiem";
+            1. Add new part record { [[type]]: _p_, [[value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal "ampm", then:
+            1. Let _v_ be the value of _tm_.[[hour]].
+            1. If _v_ is greater than 11, then:
+              1. Let _fv_ be an implementation and locale dependent String value representing "post meridiem";
+            1. Else,
+              1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
+            1. Add new part record { [[type]]: "dayperiod", [[value]]: _fv_ } as a new element of the list _result_.
```
anba: Shouldn't this part be named to "dayPeriod" for consistency with other compound names (e.g. "timeZoneName")?

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
           1. Else,
-            1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
-          1. Replace the substring of _result_ that consists of *"{ampm}"*, with _fv_.
+            1. <emu-note>Unknown token</emu-note>
+            1. Let _separatorValue_ be the substring of _pattern_ from position _beginIndex_, inclusive, to position _endIndex_, inclusive.
+            1. Add new part record { [[type]]: "separtor", [[value]]: _separatorValue_ } as a new element of the list _result_.
```
anba: Typo: "separtor" -> "separator"

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+    <emu-clause id="alg-FormatToPartDateTime" aoid="DateTimeFormatToParts">
+
+      <h1>FormatToPartDateTime(dateTimeFormat, x)</h1>
+
+      <p>
+        The FormatToPartDateTime abstract operation is called with arguments _dateTimeFormat_ (which must be an object initialized as a DateTimeFormat) and _x_ (which must be a Number value), and performs the following steps:
+      </p>
+
+      <emu-alg>
+        1. Let _parts_ be CreateDateTimeParts(_dateTimneFormat_, _x_).
```
anba: Typo: "dateTimneFormat" -> "dateTimeFormat"

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+        1. Let _result_ be ArrayCreate(0).
+        1. Let _n_ be 0.
+        1. For each _part_ in _parts_, do:
+          1. Let _O_ be ObjectCreate(%ObjectPrototype%).
+          1. Let _typeDesc_ be the PropertyDescriptor{[[Value]]: _part_.[[type]], [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true}.
+          1. Perform ? DefinePropertyOrThrow(_O_, "type", _typeDesc_).
```
anba: This can be simplified using either [CreateDataProperty](http://tc39.github.io/ecma262/#sec-createdataproperty) or [CreateDataPropertyOrThrow](http://tc39.github.io/ecma262/#sec-createdatapropertyorthrow).

caridy:
Plus, `_dateTimneFormat_` has a typo, in two places.

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
+      </emu-note>
+
+      <p>
+        The value of the [[Set]] attribute is *undefined*.
+      </p>
+    </emu-clause>
+
+    <emu-clause id="alg-CreateDateTimeParts">
+
+      <h1>CreateDateTimeParts(dateTimeFormat, x)</h1>
```
stasm: In #79 @caridy wanted to use the verb _format_ for a similar abstract operation: `FormatNumberToParts`. Perhaps we could rename this here as well for consistency to `FormatDateTimeToParts'?  This will also be more inline with the name of the existing abstract operation: `FormatDateTime`.

spec/datetimeformat.html
```diff
@@ -452,6 +556,26 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-Intl.DateTimeFormat.prototype.formatToParts">
+      <h1>get Intl.DateTimeFormat.prototype.formatToParts</h1>
+
+      <p>
+        Intl.DateTimeFormat.prototype.formatToParts is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:
```
stasm: Similar to #79, should we make this a regular method and not an accessor?  We could then remove the binding mechanism from the `formatToParts` API here.

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
+        1. For each _part_ in _parts_, do:
+          1. Set _result_ to a String value produced by concatenating _result_ and _part_.[[value]].
+        1. Return _result_.
+      </emu-alg>
+
+    </emu-clause>
+
+    <emu-clause id="alg-FormatToPartDateTime" aoid="DateTimeFormatToParts">
+
+      <h1>FormatToPartDateTime(dateTimeFormat, x)</h1>
```
stasm: Perhaps I'm missing something, but where is this abstract operation used?  The `aoid` seems to be different from the name of the operation.

spec/datetimeformat.html
```diff
@@ -224,35 +225,75 @@
...
+        1. Let _pattern_ be the value of the _dateTimeFormat_.[[pattern]].
+        1. Let _result_ be a new empty List.
+        1. Let _index_ be 0.
+        1. Let _beginIndex_ be Call(*%StringProto_indexOf%*, _pattern_, "{", *0*).
+        1. Let _endIndex_ be 0.
+        1. Repeat while _beginIndex_ is an integer index into _pattern_:
+        1. Set _endIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _beginIndex_)
+          1. If _endIndex_ is *false*, throw new Error exception.
+          1. If _beginIndex_ is greater than _index_, then:
+            1. Let _separator_ be a new Record.
```
stasm: Is `_separator_` used anywhere?

spec/datetimeformat.html
```diff
@@ -260,19 +301,82 @@
...
           1. Else,
-            1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
-          1. Replace the substring of _result_ that consists of *"{ampm}"*, with _fv_.
+            1. <emu-note>Unknown token</emu-note>
+            1. Let _separatorValue_ be the substring of _pattern_ from position _beginIndex_, inclusive, to position _endIndex_, inclusive.
+            1. Add new part record { [[type]]: "separtor", [[value]]: _separatorValue_ } as a new element of the list _result_.
+          1. Set _index_ to _endIndex_ + 1.
+          1. Set _beginIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _index_)
+        1. If _endIndex_ is less than _S_, then:
+          1. Let _separator_ to be a new Record.
```
stasm: Same here: is `_separator_` used anywhere?

zbraniecki:
@caridy - rebased on top of master and applied stas's feedback

spec/datetimeformat.html
```diff
@@ -458,6 +500,24 @@
...
+      </p>
+
+      <emu-alg>
+        1. Let _dtf_ be *this* value.
+        1. Assert: Type(_dtf_) is Object and _dtf_ has an [[initializedDateTimeFormat]] internal slot whose value is *true*.
+        1. If _date_ is not provided or is *undefined*, then
+          1. Let _x_ be *%Date_now%*().
+        1. Else,
+          1. Let _x_ be ? ToNumber(_date_).
+        1. Return ? FormatDateTimeToParts(_dtf_, _x_).
```
littledan: I don't think you can just directly return the result of FormatDateTimeToParts to ECMAScript. That returns a List of Records. You have to convert the List of Records into an Array of Objects.

zbraniecki: Hmm, it must have got lost somewhere in the process, because I remember writing the conversion here. I'll update the spec, thanks!

caridy: yeah, I remember that process as well.

zbraniecki: So, @stasm removed it in his editorial patch for my PR - https://github.com/stasm/ecma402/commit/3b35e0b63dd99bf8df350a96f80b349aef8d1f5f

I'll bring it back.

stasm: Yeah, sorry about that.  In https://github.com/tc39/ecma402/pull/64#discussion_r53446286 I asked if it was used at all and thought it was an copy-paste artifact given the mismatched `id` and `aoid`.

caridy: @stasm can you also double check that in `NumberFormatToParts`?

zbraniecki: fixed

spec/datetimeformat.html
zbraniecki: how?

zbraniecki:
@littledan - I applied your feedback.

@littledan, @caridy - @stasm suggested changing the name of the abstract operation `FormatDateTimeToParts` to something less confusing with the bound `FormatToPartsDateTime` and `FormatDateTime`.
Would you be ok to switch it to, for example `DeconstructDateTimePattern` ?

spec/datetimeformat.html
caridy: `CreateDataPropertyOrThrow(_O_, "type", _part_)` should be sufficient, not need to create the descriptor, and add each data property directly.

caridy: here is an example: https://tc39.github.io/ecma262/#sec-proxy.revocable

littledan:
Name change sounds good, and @anba's suggestion is good. I like how you've edited it generally.

zbraniecki:
cool! updated

spec/datetimeformat.html
```diff
@@ -223,16 +223,20 @@
...
+      </emu-note>
+
       <p>
         The *length* property of a DateTime format function is 1.
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-formatdatetime" aoid="FormatDateTime">
-      <h1>FormatDateTime (dateTimeFormat, x)</h1>
+    <emu-clause id="sec-formatdatetimetoparts" aoid="PartitionDateTimePattern">
```
stasm: Update the `id` here?

spec/datetimeformat.html
```diff
@@ -265,19 +278,72 @@
...
+        1. Let _result_ be an empty String.
+        1. For each _part_ in _parts_, do:
+          1. Set _result_ to a String value produced by concatenating _result_ and _part_.[[value]].
+        1. Return _result_.
+      </emu-alg>
+
+    </emu-clause>
+
+    <emu-clause id="sec-formattopartsdatetime" aoid="FormatToPartsDateTime">
+      <h1>FormatToPartsDateTime(dateTimeFormat, x)</h1>
```
stasm: `FormatDateTimeToParts`?

zbraniecki: The function it is a counter part to is "FormatDateTime" (Fn+Object), so I'm sticking to match to it.

spec/datetimeformat.html
zbraniecki: thnx! fixed

spec/datetimeformat.html
stasm: I suggest we use names that read well in English.  "Format DateTime to Parts" is a valid English sentence, and it works as a counter-part to "Format DateTime" just as well as your proposal.

spec/datetimeformat.html
stasm: @caridy I updated #79 to be in line with @zbraniecki's changes in this PR.  Should be good to review now! :)

spec/datetimeformat.html
```diff
@@ -265,19 +278,72 @@
...
+
+      <emu-alg>
+        1. Let _parts_ be ? PartitionDateTimePattern(_dateTimeFormat_, _x_).
+        1. Let _result_ be ArrayCreate(0).
+        1. Let _n_ be 0.
+        1. For each _part_ in _parts_, do:
+          1. Let _O_ be ObjectCreate(%ObjectPrototype%).
+          1. CreateDataPropertyOrThrow(_O_, "type", _part_.[[type]]).
+          1. CreateDataPropertyOrThrow(_O_, "value", _part_.[[value]]).
+          1. Perform ? DefinePropertyOrThrow(_O_, "value", _valueDesc_).
```
littledan: The preceding three lines look funny in an identical way to NumberFormat.formatToParts. Do you think you could put the two bodies in a single abstract algorithm (leaving out the first step) to reduce duplication?

caridy: that will be true for all `.formatToParts()` in the future, including relative time, etc.

littledan:
LGTM on the new version

spec/datetimeformat.html
```diff
@@ -247,37 +251,98 @@
...
-            1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
-          1. Replace the substring of _result_ that consists of *"{ampm}"*, with _fv_.
+            1. Add new part record { [[type]]: _p_, [[value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal "ampm", then:
+            1. Let _v_ be the value of _tm_.[[hour]].
+            1. If _v_ is greater than 11, then:
+              1. Let _fv_ be an implementation and locale dependent String value representing "post meridiem";
+            1. Else,
+              1. Let _fv_ be an implementation and locale dependent String value representing "ante meridiem".
+            1. Add new part record { [[type]]: *"dayPeriod"*, [[value]]: _fv_ } as a new element of the list _result_.
```
caridy: dayPeriods?

spec/datetimeformat.html
```diff
@@ -247,37 +251,98 @@
...
+            1. Let _literal_ be the substring of _pattern_ from position _beginIndex_, inclusive, to position _endIndex_, inclusive.
+            1. Add new part record { [[type]]: *"literal"*, [[value]]: _literal_ } as a new element of the list _result_.
+          1. Set _nextIndex_ to _endIndex_ + 1.
+          1. Set _beginIndex_ to Call(*%StringProto_indexOf%*, _pattern_, "}", _nextIndex_)
+        1. If _nextIndex_ is less than _length_, then:
+          1. Let _literal_ be the substring of _pattern_ from position _nextIndex_, inclusive, to position _length_, exclusive.
```
caridy: in the polyfill implementation, I found that this condition should be `_nextIndex_, exclusive`, please, verify that. /cc @zbraniecki

spec/datetimeformat.html
zbraniecki: Why `dayPeriods`? We call the token `dayPeriod` - https://github.com/andyearnshaw/Intl.js/blob/master/src/12.datetimeformat.js#L1011

caridy: hmm, I got confused, look two lines above the link you provided, it is using dayPeriods, so I was confused here. we need to update the polyfill then :)

zbraniecki: Why? 

The internal `resolveDateString` call uses `dayPeriods`, both before formatToParts landed [0] and after.
The public token name is `dayPeriod`.

What do you need me to update in the polyfill?

[0] https://github.com/andyearnshaw/Intl.js/commit/a9743c360cab41dc75230e9a0950931abdabb9a0#diff-7eb52b366866677666470e019283c8eaL2489

caridy: It is confusing to have two things that are so similar, `dayPeriod` and `dayPeriods`. But I will review it myself, see if that makes sense or not.

spec/datetimeformat.html
zbraniecki: great catch! Fixed. It should be exclusive of course. (we don't need `}` as part of the trailing literal)

spec/datetimeformat.html
zbraniecki: so, in CLDR we have a list `dayPeriods` - https://github.com/andyearnshaw/Intl.js/blob/2fd2654d2adc8f870fa3f0616707d333135cc9f1/locale-data/json/en-US.json#L168

And then in formatToParts we create a token with a single `dayPeriod`.

zbraniecki:
Heads up, I rebased on top of master and moved `prototype.format*` functions to throw if the params are wrong instead of asserting.

caridy:
LGTM @zbraniecki, now we should wait for consensus at TC39 before merging this.

littledan:
Not just consensus at TC39 but Stage 4, with two implementations (I'd prefer two browser implementations, personally).

zbraniecki:
@littledan - we now have two implementations - one in SpiderMonkey and one in Intl.js. I'd love to see two browser implementations but that depends on the implementers. Do you know when we can see the feature in V8?

zbraniecki:
With test262 in pipeline, I'm planning to ask for the second implementer to step up at the current TC39 meeting. I'd like to break the chicken-egg problem between 2 impl. and stage 4.

spec/datetimeformat.html
```diff
@@ -247,37 +251,99 @@
...
+
+      <emu-alg>
+        1. Let _parts_ be ? PartitionDateTimePattern(_dateTimeFormat_, _x_).
+        1. Let _result_ be ArrayCreate(0).
+        1. Let _n_ be 0.
+        1. For each _part_ in _parts_, do:
+          1. Let _O_ be ObjectCreate(%ObjectPrototype%).
+          1. Perform ? CreateDataPropertyOrThrow(_O_, "type", _part_.[[type]]).
+          1. Perform ? CreateDataPropertyOrThrow(_O_, "value", _part_.[[value]]).
+          1. Perform ? CreateDataProperty(_result_, ? ToString(_n_), _O_).
```
leobalter: `ToString(_n_)` will not throw. `?` can be replaced by `!`.

Actually I believe none of these last 3 lines will throw anything. All of the `?` can be replaced by `!` on them.

spec/datetimeformat.html
```diff
@@ -247,37 +251,99 @@
         1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"useGrouping"`, *false*).
         1. Let _nf2_ be ? Construct(%NumberFormat%, « _nfLocale_, _nf2Options_ »).
```
leobalter: Once again, I believe this construct and the other 4 lines above won't return any abrupt completion.

caridy:
merged as 4a711f0b80c9ea09a64f273a94a36bc390f0564f

PR #67, opened 2016-01-25
Refactor to group abstract operations per section for `NumberFormat` and `DateTimeFormat`

caridy:
__Disclaimer: This PR does not modify any algo, we just reshuffle them for organizational purposes.__

### Reasoning

* "Number Format Functions" holds several abstract operations, and it was a sub-section under "Properties of the Intl.NumberFormat Prototype Object", which is confusing, specially once we add more properties to those objects that could call the same abstract operations. The same applies to "DateTime Format Functions".
* `InitializeNumberFormat()` and `InitializeDateTimeFormat()` abstract operations were under the corresponding constructor section.

### Notes
* moving all abstract operations of NumberFormat Objects into an abstract sub-section
* moving all abstract operations of DateTimeFormat Objects into an abstract sub-section.

### Render Version: 

* https://rawgit.com/caridy/c80267817f2736db7de8/raw/f7df1b2c9de232a69dbcf9e6e4eceb6966fb9eeb/ecma-402-pr1.html#numberformat-objects

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -1,12 +1,8 @@
 <emu-clause id="numberformat-objects">
   <h1>NumberFormat Objects</h1>
 
-  <emu-clause id="sec-intl-numberformat-constructor">
-    <h1>The Intl.NumberFormat Constructor</h1>
-
-    <p>
-      The NumberFormat constructor is a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in <emu-xref href="#sec-internal-slots"></emu-xref>.
-    </p>
+  <emu-clause id="sec-numberformat-abstracts">
+    <h1>Abstract Operations For NumberFormat Objects</h1>
 
     <emu-clause id="sec-InitializeNumberFormat" aoid="InitializeNumberFormat">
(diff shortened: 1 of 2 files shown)


spec/numberformat.html
```diff
@@ -456,6 +255,219 @@
...
+      <p>
+        The initial value of the @@toStringTag property is the string value *"Object"*.
+      </p>
+      <p>
+        This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
+      </p>
+    </emu-clause>
+
+    <emu-clause id="sec-Intl.NumberFormat.prototype.format">
+      <h1>get Intl.NumberFormat.prototype.format</h1>
```
zbraniecki: Did you want to remove `get` here as well (like you did in DateTimeFormat)?

zbraniecki: oh, you added it, nevermind.

zbraniecki:
+1.

Cool! Let's land it before we move forward with revision 3 changes.

rwaldron:
@caridy this LGTM, thanks for putting in the time and effort to make these changes

Issue #57, opened 2015-11-24
[Compatibility Hazard] ECMA 402 2nd Edition Changed the [[Call]] Behavior of Intl Constructors


ericf:
The 1st Edition of ECMA 402 specified the [[Call]] behavior for `Intl` constructors; e.g. [`Intl.DateTimeFormat.call(this [, locales [, options]])`](http://ecma-international.org/ecma-402/1.0/index.html#sec-12.1.2.1) to return the `this` context object that was passed-in. In the 2nd Edition, the [[Call]] behavior no longer states that the passed-in context object should be retuned. **This change is a potential compatibility hazard.**

As the developer and maintainer of the popular [FormatJS](http://formatjs.io/) i18n libraries, I've begun receiving issues from developers testing **Chrome Canary (49)** that their dates and numbers were failing for format, causing an Error to be thrown.

The high-level framework integration libs that are part of FormatJS — [react-intl](https://github.com/yahoo/react-intl), [ember-intl](https://github.com/yahoo/ember-intl), [handlebars-intl](https://github.com/yahoo/handlebars-intl) — are used by many web apps, including many of Yahoo's web apps. All these libraries use the underlying [intl-format-cache](https://github.com/yahoo/intl-format-cache) which memoizes the `Intl` constructors because they are expensive to create. The memoization technique essentially does the following:

``` js
function constructIntlInstance(IntlConstructor) {
    return function () {
        var args = Array.prototype.slice.call(arguments);
        var instance = Object.create(IntlConstructor.prototype);
        IntlConstructor.apply(instance, args);
        return instance;
    };
}
```

**Note:** That this code depends on the following invariant:

``` js
var instance = Object.create(IntlConstructor.prototype);
instance === IntlConstructor.call(instance); // true
```

**It is dependent on the the `Intl` constructors being `.call()`-able and the returning the context object passed-in.** This `.call()` behavior for the `Intl` constructors is supported in all ECMA 402 1st Edition implementations.

[...]

PR #70, opened 2016-02-03
Various fixes

anba:
The changes are for the most part only small bug fixes. I've split them into multiple commits, so it's easier to review them. Please let me know if any of the proposed changes are unclear.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -458,7 +458,7 @@
         1. Let _nf_ be *this* value.
         1. If Type(_nf_) is not Object, throw a *TypeError* exception.
         1. If _nf_ does not have an [[initializedNumberFormat]] internal slot, throw a *TypeError* exception.
-        1. If _nf_.[[initializedNumberFormat]] is *undefined*, then
+        1. If _nf_.[[boundFormat]] is *undefined*, then
           1. Let _F_ be a new built-in function object as defined in Number Format Functions (<emu-xref href="#sec-number-format-functions"></emu-xref>).
           1. Let _bf_ be BoundFunctionCreate(_F_, _nf_, « »).
           1. Perform ! DefinePropertyOrThrow(_bf_, `"length"`, PropertyDescriptor {[[Value]]: 1, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true*}).
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -379,8 +379,8 @@
 
       <emu-alg>
         1. Let _availableLocales_ be *%NumberFormat%*.[[availableLocales]].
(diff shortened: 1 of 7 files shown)


spec/numberformat.html
```diff
@@ -457,10 +457,11 @@
       <emu-alg>
         1. Let _nf_ be *this* value.
         1. If Type(_nf_) is not Object, throw a *TypeError* exception.
-        1. If _nf_ does not have a [[boundFormat]] internal slot, throw a *TypeError* exception.
-        1. If _nf_.[[boundFormat]] is *undefined*, then
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot, throw a *TypeError* exception.
+        1. If _nf_.[[initializedNumberFormat]] is *undefined*, then
```
caridy: last time I checked `[[initializedNumberFormat]]`, it was a flag to indicate that the instance was correctly created (soft of a dummy flag). As for this condition here, it is just to define the boundFormat internal slot once (the first time .format is used).

anba: Intl.NumberFormat and Intl.DateTimeFormat both use the internal slot name [[boundFormat]], but we need to check that the **this** value in [Intl.NumberFormat.prototype.format](https://tc39.github.io/ecma402/#sec-Intl.NumberFormat.prototype.format) is a NumberFormat instance, not a DateTimeFormat instance. Otherwise we'd violate the assertion in [NumberFormat format function, step 2](https://tc39.github.io/ecma402/#sec-number-format-functions).

anba: We may even need to restore the `If _nf_.[[initializedNumberFormat]] is *false*, throw a *TypeError* exception.` step from the first and second edition (the "this NumberFormat object" phrase validation steps), depending on how the #57 gets resolved.

caridy: I noticed that you didn't change this line for collator and datetimeformat, so I guess this one is just a mistake. Other than this, everything else looks great. I can merge it once you fix this small detail.

caridy: ah, I missed those couple of messages. The logic there is (for number, datetime and collator), we check for identity first (checking if the corresponding initialized* internal slot exists, then we fork on whether of not the `boundFormat` function was created. I don't see the problem you're describing above.

caridy: as for #57, nothing will change in the algos, just a non-normative note that will be clarify!

anba: > I noticed that you didn't change this line for collator and datetimeformat, so I guess this one is just a mistake.

No, I've changed all three constructors: https://github.com/anba/ecma402/commit/c73b7cf3ab17bc3bd80f241eb0f83674e3be4a82

anba: Ups. Yes, that's a typo NumberFormat. Sorry for the confusion!

caridy: n.p., I can merge this tomorrow!

PR #74, opened 2016-02-11
Ensure ResolveLocale with BestFitMatcher returns valid language tags

anba:
Add the supported Unicode locale extension sequence at the end of the
'foundLocale' language tag, but before any private-use sequence. Also
call CanonicalizeLanguageTag to ensure the newly formed language tag
is canonicalized (This step is only required for implementations which
implement the optional Unicode locale extension canonicalization from
RFC6067 and when the 'relevantExtensionKeys' list is not sorted.).

Fixes https://bugs.ecmascript.org/show_bug.cgi?id=1456

### DIFF
--- a/spec/conventions.html
+++ b/spec/conventions.html
@@ -87,6 +87,11 @@
           <td>The initial value of the *"includes"* data property of the intrinsic %StringPrototype% (ES2016, 21.1.3.7)</td>
         </tr>
         <tr>
+          <td>%StringProto_indexOf%</td>
+          <td>"String.prototype.indexOf"</td>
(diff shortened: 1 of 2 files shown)


spec/conventions.html
```diff
@@ -87,6 +87,11 @@
           <td>The initial value of the *"includes"* data property of the intrinsic %StringPrototype% (ES2016, 21.1.3.7)</td>
         </tr>
         <tr>
+          <td>%StringProto_indexOf%</td>
```
caridy: hmm, you decided to bring this back!

caridy:
@anba +1 from me, but we should probably wait for @rwaldron to take a look at this since it was involved in the original discussion.

caridy:
I did a second pass on this, including the referral implementation in Intl.js! merging!

anba:
\o/

PR #77, opened 2016-02-16
close #53: editorial fix to avoid using abstract op as values

### LINKED ISSUE #53: Rework spec text in 9.2.8: SupportedLocales (availableLocales, requestedLocales, options) to link to the abstract operations
image

there isn't a notion of an abstract op as a value, instead we should just call the corresponding abstract directly in the then or the else.

caridy:


ericf:
:+1:

Issue #46, opened 2015-10-16
Intl.Locale API


zbraniecki:
I started working on a spec and polyfill for Intl.Locale API.

I should have some proposal for this next week. Currently I have five functions on it:
 * isStructurallyValidLanguageTag(locale)
 * canonicalizeLocaleList(locales)
 * resolveLocale(availableLocales, requestedLocales, options, relevantExtensionKeys, localeData)
 * prioritizeAvailableLocales(availableLocales, requestedLocales, defaultLocale)
 * getDirection(locale)

The rationale for those five are:

* isStructurallyValidTag
This allows testing language tag for sanitation/validation purposes.

* canonicalizeLocaleList
This is the core function allowing all language negotiation operations

* resolveLocale

this exposes the core function used by all Intl formatters that allows people to write their own custom formatters or polyfills.

* prioritizeAvailableLocales

This is basic language negotiation function for all localization libraries. It takes requestedLocales list (navigator.languages or custom), availableLocales (app provided locales), and a fallback defaultLocale (default app locale, not default host env locale) and returns the list of prioritized available locales that best matches user requested locales.

* getDirection

Simple function allowing l10n libraries to set html.dir based on their negotiated locale.

@caridy  - One question I'd like to get help with is about using `Set` instead of `Array`. In all cases where we carry a list of locales, we always want it deduplicated and `Set` does just that. It would be sweet to use it instead of an `Array`, but it's API seems to be significantly less user friendly than that of the `Array` - it's impossible to just retrieve `locales[0]`, it's not easy to `map` etc.
Should I stick to `Array` or try to use `Set`?

caridy:
Two things:

 * You can certainly use a Set internally as part of the logic, but let's stick to an array for arguments, I haven't seen any precedent of using a Set as arguments in Ecma.
 * Instead of implementing a new polyfill (once we settle on the api) you can just submit a PR for Intl.js, which already implements those abtract operations.

zbraniecki:
sure! thanks for the feedback :)

zbraniecki:
@caridy - I started playing with Intl.js and ended up separating `src/locale.js` out of `src/core.js` and then carved out the common functions into `src/utils.js` because `src/core.js` is huuge :)
Not sure if you're interested in any sort of refactor like that - if not, I'll bring it back into `src/core.js`.

Branch is here: https://github.com/zbraniecki/Intl.js/tree/locale

[...]

PR #78, opened 2016-02-18
Exposing `CanonicalizeLocaleList()` abstract operation via `Intl.getCanonicalLocales()`

caridy:
Feature from https://github.com/tc39/ecma402/issues/46

This is probably the only new feature we will be able to get into 3rd edition.

This PR also opens the door for adding more function properties to `Intl` in the future, very similar to `Math` in 262.

/cc @littledan @zbraniecki @ericf @rwaldron


zbraniecki:
LGTM

PR #80, opened 2016-02-27
[Editorial] improving cross linking and other small things

caridy:
* all links to be lowercase (to match 262)
* all reference to ES2016 should use `<emu-xref>`
* supersede methods from 402 should be linkable (using "sup-" prefix instead of "sec-" for those)
* correct spacing on headers and invocations
* consistently use a comma before "the following steps are taken"

### DIFF
--- a/spec/annexes.html
+++ b/spec/annexes.html
@@ -25,10 +25,10 @@
           The set of available locales for each constructor (<emu-xref href="#sec-internal-slots"></emu-xref>)
         </li>
         <li>
-          The BestFitMatcher algorithm (<emu-xref href="#sec-BestFitMatcher"></emu-xref>)
+          The BestFitMatcher algorithm (<emu-xref href="#sec-bestfitmatcher"></emu-xref>)
         </li>
         <li>
-          The BestFitSupportedLocales algorithm (<emu-xref href="#sec-BestFitSupportedLocales"></emu-xref>)
+          The BestFitSupportedLocales algorithm (<emu-xref href="#sec-bestfitsupportedlocales"></emu-xref>)
         </li>
       </ul>
     </li>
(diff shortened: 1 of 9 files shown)


spec/annexes.html
```diff
@@ -25,10 +25,10 @@
           The set of available locales for each constructor (<emu-xref href="#sec-internal-slots"></emu-xref>)
         </li>
         <li>
-          The BestFitMatcher algorithm (<emu-xref href="#sec-BestFitMatcher"></emu-xref>)
+          The BestFitMatcher algorithm (<emu-xref href="#sec-bestfitmatcher"></emu-xref>)
```
ericf: Curious… why aren't the `emu-xref`s self-closing elements?

spec/index.html
```diff
@@ -1,58 +1,49 @@
...
-    <emu-import href="./negotiation.html"></emu-import>
-    <emu-import href="./collator.html"></emu-import>
-    <emu-import href="./numberformat.html"></emu-import>
-    <emu-import href="./datetimeformat.html"></emu-import>
-    <emu-import href="./locale-sensitive-functions.html"></emu-import>
-    <emu-import href="./annexes.html"></emu-import>
-
-  </body>
-</html>
+<meta charset="ascii">
```
ericf: Why not `utf-8`?

spec/annexes.html
caridy: http://bterlson.github.io/ecmarkup/#emu-xref because you can have a text, instead of the index value.

spec/index.html
caridy: not sure, this came from 262.

ericf:
:+1:

PR #82, opened 2016-02-29
closes #76: annex B - 'length' property of DateTime Format Function set to 1 in 2nd edition

### LINKED ISSUE #76: 'length' property of DateTime Format Function
The 'length' property of the DateTime Format function returned from [get Intl.DateTimeFormat.prototype.format](https://tc39.github.io/ecma402/#sec-Intl.DateTimeFormat.prototype.format) changed from 0 to 1 in ECMA-402, 2nd ed. 

In ECMA-402, 1st ed:
> a. Let F be a Function object, with internal properties set as specified for built-in functions in ES5, 15, or successor, and the length property set to 0, that takes the argument date and performs the following steps:

In ECMA-402, 2nd ed:
> a. Let F be a new built-in function object as defined in 12.3.4.
b. The value of F’s length property is 1.

The change is not listed in Annex B, so I'm not sure if it was intentional. 

(V8 and SpiderMonkey still return 0, test262 also tests for 0 (https://github.com/tc39/test262/blob/master/test/intl402/DateTimeFormat/prototype/format/12.3.2_1_a_L15.js). Only JavaScriptCore already returns 1.)

caridy:


stasm:
:+1:

spec/annexes.html
```diff
@@ -116,5 +116,8 @@
     <li>
       <emu-xref href="#sec-the-intl-collator-constructor"></emu-xref>, <emu-xref href="#sec-intl-numberformat-constructor"></emu-xref>, <emu-xref href="#sec-intl-datetimeformat-constructor"></emu-xref> In ECMA-402, 1st Edition, constructors could be used to create Intl objects from arbitrary objects. This is no longer possible in 2nd Edition.
     </li>
+    <li>
+      <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the _length_ property of the function object _F_ was set to 0. In 2nd Edition the value is set to 1.
```
stasm: In order to be more precise, you could say:

    In 2nd Edition _length_ is set to 1.

anba:
LGTM

PR #84, opened 2016-03-25
ECMA-402 v1 legacy constructor semantics compromise

littledan:
This patch addresses #57 by allowing certain legacy constructor
patterns to coexist with the new guarantees in ECMA-402 v2, which
allows for a pattern where all internal slots exist from the
beginning of the object's lifetime. The compromise is based on
storing a "real" object inside of a symbol-named property to allow
for object initialization in cases of the
Intl.<constructor>.call(Object.create(Intl.<constructor>) pattern.
Legacy methods have to forward their calls to this "real" object.

This patch specifies the change for Intl.NumberFormat, but an
analogous change would also be needed for Intl.DateTimeFormat and
Intl.Collator. This version is being sent out for review for feedback
from users and implementors. A sample implementation in V8 can
be found at
https://codereview.chromium.org/1828543007

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -360,6 +360,27 @@
         It is recommended that implementations use the time zone information of the IANA Time Zone Database.
       </emu-note>
     </emu-clause>
+
+    <emu-clause id="sec-unwrapdatetimeformat" aoid="UnwrapDateTimeFormat">
+      <h1>UnwrapDateTimeFormat( dtf )</h1>
+      <p>
+        The UnwrapDateTimeFormat abstract operation gets the underlying DateTimeFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        1. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot and ? InstanceofOperator(_dtf_, %DateTimeFormat%),
+          1. Let _dtf_ be ? RequireObjectCoercible(Get(_dtf_, Intl.[[FallbackSymbol]])).
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        2. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot,
+          1. Throw a *TypeError* exception.
+        1. Return _dtf_.
+      </emu-alg>
+    </emu-clause>
+  </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-intl-datetimeformat-constructor">
@@ -379,7 +400,18 @@
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%DateTimeFormatPrototype%"`, &laquo; [[initializedIntlObject]], [[initializedDateTimeFormat]], [[locale]], [[calendar]], [[numberingSystem]], [[timeZone]], [[weekday]], [[era]], [[year]], [[month]], [[day]], [[hour]], [[minute]], [[second]], [[timeZoneName]], [[hour12]], [[hourNo0]], [[pattern]], [[boundFormat]] &raquo;).
-        1. Return ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+        1. Perform ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+      </emu-alg>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        4. Let _this_ be the *this* value.
+        1. If NewTarget is *undefined* and ? InstanceofOperator(_this_, %NumberFormat%),
+          1. Perform ? DefineOwnPropertyOrThrow(_this_, Intl.[[FallbackSymbol]], { [[Value]]: _dateTimeFormat_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
+          1. Return _this_.
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        1. Return _dateTimeFormat_.
       </emu-alg>
     </emu-clause>
   </emu-clause>
@@ -512,7 +544,7 @@
       <emu-alg>
         1. Let _dtf_ be *this* value.
         1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
-        1. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot, throw a *TypeError* exception.
+        1. Let _dtf_ be ? UnwrapDateTimeFormat(_dtf_).
         1. If _dtf_.[[boundFormat]] is *undefined*, then
           1. Let _F_ be a new built-in function object as defined in DateTime Format Functions (<emu-xref href="#sec-datetime-format-functions"></emu-xref>).
           1. Let _bf_ be BoundFunctionCreate(_F_, _dft_, &laquo; &raquo;).
@@ -545,7 +577,7 @@
       <h1>Intl.DateTimeFormat.prototype.resolvedOptions ()</h1>
 
       <p>
-        This function provides access to the locale and formatting options computed during initialization of the object.
+        This function provides access to the locale and formatting options computed during initialization of the object. This function initially invokes the internal algorithm UnwrapDateTimeFormat to get the %DateTimeFormat% object on which to operate.
       </p>
 
       <p>
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -12,6 +12,10 @@
     The Intl object is not a function object. It does not have a [[Construct]] internal method; it is not possible to use the Intl object as a constructor with the *new* operator. The Intl object does not have a [[Call]] internal method; it is not possible to invoke the Intl object as a function.
   </p>
 
+  <p>
+    The Intl object has an internal slot, [[FallbackSymbol]], which is a new %Symbol% in the current realm.
+  </p>
+
   <emu-clause id="sec-constructor-properties-of-the-intl-object">
     <h1>Constructor Properties of the Intl Object</h1>
 
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -448,6 +448,26 @@
         1. Return _m_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-unwrapnumberformat" aoid="UnwrapNumberFormat">
+      <h1>UnwrapNumberFormat(nf)</h1>
+      <p>
+        The UnwrapNumberFormat abstract operation gets the underlying NumberFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot and ? InstanceofOperator(_nf_, %NumberFormat%),
+          1. Let _nf_ be ? RequireObjectCoercible(Get(_nf_, Intl.[[FallbackSymbol]])).
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        2. If _nf_ does not have an [[initializedNumberFormat]] internal slot,
+          1. Throw a *TypeError* exception.
+        1. Return _nf_.
+      </emu-alg>
+    </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-intl-numberformat-constructor">
@@ -467,7 +487,18 @@
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _numberFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%NumberFormatPrototype%"`, &laquo; [[initializedIntlObject]], [[initializedNumberFormat]], [[locale]], [[numberingSystem]], [[style]], [[currency]], [[currencyDisplay]], [[minimumIntegerDigits]], [[minimumFractionDigits]], [[maximumFractionDigits]], [[minimumSignificantDigits]], [[maximumSignificantDigits]], [[useGrouping]], [[positivePattern]], [[negativePattern]], [[boundFormat]] &raquo;).
-        1. Return ? InitializeNumberFormat(_numberFormat_, _locales_, _options_).
+        1. Perform ? InitializeNumberFormat(_numberFormat_, _locales_, _options_).
+      </emu-alg>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        4. Let _this_ be the *this* value.
+        1. If NewTarget is *undefined* and ? InstanceofOperator(_this_, %NumberFormat%),
+          1. Perform ? DefineOwnPropertyOrThrow(_this_, Intl.[[FallbackSymbol]], { [[Value]]: _numberFormat_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
+          1. Return _this_.
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        6. Return _numberFormat_.
       </emu-alg>
     </emu-clause>
   </emu-clause>
@@ -577,7 +608,7 @@
       <emu-alg>
         1. Let _nf_ be *this* value.
         1. If Type(_nf_) is not Object, throw a *TypeError* exception.
-        1. If _nf_.[[initializedNumberFormat]] is *true*, throw a *TypeError* exception.
+        1. Let _nf_ be ? UnwrapNumberFormat(_nf_);
         1. If _nf_.[[boundFormat]] is *undefined*, then
           1. Let _F_ be a new built-in function object as defined in Number Format Functions (<emu-xref href="#sec-number-format-functions"></emu-xref>).
           1. Let _bf_ be BoundFunctionCreate(_F_, _nf_, &laquo; &raquo;).
@@ -591,7 +622,7 @@
       <h1>Intl.NumberFormat.prototype.resolvedOptions ()</h1>
 
       <p>
-        This function provides access to the locale and formatting options computed during initialization of the object.
+        This function provides access to the locale and formatting options computed during initialization of the object. This function initially invokes the internal algorithm UnwrapNumberFormat to get the %NumberFormat% object on which to operate.
       </p>
       <p>
         The function returns a new object whose properties and attributes are set as if constructed by an object literal assigning to each of the following properties the value of the corresponding internal slot of this NumberFormat object (see <emu-xref href="#sec-properties-of-intl-numberformat-instances"></emu-xref>): locale, numberingSystem, style, currency, currencyDisplay, minimumIntegerDigits, minimumFractionDigits, maximumFractionDigits, minimumSignificantDigits, maximumSignificantDigits, and useGrouping. Properties whose corresponding internal slots have the value *undefined* are not assigned.
--- a/spec/conventions.html
+++ b/spec/conventions.html
@@ -24,6 +24,10 @@
     For ECMAScript objects, this standard may use variable-named internal slots: The notation "[[&lt;_name_&gt;]]" denotes an internal slot whose name is given by the variable name, which must have a String value. For example, if a variable _s_ has the value *"a"*, then [[&lt;_s_&gt;]] denotes the [[&lt;_a_&gt;]] internal slot.
   </p>
 
+  <p>
+    This specification uses blocks demarcated as <span class="normative-optional">Normative Optional</span> to denote the sense of <a href="https://tc39.github.io/ecma262/#sec-additional-ecmascript-features-for-web-browsers">Annex B</a> in ECMA 262. That is, normative optional sections are required when the ECMAScript host is a web browser. The content of the section is normative but optional if the ECMAScript host is not a web browser.
+  </p>
+
   <emu-clause id="sec-402-well-known-intrinsic-objects">
     <h1>Well-Known Intrinsic Objects</h1>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -25,6 +25,20 @@
     margin-bottom: 10px;
     padding-left: 10px;
   }
+
+emu-normative-optional {
+    margin: 1em 0;
+    border-left: 5px solid #ff6600;
+    display: block;
+    background: #ffeedd;
+}
+
+span.normative-optional {
+    padding-left: 5px;
+    text-transform: uppercase;
+    color: #884400;
+}
+
 </style>
 <img src="img/ecma-logo.jpg">
 <pre class=metadata>


spec/numberformat.html
```diff
@@ -346,6 +365,13 @@
 
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
+        1. Let _this_ be the receiver.
+        1. NOTE: The following step and nested steps implement legacy compatibility semantics
```
caridy: this is the first time I see an arbitrary link in an ecma spec, do we really need to link to the bug in the algo?

spec/numberformat.html
```diff
@@ -346,6 +365,13 @@
...
+           for ECMA-402 v1. See <a href="https://github.com/tc39/ecma402/issues/57">this bug</a>
+           for details.
+        1. If _this_ is not *undefined*, and ? InstanceofOperator(_this_, %NumberFormat%),
+          1. Perform ? DefineOwnPropertyOrThrow(_this_, Intl.[[FallbackSymbol]], { [[Value]]: Construct(%NumberFormat%, locales, options), [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
```
caridy: `_locales_, _options_`

spec/numberformat.html
```diff
@@ -328,6 +328,25 @@
...
+
+    <emu-clause id="sec-unwrapnumberformat" aoid="UnwrapNumberFormat">
+      <h1>UnwrapNumberFormat(nf)</h1>
+      <p>
+        The UnwrapNumberFormat abstract operation gets the underlying NumberFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-alg>
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot,
```
caridy: all new internally slots in camel case: `[[InitializedNumberFormat]]`

caridy:
LGTM

spec/numberformat.html
littledan: Just trying to be useful by including meaningful non-normative notes, but maybe we can say people should find this information via git blame instead.

spec/numberformat.html
littledan: This isn't a new internal slot, this is an existing one.

spec/numberformat.html
littledan: Fixed

littledan:
Before this is committed, we need feedback on whether this patch ensures compatibility with old Intl.js versions and other uses of ECMA 402. It also needs to be ported to the other two relevant constructors, which I'll do as soon as I hear back about the compatibility of this change.

littledan:
I tested this patch on a website which uses an old version of Intl.js. I observed that the v2 semantics led to a console message being printed indicating a failure to format, whereas the v1 semantics and those in this patch seemed to load successfully. Does anyone else have feedback before I port these semantics to the other cases?

littledan:
@tschneidereit @thetalecrafter @bterlson

thetalecrafter:
Given that the known incompatible library only operated on `NumberFormat` and `DateTimeFormat`, could we leave out applying the compatibility fix to `Collator`?

allenwb:
@thetalecrafter 
+1 WRT `Collator`

@caridy @littledan 
Could we add this as an "Annex B" within Ecma-402?  It seems like something we can get rid of as Intl.js falls out of usage.

tschneidereit:
CC @jswalden as our expert on Intl.

littledan:
OK, I tested a version which just does NumberFormat and constructor (same link), with code structured so it should be easy to turn off later, and it worked on the broken-in-v2 page. I uploaded a spec version which just applies to those. What do you think?

littledan:
Re: Annex B. Could we have some kind of emu formatting to put the "historical web compat" parts in-line rather than out-of-line in an appendix (@bterlson @domenic)? I think this would make things easier for implementors and users if they didn't have to look way at the end, given how this is actually a little invasive. I agree that it could be removed after a number of years, but not yet.

By the way, Intl.js has been updated to work with the v2 spec, and I don't know of any serious competitors; I think we're just waiting for old versions to fall out of use.

jswalden:
This in principle (modulo one aspect of the change that I'm probably misreading, detail to follow) seems reasonable enough to me.  I'm definitely with the people arguing for putting this in some sort of temporary-compatibility holding pattern for later removal.  The subclassing tactic `Intl` used to use is a wart upon the proposal and the spec, that we can and should endeavor to remove, even if it's not an immediately-possible thing.

spec/numberformat.html
```diff
@@ -328,6 +328,25 @@
...
+          1. If ? InstanceofOperator(_nf_, %NumberFormat%),
```
jswalden: I'm probably misreading this, but are you sure this shouldn't be "If _not_ ? InstanceofOperator(_nf_, %NumberFormat%),"?  It seems to me that if _nf_ is a NumberFormat, we should be returning it.  Only in the case where _nf_ isn't a NumberFormat, should we look at _nf_[`Intl.[[FallbackSymbol]]`].

And a similar change needs to be made, it seems to me, to the `Intl.NumberFormat` algorithm.

And similar changes would need to be made to the `DateTimeFormat` patch.

Again, probably misreading this, but I've been staring at it for a good ten minutes and can't parse it as anything other than doing nothing special if the object being acted upon is a non-`NumberFormat`.

jswalden: Ah!  Till explained how I was being stupid in misreading this elsewhere.  Carry on, this change looks good!

caridy:
@allenwb, @littledan Let's discuss the course of action for the spec notes and later removal during tomorrow's meeting, we have an agenda item for this.

littledan:
At TC39 today, we got to consensus on including this proposal, as long as it's marked in the spec as normative-optional and something that we'd hope to remove if/when possible. This marking may be done inline if it makes the resulting spec easier to follow, as long as the affected sections are indeed appropriately marked. (@erights, could you verify this interpretation of the conclusion?) (@bterlson any suggestions for how to mark it as such, or should I cook something up?)

spec/datetimeformat.html
```diff
@@ -296,6 +296,26 @@
...
+
+    <emu-clause id="sec-unwrapdatetimeformat" aoid="UnwrapDateTimeFormat">
+      <h1>UnwrapDateTimeFormat(nf)</h1>
+      <p>
+        The UnwrapDateTimeFormat abstract operation gets the underlying DateTimeFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-alg>
+        1. If _nf_ does not have an [[initializedDateTImeFormat]] internal slot,
```
ericf: Typo: `initializedDateTImeFormat` --> `initializedDateTimeFormat`

erights:
>  (@erights, could you verify this interpretation of the conclusion?)

Yes. Thanks for checking.

ericf:
For posterity, the code that depended on the ECMA402 v1 constructor behavior was in the [`intl-format-cache`](https://www.npmjs.com/package/intl-format-cache) package (not the [`Intl.js`](https://www.npmjs.com/package/intl) polyfill package) and was fixed in a SemVer patch release ~4 months ago.

Last week I was able to supply @littledan with a site which is still using an older version of the `intl-format-cache` package so that he could verify that without this change the site breaks, and when applying this change the site works as expected.

littledan:
This proposal reached consensus at the March 2016 TC39 meeting.

littledan:
The remaining change needed it so make it clear that the text here needs to be normative-optional. At TC39, there was consensus that this can be done in-line, by putting a box around the optional parts, rather than out-of-line as in Annex B in ECMA 262, as long a the denotation of normative-optional was the same. @caridy signed up for making this editorial change.

littledan:
@caridy What's the status of this change? Need anything from me?

caridy:
@littledan I will work with @bterlson on the normative-optional representation, and then we can tackle this one.

zbraniecki:
What's the status of this one? Both V8 and SM are failing test262 init tests and it seems that we're waiting for this to be resolved before we fix them.

littledan:
@zbraniecki I think we are blocking on tooling work. @bterlson @caridy are either of you thinking of doing these tooling changes? If not, I can make a patch for a 262-style out-of-line Annex B, though the consensus in committee was for an inline normative-optional change if I remember correctly. cc @tcare

jungshik:
@littledan, did the consensus @TC39 include having this added to Collator as well? As @thetalecrafter  mentioned here (and I mentioned in #57), Collator does not need to have a legacy-compat provision, does it? 

BTW,  test262 has the following 3 tests for v1 behavior:  intl402/Collator/10.1.2.1_4 , intl402/DateTimeFormat/12.1.2.1_4 and  intl402/NumberFormat/11.1.2.1_4 (as far as I can tell).  Would it be ok to just drop them for now? ( 1[0-2].1.2.1_4 all refer to the part of Ecma 402 v1 spec removed/revised in v2 spec).

caridy:
@littledan @jungshik during last meeting we have decided that we will use the html and css directly on the spec text instead of adding some new stuff to the tooling. @littledan will send the proper PR at some point before next meeting.

As for the collator and the tests questions, I think that makes sense, but I will let @littledan to confirm.

littledan:
I've updated the pull request to use inline, explicitly marked normative optional text. This doesn't have ecmarkup support yet, but it's done in a way that's analogous to emu-note, so it should be easy to put it in. What do you think?

caridy:
thanks @littledan, I will review it soon.

spec/datetimeformat.html
```diff
@@ -360,6 +360,27 @@
         It is recommended that implementations use the time zone information of the IANA Time Zone Database.
       </emu-note>
     </emu-clause>
+
+    <emu-clause id="sec-unwrapdatetimeformat" aoid="UnwrapDateTimeFormat">
+      <h1>UnwrapDateTimeFormat( dtf )</h1>
```
caridy: _dtf_

spec/datetimeformat.html
```diff
@@ -360,6 +360,27 @@
...
+      <p>
+        The UnwrapDateTimeFormat abstract operation gets the underlying DateTimeFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        1. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot and ? InstanceofOperator(_dtf_, %DateTimeFormat%),
```
caridy: ...` is equal *true*, then`

caridy: in general I haven't seen the use of ? in conditions... but I suspect that's ok.

spec/datetimeformat.html
```diff
@@ -360,6 +360,27 @@
...
+          1. Let _dtf_ be ? RequireObjectCoercible(Get(_dtf_, Intl.[[FallbackSymbol]])).
```
caridy: split this into two lines for readability.

spec/datetimeformat.html
```diff
@@ -360,6 +360,27 @@
...
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        2. If _dtf_ does not have an [[initializedDateTimeFormat]] internal slot,
```
caridy: `1.` instead of `2.`

caridy: ...`, then`

spec/numberformat.html
```diff
@@ -448,6 +448,26 @@
         1. Return _m_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-unwrapnumberformat" aoid="UnwrapNumberFormat">
+      <h1>UnwrapNumberFormat(nf)</h1>
```
caridy: _nf_

spec/numberformat.html
```diff
@@ -448,6 +448,26 @@
...
+      <p>
+        The UnwrapNumberFormat abstract operation gets the underlying NumberFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot and ? InstanceofOperator(_nf_, %NumberFormat%),
+          1. Let _nf_ be ? RequireObjectCoercible(Get(_nf_, Intl.[[FallbackSymbol]])).
```
caridy: split...

spec/numberformat.html
```diff
@@ -448,6 +448,26 @@
...
+      </emu-alg>
+      </div></emu-normative-optional>
+      <emu-alg>
+        2. If _nf_ does not have an [[initializedNumberFormat]] internal slot,
```
caridy: `1.`

littledan: Starting at 2 is deliberate since it follows earlier algorithm steps.

littledan:
I'll get back to you with these changes by January 30th, if that's in time. There are also some formatting implementation changes that I plan to do to bring things in line with what I've done in another repository. I'd be interested in feedback from others about whether this formatting is acceptable from a spec perspective, cc @bterlson @allenwb

caridy:
@erights this it the PR that @littledan mentioned when talking about Annex B, please review it.

erights:
this is not what i thought i was volunteering to review. i thought i was volunteering to review language about what annex b means. did i misunderstand?

littledan:
Thanks for this!

caridy:
thanks @anba, I can take care of this.

spec/datetimeformat.html
```diff
@@ -379,7 +400,18 @@
...
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%DateTimeFormatPrototype%"`, &laquo; [[initializedIntlObject]], [[initializedDateTimeFormat]], [[locale]], [[calendar]], [[numberingSystem]], [[timeZone]], [[weekday]], [[era]], [[year]], [[month]], [[day]], [[hour]], [[minute]], [[second]], [[timeZoneName]], [[hour12]], [[hourNo0]], [[pattern]], [[boundFormat]] &raquo;).
-        1. Return ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+        1. Perform ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+      </emu-alg>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        4. Let _this_ be the *this* value.
+        1. If NewTarget is *undefined* and ? InstanceofOperator(_this_, %NumberFormat%),
+          1. Perform ? DefineOwnPropertyOrThrow(_this_, Intl.[[FallbackSymbol]], { [[Value]]: _dateTimeFormat_, [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *false* }).
```
caridy: That is covered by `InstanceofOperator(_this_, %NumberFormat%)` in the previous step I think.

anba: Not if `Intl.NumberFormat[Symbol.hasInstance]` was redefined to add a custom `instanceof` behaviour, like in this test: https://github.com/mozilla/gecko-dev/blob/3ed98bee2a88a4d953d22520e727344bf8e617b8/js/src/tests/Intl/NumberFormat/call.js#L62-L85

spec/datetimeformat.html
caridy: I went with the first option because I'm not sure about the primitive values. @littledan can you double check this?

caridy: fixed by 7eb2c2b149ede6c192d9ba683191b7f8d7d05158

spec/datetimeformat.html
caridy: fixed by 7eb2c2b149ede6c192d9ba683191b7f8d7d05158

spec/datetimeformat.html
caridy: fixed by 7eb2c2b149ede6c192d9ba683191b7f8d7d05158

spec/datetimeformat.html
```diff
@@ -379,7 +400,18 @@
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%DateTimeFormatPrototype%"`, &laquo; [[initializedIntlObject]], [[initializedDateTimeFormat]], [[locale]], [[calendar]], [[numberingSystem]], [[timeZone]], [[weekday]], [[era]], [[year]], [[month]], [[day]], [[hour]], [[minute]], [[second]], [[timeZoneName]], [[hour12]], [[hourNo0]], [[pattern]], [[boundFormat]] &raquo;).
-        1. Return ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+        1. Perform ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
+      </emu-alg>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        4. Let _this_ be the *this* value.
+        1. If NewTarget is *undefined* and ? InstanceofOperator(_this_, %NumberFormat%),
```
caridy: fixed by 7eb2c2b149ede6c192d9ba683191b7f8d7d05158

spec/intl.html
```diff
@@ -12,6 +12,10 @@
     The Intl object is not a function object. It does not have a [[Construct]] internal method; it is not possible to use the Intl object as a constructor with the *new* operator. The Intl object does not have a [[Call]] internal method; it is not possible to invoke the Intl object as a function.
   </p>
 
+  <p>
+    The Intl object has an internal slot, [[FallbackSymbol]], which is a new %Symbol% in the current realm.
```
caridy: @littledan can you comment here?

spec/numberformat.html
```diff
@@ -448,6 +448,26 @@
...
+    <emu-clause id="sec-unwrapnumberformat" aoid="UnwrapNumberFormat">
+      <h1>UnwrapNumberFormat(nf)</h1>
+      <p>
+        The UnwrapNumberFormat abstract operation gets the underlying NumberFormat operation
+        for various methods which implement ECMA-402 v1 semantics for supporting initializing
+        existing Intl objects.
+      </p>
+      <emu-normative-optional><span class="normative-optional">Normative Optional</span><div class="normative-optional-contents">
+      <emu-alg>
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot and ? InstanceofOperator(_nf_, %NumberFormat%),
```
caridy: fixed by 7eb2c2b149ede6c192d9ba683191b7f8d7d05158

spec/intl.html
littledan: Would undefined be a reasonable \[\[Description]] value here? In V8, I used "IntlFallback", but for no good reason. There's no way that you can access this symbol as the property of something else, and I don't see a reason to make one, so "Intl.fallback" or "Symbol.intlFallback" would not be all that compelling in my opinion.

spec/datetimeformat.html
littledan: That was all unintentional, but I agree with @caridy's reading.

caridy: @anba will something like this be sufficient? 

image

or should we just throw a TypeError if this-value is not an object before calling DefineOwnPropertyOrThrow?

littledan: @caridy I don't see why that's necessary; I'm fine with this throwing due to InstanceOf.

caridy: Honestly, I feel the same @littledan, but @anba seems to be convinced about a potential issue here, and I trust him :)

littledan: Oh, I missed the above comment--yeah, the text you wrote sounds good and (I think) it should guard against it.

caridy: @anba can you confirm?

anba: Both solutions (guard or throwing) work, and I'm fine with either of them. 😄

spec/intl.html
anba: I assume @jswalden asked about adding a description for debugging purposes, I can try to catch him later on IRC, so he can chime in.

caridy: solved by 44107b64949017960c905ebb876216a2075973e7

allenwb: > The Intl object has an internal slot, [[FallbackSymbol]], which is a new %Symbol% in the current realm.

In this case, shouldn't something like %FallbackSymbol% be added to Table 1 and the accesses performed like: Get(%Intl%,%FallbackSymbol)? Basically, internal slots are not identified using symbols.  It either needs to be an internal slot named FallBackSymbol or a own property whose key is %FallBackSymbol%. You should be muddling these distinct concepts.

littledan: I see no problem with putting it as %FallbackSymbol%, but I also don't quite understand the problem with the current phrasing. When I wrote it out with this internal slot notation, I was picturing it as being "in a separate namespace"--it's just an internal slot that is called \[\[FallbackSymbol]] which contains a symbol. Is the problem that this is a one-off internal slot, rather than a name that's reused over a class of objects?

allenwb: Ah, I misunderstood what you were trying to say (which is the root problem).  I thought you were saying that the "new  %Symbol%" was the key/name of the internal slot.  If the value is a realm specific intrinsic symbol, it probably needs to be added to Table 1 (of this spec).  Note that in Ecma-262 we don't have any realm-specific intrinsic symbols so rather than appearing in the corresponding Table 6 (Well known intrinsic objects) intrinsic symbols are listed in Ecma-262 Table 1 (Well known symbols).

We should probably also refer it the symbol as @@FallbackSymbol.

So, the spec statement could be: "The Intl object has an internal slot, [[FallbackSymbol]] whose value is initialized to  @@FallbackSymbol

littledan: This is an interesting point about being realm-specific vs cross-realm. In V8, I ended up implementing it as a cross-realm symbol, actually. @anba What did you do for SpiderMonkey? Do you have an opinion about which way it should be?

anba: > What did you do for SpiderMonkey? Do you have an opinion about which way it should be?

It's currently implemented as a realm-specific symbol in SpiderMonkey. Hm, if it isn't a realm-specific symbol, I wonder if we should remove the InstanceofOperator call in the unwrapping operations, so that a cross-realm object initialized as a Intl.DateTimeFormat/NumberFormat object works the same way as a proper cross-realm Intl.DateTimeFormat/NumberFormat object.  (`Intl.NumberFormat.prototype.resolvedOptions.call(crossRealmNumberFormat)` currently works, but `Intl.NumberFormat.prototype.resolvedOptions.call(crossRealmObjectInitializedAsNumberFormat)` throws a TypeError.)

erights:
If the kludge serves its purpose either way, I would prefer the realm specific approach as that narrows the visibility of the kludge.

littledan:
How does realm-specific make the kludge narrower? I don't see how cross-realm vs realm-specific differ in terms of breadth.

allenwb:
It seems likely that the use of a per realm symbol will mean that passing an object that actually uses the fallback may fail with the object is passed between realms.  Of course, the use of InstanceOfOperator as part of the prerequisite for setting the fallback also seems problematic in cross-realm situations.  Note that the this value (presumably usually Intl) passed to Intl constructors is not necessarily from the same realm as the actual constructor function that was called. 

Cross-realm well known symbols are generally used when we want objects that depend upon symbol based hooks to work consistently when they are passed between realms.

It's still not clear to me why a symbol accessed property is needed. As far as I can tell the actual symbol isn't explicitly exposed to userland. Under what circumstance would usernd code need to use the symbol to access the associated property. If seems like an internal slot would be a better place to capture the fallback object.

littledan:
The purpose of this proposal is to make it so that you can take an object that was initialized with none of these Intl internal slots and then make it act like a DateTimeFormat or NumberFormat instance, without adding an internal slot to an existing object. You can get at this symbol from JS in a way that @anba's tests show:

```
Object.getOwnPropertySymbols(Intl.NumberFormat.call(Object.create(Intl.NumberFormat.prototype)))[0]
```

erights:
The purpose of the proposal is to introduce the minimal kludge we can to allow old-bad-code-that-we-dare-not-break to not break while doing all we can to discourage new code from needlessly relying on the presence of this kludge. If too much new code accidentally comes to rely on the kludge anyway, we'll never be able to kill it. Otherwise, we might. We have several such successes behind us already.

allenwb:
Ah, I missed the application of the constructor to an arbitrary object part. 

Clearly, you can get at the symbol that way.  Isn't that a problem?  It opens the compatibility hack up to various misuses or even attacks.  Wouldn't it be better to use a "soft field" to associate the fallback value with an arbitrary object?  EG, an specification level cross-realm weak map like registry (of course, left to implementations to decide how to actually represent it). There would be no exposure to userland tampering and no cross-realm issues (except that the instanceof tests are still a problem).

littledan:
How would this differ in observable semantics from the v1 constructor semantics?

littledan:
This would be readily implementable in V8. We'd just use our private symbol functionality (which is inadmissible to spec since Proxies would see private symbol access, but would not see WeakMap access), rather than an ordinary symbol.

If I understand correctly (please correct me if I'm wrong), basically you're suggesting that we go back to the v1 semantics, but phrased differently, and with the scope restrictions we worked out above (only for NumberFormat and DateTimeFormat, and only if the thing you're constructing `instanceof` that constructor, plus the object checks recommended by @anba).

I guess the only downside I see is debugging. With a public symbol, you can see how it's represented, but with a WeakMap, it's secret internal state. On the other hand, there's lots of other stuff that's in internal state; this wouldn't be particularly different.

Given that this is a bug fix, it might not be too late to squeeze it into ES2017. What do people think? I don't have a particular objection to it.

allenwb:
> If I understand correctly (please correct me if I'm wrong), basically you're suggesting that we go back to the v1 semantics, but phrased differently, and with the scope restrictions we worked out above (only for NumberFormat and DateTimeFormat, and only if the thing you're constructing instanceof that constructor, plus the object checks recommended by @anba).

Basically yes. The V1 semantics can be implemented on any platform that hosts some sort of soft field capability and intl methods that use the fallback protocol. What we want too make normatively optional is the ability to setup a fallback for ordinary, non-formatter objects. 

I don't think we should be using instanceof to determine if we need to setup a fallback or in unwrapping because of the cross realm issues relating to instanceof.  Instead of instanceof, internal slot existence tests should suffice for determining whether a constructor needs to setup a fallback object or wither unwrapping needs to be done.

littledan:
I don't see how internal slot existence checks could work. We need to be able to "instantiate" an object that was created like `Object.create(Intl.NumberFormat.prototype)`.

allenwb:
> don't see how internal slot existence checks could work. We need to be able to "instantiate" an object that was created like Object.create(Intl.NumberFormat.prototype).

Allowed by using Intl.NumberFormat to initialize the object, right?

NumberFormat checks if its this value has one or more of the internal slots of a NumberFormat instance.  If not, it instantiates a fallback NumberFormat object and associates it with the this value using some implementation specific soft field mechanism.

littledan:
The purpose of the instanceof check is to *narrow* the set of things that we'll go back and instantiate. The narrowing isn't driven by implementation or compatibility concerns, but by design goals from TC39 normative-optional practice to make compatibility fixups as narrow as possible. We could also go to v1 semantics, which didn't have this narrowing.

The UnwrapNumberFormat operation already does not get the fallback object for objects which are actual NumberFormat instances with the internal slot; the constructor does not do this check, but I would be fine with adding it for further narrowing, but it would only change behavior in extremely obscure test cases like:
```js
  Intl.NumberFormat.call(Intl.NumberFormat())
```

What the `instanceof` check does is make sure that we don't instantiate calls like this as a NumberFormat pseudo-instance:

```
  Intl.NumberFormat.call({})
```

Issue #88, opened 2016-05-03
Intl.DateTimeFormat should localize era names as well


watilde:
In [annex-implementation-dependent-behaviour](https://tc39.github.io/ecma402/#annex-implementation-dependent-behaviour), it is mentioned that `DateTimeFormat` should localize era names: [annexes.html#L102](https://github.com/tc39/ecma402/blob/a47b02e1b1b83c2df4e33925d372fc68bb84219d/spec/annexes.html#L102), but in the part of [12.1.6FormatDateTime](https://tc39.github.io/ecma402/#sec-formatdatetime) linked from the above, there is no description about the localized era names. Probably it can be bug.

The description to resolve it may can be like this:
```
If _p_ is *"era"*, then the String value may also depend on whether _dateTimeFormat_ has a [[era]] internal slot
```

Related links: https://github.com/andyearnshaw/Intl.js/issues/164, https://github.com/andyearnshaw/Intl.js/pull/166

Thanks.

caridy:
merged as b0c9b6bd834780b74d4efe9448a16f19bcbc590b

littledan:
Did we get any implementations of this, besides Intl.js? Are there any tests in test262?

jungshik:
Era is localized in v8, I  believe because cldr has data for them.

PR #89, opened 2016-05-07
Intl.DateTimeFormat should localize era names as well

watilde:
Related issue: https://github.com/tc39/ecma402/issues/88


caridy:
LGTM.

littledan:
Has anyone written test262 tests for this change?

caridy:
I was planning to do so, or get @watilde (who has been very responsive to help). Additionally, we can get implementers to contribute back their existing tests since `era` is already supported in Chrome, FF and IE/Edge.

watilde:
Wow, I should add the test case to test262 at the same time when I wrote this change. Sure thing, I will try to update the test case to cover this patch!

PR #94, opened 2016-07-14
Move Array.prototype.toLocaleString's separator-computation step to just before the loop that first uses it

jswalden:
I was reviewing a SpiderMonkey patch that put the step in its current order, and I was going to tell the author to move the step out of order, then I realized I could just change the spec.  :-)  (Theoretically we could bikeshed the step even one later, but I prefer to view "Let k be 1" and the loop on k as a single logical step.)

I'm also filing a PR against ecma262 for the same issue there.


allenwb:
see https://github.com/tc39/ecma262/pull/636#issuecomment-232761741

jswalden:
The argument there makes sense, closing.

PR #96, opened 2016-07-26
Editorial: use dot notation for accessing internal slots & co.

caridy:
## use dot notation for accessing internal slots

This is a follow up from https://github.com/tc39/ecma262/pull/591

Specifically, change:

```
   the [[Foo]] internal slot of _O_  (132 occurrences)
```
and
```
   _O_'s [[Foo]] internal slot (265 occurrences)
```
to
```
   _O_.[[Foo]]
```

while preserving some references when using "Internal Slot" in a sentence to test for identity, and removing "the value of" wording in front.

## Length default property of functions is 0

This is a follow up from https://github.com/tc39/ecma262/pull/597

* removing unnecessary definitions

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -48,7 +48,7 @@
       </p>
 
       <emu-alg>
-        1. If _collator_ has an [[initializedIntlObject]] internal slot with value *true*, throw a *TypeError* exception.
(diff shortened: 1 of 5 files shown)


srl295:
looks good…

zbraniecki:
lgtm!

PR #100, opened 2016-08-14
FormatToParts: aggregation of PR #64 and PR #79

caridy:
DateTimeFormat.prototype.formatToParts()
* https://github.com/tc39/ecma402/pull/64

NumberFormat.prototype.formatToParts(): 
* https://github.com/tc39/ecma402/pull/79

These features are Stage 4, and are ready to be merged.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -265,7 +265,6 @@
             1. Add new part record { [[type]]: *"literal"*, [[value]]: _literal_ } as a new element of the list _result_.
           1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
           1. If _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then:
-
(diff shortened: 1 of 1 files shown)


caridy:
closes #79 and closes #64

zbraniecki:
@caridy - I think you skipped in your aggregation the whole part of the NumberFormat patch that changed `PartitionNumberPattern` algo to return a `List`.

caridy:
hmm, where was that?

zbraniecki:
@caridy Look at the algo in the merged spec: https://github.com/tc39/ecma402/blob/master/spec/numberformat.html#L114-L148 vs the algo in the PR: https://github.com/stasm/ecma402/blob/number-format-to-parts/spec/numberformat.html#L114-L199

I didn't check if the rest of aggregated PR matches the orignal PRs, but this is definitely missing and makes the spec broken because callers expect this algo to return parts now, while in the spec it returns a string.

caridy:
I will review in details tomorrow.

PR #103, opened 2016-08-30
Normative: Require the latest available Unicode version

mathiasbynens:
…instead of a fixed version number.

Ref. https://github.com/tc39/ecma262/pull/620.


caridy:
+1

PR #104, opened 2016-09-01
Separate out FormatNumberString operation

zbraniecki:
This will allow us to not have to initialize NumberFormat in PluralRules - https://github.com/tc39/proposal-intl-plural-rules/pull/15

@caridy - can you review? I'm wondering if we should name the first parm `options` instead of `numberFormat`.

@eemeli - does it look good to you?

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -103,6 +103,23 @@
       </p>
     </emu-clause>
 
+    <emu-clause id="sec-formatnumberstring" aoid="FormatNumberToString">
+      <h1>FormatNumberToString ( _numberFormat_, _x_ )</h1>
+
+      <p>
+        The FormatNumberToString abstract operation is called with arguments _numberFormat_ (which must be an object with fields minimumSignificantDigits, maximumSignificantDigits, minimumIntegerDigits, minimumFractionDigits and maximumFractionDigits), and _x_ (which must be a Number value), and returns _x_ as a string value with digits formatted according to the 5 formatting parameters.
+      </p>
+
+      <emu-alg>
+        1. Assert: _numberFormat_.[[initializedIntlObject]] is true.
+        1. If the _numberFormat_.[[minimumSignificantDigits]] and _numberFormat_.[[maximumSignificantDigits]] are present, then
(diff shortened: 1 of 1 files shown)


eemeli:
LGTM

spec/numberformat.html
```diff
@@ -103,6 +103,22 @@
       </p>
     </emu-clause>
 
+    <emu-clause id="sec-formatnumberstring" aoid="FormatNumberString">
+      <h1>FormatNumberString ( _numberFormat_, _x_ )</h1>
```
caridy: I don't particular like this name, because it tells me nothing. Maybe `FormatNumberToString`

spec/numberformat.html
```diff
@@ -103,6 +103,22 @@
...
+
+      <p>
+        The FormatNumberString abstract operation is called with arguments _numberFormat_ (which must be an object with fields minimumSignificantDigits, maximumSignificantDigits, minimumIntegerDigits, minimumFractionDigits and maximumFractionDigits), and _x_ (which must be a Number value), and returns _x_ as a string value with digits formatted according to the 5 formatting parameters.
```
caridy: I normally prefer assertions in the algo than just here in the description of the algo, but I'm ok with that for now.

spec/numberformat.html
```diff
@@ -103,6 +103,22 @@
...
+      </p>
+
+      <emu-alg>
+        1. If the _numberFormat_.[[minimumSignificantDigits]] and _numberFormat_.[[maximumSignificantDigits]] are present, then
```
caridy: I believe we should also assert that the first argument is a ready instance of intl, e.g.:

```
Assert: _numberFormat_.[[initializedIntlObject]] is true.
```

caridy: since all records with `[[initializedIntlObject]]` are formatters, the rest of the assertion is what you have in the header.

spec/numberformat.html
zbraniecki: changed  in the updated patch

spec/numberformat.html
zbraniecki: leaving it for now

spec/numberformat.html
zbraniecki: added the assertion

PR #114, opened 2016-11-25
Add mnfdDefault to SetNumberFormatDigitOptions

zbraniecki:
I'm implementing the `SetNumberFormatDigitOptions` as part of the https://bugzilla.mozilla.org/show_bug.cgi?id=1270146 and noticed that this regresses `NumberFormat`.

What happens is that if the user doesn't manually pass `minimumFractionDigits` option, `minimumFractionDigits` doesn't have the default value in `SetNumberFormatDigitOptions` and `mnfd` ends up being `undefined`.

Then, `mnfd` is passed as the third argument to `GetNumberOption` for `maximumFractionDigits` and that fails the assertion that the third argument is a number.

This patch fixes the bug and also shifts the order of operations in `SetNumberFormatDigitOptions` to better replicate previous behavior with regards to `*SignificantDigits` - the two values were not set at all on the `intlObj` if they were not passed in `options`.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -5,7 +5,7 @@
     <h1>Abstract Operations For NumberFormat Objects</h1>
 
     <emu-clause id="sec-setnfdigitoptions" aoid="SetNumberFormatDigitOptions">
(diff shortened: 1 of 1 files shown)


caridy:
LGTM, let's merge this patch.

PR #116, opened 2016-12-15
Invoke `SetNumberFormatDigitOptions` as fallible in `InitializeNumberFormat`

jswalden:
SetNumberFormatDigitOptions is fallible when provided a user-controllable options object, so invoke it as such.  (This change is also needed in [caridy/intl-plural-rules-spec](https://github.com/caridy/intl-plural-rules-spec), if the person who takes this has access to that spec as well.)


caridy:
@zbraniecki can you update the plural rule spec as well with this fix? I'm planning to do a clean up of the spec before the next meeting.

PR #118, opened 2016-12-17
Consistently use the name `style` for the variable denoting format style in `InitializeNumberFormat`

jswalden:
Looks like someone decided *style* was a good name, then only half-converted things, I guess?  I could also make the name *s* if desired, but the longer name seems slightly better (except to the extent that prior spec versions used *s* as the name).


zbraniecki:
lgtm!

PR #120, opened 2017-01-12
Editorial Changes to Align with 262

caridy:
* upgrading ecmarkup
* clear separation between "internal slots", "record fields" and "key properties", just because they all have the same name does not mean they are the same, or should be the same, this is needed to align with the InternalSlots guidelines from 262
* rename all internal slots to match 262

This PR does not modify any logic or semantic in any way, and hopefully it will go smooth. Please, take time to review all the details in case I have missed something.


spec/collator.html
```diff
@@ -73,23 +76,22 @@
           1. If the string given in the Type column of the row is *"boolean"* and value is not *undefined*, then
             1. Let _value_ be ! ToString(_value_).
           1. Set _opt_.[[&lt;_key_&gt;]] to _value_.
-        1. Let _relevantExtensionKeys_ be %Collator%.[[relevantExtensionKeys]].
-        1. Let _r_ be ResolveLocale(%Collator%.[[availableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
-        1. Set _collator_.[[locale]] to _r_.[[locale]].
+        1. Let _relevantExtensionKeys_ be %Collator%.[[RelevantExtensionKeys]].
+        1. Let _r_ be ResolveLocale(%Collator%.[[AvailableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
+        1. Set _collator_.[[Locale]] to _r_.[[locale]].
```
zbraniecki: Shouldn't the field on `_r_` also be `Locale`?

caridy: the problem with ResolveLocale is that it creates an arbitrary record with fields, and we use that to match the keys provided via `options`. We could attempt to do the same, but those are not internal slots, just fields in a record. I think we can keep them like that. I know it is confusing though.

rwaldron:
👏

spec/collator.html
```diff
@@ -73,23 +76,22 @@
...
           1. If _key_ is *"co"*, then
-            1. Let _property_ be *"collation"*.
             1. Let _value_ be _r_.[[co]].
             1. If _value_ is *null*, let _value_ be *"default"*.
+            1. Set _collator_.[[Collation]] to _value_.
           1. Else use the row of <emu-xref href="#table-collator-options"></emu-xref> that contains _key_ in the Key column:
-            1. Let _property_ be the name given in the Property column of the row.
             1. Let _value_ be _r_.[[&lt;_key_&gt;]].
             1. If the name given in the Type column of the row is *"boolean"*, let _value_ be the result of comparing value with *"true"*.
-          1. Set _collator_.[[&lt;_property_&gt;]] to _value_.
```
rwaldron: So happy to get rid of these.

rwaldron:
LGTM!

spec/collator.html
caridy: It took some time, and some big refactors, but we are getting there :). Still using it in a couple of places where a record with fields is created and matching some options keys, but it is at least narrowed down to one or two abstracts at the moment.

littledan:
If you're changing the internal names to upper case, should the `resolvedOptions()` definition be updated? I don't see an update. Given that there will be a mismatch in the names, maybe the definitions could be rewritten in a more algorithmic style, to explicitly create the object and add the individual properties based on the fields.

caridy:
@littledan I have created a new issue https://github.com/tc39/ecma402/issues/124 to solve that in a different patch to eliminate the usage of `[[<>]]` entirely. Are we good to merge this?

spec/collator.html
```diff
@@ -15,28 +15,31 @@
...
             <td>kn</td>
+            <td>[[Numeric]]</td>
             <td>numeric</td>
             <td>"boolean"</td>
             <td></td>
           </tr>
           <tr>
             <td>kf</td>
+            <td>[[CaseFirst]]</td>
             <td>caseFirst</td>
```
littledan: Editorial suggestion: now that this is just a property name and not also an internal slot, maybe it should be written as `"caseFirst"` rather than caseFirst.

littledan:
Sorry for the slow response, thanks for fixing up all these issues, retroactive LGTM.

spec/collator.html
caridy: @littledan e4eaa7237cdb09da0139a2b9377e0a6d795ee69f

PR #135, opened 2017-03-17
Support hc extension key in DateTimeFormat.

### LINKED ISSUE #105: Complete the extension keys + options + resolvedOptions set
This is the first part of the spin-off from #68.

Currently, there are two ways of providing parameters into Intl APIs: via extension keys on locale tag and via `options`.
There is also one way to test which values has been assigned to the options: `resolvedOptions`.

The issue is that the list of extension keys only partially overlap with the list of options. Sometimes there is an option, but we ignore the available extension key, sometimes we don't have an option and only accept the param via extension key, sometimes we allow both.

In general, my recommendation is to get to the point where all params that are available through extension keys were also available through options, and options would take precedence.

Here's the break out per API:
# Collator
- Extension Keys:
  - co
  - kn
  - kf
- Options:
  - numeric
  - caseFirst
  - (new) collation
- resolvedOptions:
  - collation
  - numeric
  - caseFirst

This one is the simplest. We already allow for `kn` and `kf` and their counterparts in options. The only missing option is `collation`.
# NumberFormat
- Extension Keys:
  - nu
  - (new) cu
- Options:
  - currency
  - (new) numberingSystem
- resolvedOptions:
  - numberingSystem
  - currency

Here we have extension keys supporting `nu` but not `cu` and options that support the reverse.
Adding `cu` to extension keys and `numberingSystem` to options gives us complete coverage.
# DateTimeFormat
- Extension Keys:
  - nu
  - ca
  - (new) hc
  - (new) tz
- Options:
  - timeZone
  - (new) calendar
  - (new) hourCycle
  - (new) numberingSystem
  - (deprecate) hour12
- resolvedOptions:
  - timeZone
  - numberingSystem
  - calendar
  - (new) hourCycle
  - (deprecate) hour12

Here's the most complex situation.
We don't support an extension key `hc` (hourCycle) which can take one of four values (h12, h11, h23, h24)[0] and instead we use a boolean `hour12` option.
I suggest that we add `hourCycle` option and `hc` to extension keys and optionally deprecate `hour12`.

On top of that, we should add `tz` to relevant extension keys, and add `calendar` and `numberingSystem` to `options`.

[0] http://www.unicode.org/reports/tr35/#Locale_Extension_Key_and_Type_Data

zbraniecki:
That's the first step to fix #105

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -115,17 +115,23 @@
           1. If _p_ not *undefined*, then
             1. Set _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the row to _p_.
         1. If _dateTimeFormat_ has an internal slot [[Hour]], then
-          1. Let _hc_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h12"*, *"h24"* &raquo;, *undefined*).
+          1. Let _hc_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;, *undefined*).
           1. Let _hr12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
-          1. If _hr12_ is *true*, then
-            1. Set _hc_ to *"h12"*.
-          1. Else, if _hr12_ is *false*, then
-            1. Set _hc_ to *"h24"*.
+          1. If _hr12_ is not *undefined*, then
+            1. Let _hourNo0_ be Get(_dataLocaleData_, *"hourNo0"*).
+            1. If _hr12_ is *true*, then
+              1. If _hourNo0_ is *true*, then
+                1. Set _hc_ to *"h11"*.
+              1. Else,
+                1. Set _hc_ to *"h12"*.
+            1. Else, if _hr12_ is *false*, then
+              1. If _hourNo0_ is *true*, then
+                1. Set _hc_ to *"h23"*.
+              1. Else,
+                1. Set _hc_ to *"h24"*.
           1. If _hc_ is not *undefined*, then
             1. Set _dateTimeFormat_.[[HourCycle]] to _hc_.
-          1. If _dateTimeformat_.[[HourCycle]] is *"h12"*, then
-            1. Let _hourNo0_ be Get(_dataLocaleData_, *"hourNo0"*).
-            1. Set _dateTimeFormat_.[[HourNo0]] to _hourNo0_.
+          1. If _dateTimeformat_.[[HourCycle]] is *"h11"* or *"h12"*, then
             1. Let _pattern_ be Get(_bestFormat_, *"pattern12"*).
           1. Else,
             1. Let _pattern_ be Get(_bestFormat_, *"pattern"*).
@@ -283,9 +289,11 @@
             1. Let _v_ be the value of _tm_'s field whose name is the Internal Slot column of the matching row.
             1. If _p_ is *"year"* and _v_ ≤ 0, let _v_ be 1 - _v_.
             1. If _p_ is *"month"*, increase _v_ by 1.
-            1. If _p_ is *"hour"* and _dateTimeFormat_.[[HourCycle]] is *"h12"*, then
+            1. If _p_ is *"hour"* and _dateTimeFormat_.[[HourCycle]] is *"h11"* or *"h12"*, then
               1. Let _v_ be _v_ modulo 12.
-              1. If _v_ is 0 and _dateTimeFormat_.[[HourNo0]] is *true*, let _v_ be 12.
+              1. If _v_ is 0 and _dateTimeFormat_.[[HourCucle]] is *"h12"*, let _v_ be 12.
+            1. If _p_ is *"hour"* and _dateTimeFormat_.[[HourCycle]] is *"h24"*, then
+              1. If _v_ is 0, let _v_ be 24.
             1. If _f_ is *"numeric"*, then
               1. Let _fv_ be FormatNumber(_nf_, _v_).
             1. Else if _f_ is *"2-digit"*, then
@@ -415,7 +423,7 @@
 
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
-        1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%DateTimeFormatPrototype%"`, &laquo; [[InitializedIntlObject]], [[InitializedDateTimeFormat]], [[Locale]], [[Calendar]], [[NumberingSystem]], [[TimeZone]], [[Weekday]], [[Era]], [[Year]], [[Month]], [[Day]], [[Hour]], [[Minute]], [[Second]], [[TimeZoneName]], [[HourCycle]], [[HourNo0]], [[Pattern]], [[BoundFormat]] &raquo;).
+        1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, `"%DateTimeFormatPrototype%"`, &laquo; [[InitializedIntlObject]], [[InitializedDateTimeFormat]], [[Locale]], [[Calendar]], [[NumberingSystem]], [[TimeZone]], [[Weekday]], [[Era]], [[Year]], [[Month]], [[Day]], [[Hour]], [[Minute]], [[Second]], [[TimeZoneName]], [[HourCycle]], [[Pattern]], [[BoundFormat]] &raquo;).
         1. Perform ? InitializeDateTimeFormat(_dateTimeFormat_, _locales_, _options_).
       </emu-alg>
       <emu-normative-optional>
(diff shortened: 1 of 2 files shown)


zbraniecki:
@caridy - r?

caridy:
will take a look at is next week. also remember that we want to document this process for future similar enhancements.

zbraniecki:
@caridy  - ping?

spec/datetimeformat.html
```diff
@@ -592,7 +600,7 @@
       </p>
 
       <p>
-        The function returns a new object whose properties and attributes are set as if constructed by an object literal assigning to each of the following properties the value of the corresponding internal slot of this DateTimeFormat object (see <emu-xref href="#sec-properties-of-intl-datetimeformat-instances"></emu-xref>): locale, calendar, numberingSystem, timeZone, hour12, weekday, era, year, month, day, hour, minute, second, and timeZoneName. Properties whose corresponding internal slots have the value *undefined* are not assigned.
+        The function returns a new object whose properties and attributes are set as if constructed by an object literal assigning to each of the following properties the value of the corresponding internal slot of this DateTimeFormat object (see <emu-xref href="#sec-properties-of-intl-datetimeformat-instances"></emu-xref>): locale, calendar, numberingSystem, timeZone, hourCycle, weekday, era, year, month, day, hour, minute, second, and timeZoneName. Properties whose corresponding internal slots have the value *undefined* are not assigned.
```
littledan: Could you add support here for emulating the old hour12 field in resolvedOptions when hourCycle is one of the forms supported by it?

caridy: I'm not sure what are you asking here @littledan, can you clarify?

littledan: If anyone is using the `hour12` field of the resolvedOptions(), this patch makes it go away. That risks web compatibility. For that reason, I'd rather we not merge this patch.

One way to resolve this would be to hold around the `hour12` option that the users passed in, and then just pass that back out. I'd suggest, on the other hand, to just use the \[\[HourCycle]] internal slot that we already have: if the hourCycle is one of the two values that hour12 would provide, then use that in an `hour12` property; otherwise, omit the `hour12` property.

In my opinion, it's OK to have `hour12` in the output even if it wasn't in the input. This is just another instance of the pattern where `resolvedOptions` has options that the user didn't provide, but were chosen by, e.g., locale selection or matching a pattern.

I've made this suggestion a couple times in the past; I must not have been explaining myself clearly. I'll follow up with spec text to explain the change.

caridy: Oh, you're absolutely right! I didn't think of this before.

zbraniecki:
Updated. @caridy and @littledan - can you re-review please?

spec/datetimeformat.html
```diff
@@ -592,7 +600,11 @@
...
+      </p>
+
+      <p>
+        For web compatibility reasons, if the property hourCycle is set to values *h12* or *h24*, the corresponding property hour12 should be set as well.
```
littledan: This works, but I'm not a huge fan of this wording for a couple minor reasons:
- Seems like, suddenly, `hourCycle` comes last in the enumeration order of the resolvedOptions object. This is extremely minor, but it's a rather unmotivated (only by spec wording constraints) observable effect of the change.
- Comments like "for web compatibility reasons" belong in notes, rather than normative text (otherwise it looks a little bit like an "Annex B" normative-optional thing). There are tons and tons of things in all kinds of specs are for web compatibility reasons; probably better to just say what the behavior is.

It would be best to reword the resolvedOptions methods in a more imperative way, as suggested in https://github.com/tc39/ecma402/issues/124 .

spec/datetimeformat.html
```diff
@@ -113,12 +114,16 @@
           1. Let _p_ be Get(_bestFormat_, _prop_).
           1. If _p_ not *undefined*, then
             1. Set _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the row to _p_.
-        1. Let _hr12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
         1. If _dateTimeFormat_ has an internal slot [[Hour]], then
-          1. If _hr12_ is *undefined*, then
-            1. Let _hr12_ be Get(_dataLocaleData_, *"hour12"*).
-          1. Set _dateTimeFormat_.[[Hour12]] to _hr12_.
+          1. Let _hc_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h12"*, *"h24"* &raquo;, *undefined*).
```
littledan: Should h11 and h23 be in this list as well? (I thought that was the whole point of the feature.)

spec/datetimeformat.html
```diff
@@ -278,7 +283,7 @@
             1. Let _v_ be the value of _tm_'s field whose name is the Internal Slot column of the matching row.
             1. If _p_ is *"year"* and _v_ ≤ 0, let _v_ be 1 - _v_.
             1. If _p_ is *"month"*, increase _v_ by 1.
-            1. If _p_ is *"hour"* and _dateTimeFormat_.[[Hour12]] is *true*, then
+            1. If _p_ is *"hour"* and _dateTimeFormat_.[[HourCycle]] is *true*, then
```
littledan: Rather than true, do you want to compare it to h11 and h12?

spec/datetimeformat.html
zbraniecki: do you want to block landing of this until #124 is resolved?

I'd be happy to apply any wording you can suggest here, but would prefer not to wait for #124.

spec/datetimeformat.html
zbraniecki: No, h11 and h23 can be added later to `hourCycle`. For now, we do not support them, and I'd prefer to add new features (like support for other hc options) in a separate proposal leaving this just to migrating the option name from a binary, to multioption.

spec/datetimeformat.html
zbraniecki: since I don't want to add other hour cycle formats yet, I'll just switch it to 'h12'.

spec/datetimeformat.html
littledan: I'd prefer to add hourCycle together with the other possible values. What's the reason to wait? Is the main motivation for this feature to have the `hc` BCP 47 option?

zbraniecki: > What's the reason to wait? 

Supporting more options requires algorithm changes around hour token formatting.

Ok, I'll work on that, but this will increase the size of this patch substantially and require a re-review.

zbraniecki:
@littledan - switched the patch to cover `h11` and `h23` removing `hourNo0`. Can you review it please?

littledan:
Oh, I had forgotten about `hourNo0`--maybe because it's not implemented, tested or documented anywhere but the spec (see #149). Good catch! Because of that, I'd be happy to remove it like this.

Just for completeness, another option for the API surface would be to, instead, reinforce `hourNo0`. It could be used even when `hour12` is false, and then the `hc` key could be calculated as the combination of the `hourNo0` and `hour12` options. (The prohibition on hour12 = false and hourNo0 = true could be removed too, to provide `hc23`.) Maybe this could be considered a "cleaner" option, though I don't have a real preference.

LGTM for this, but if @caridy, @jungshik or @srl295 could weigh in on this API vs that potential one.

spec/datetimeformat.html
```diff
@@ -622,8 +640,7 @@
...
       <li>[[TimeZone]] is a String value with the IANA time zone name of the time zone used for formatting.</li>
       <li>[[Weekday]], [[Era]], [[Year]], [[Month]], [[Day]], [[Hour]], [[Minute]], [[Second]], [[TimeZoneName]] are each either *undefined*, indicating that the component is not used for formatting, or one of the String values given in <emu-xref href="#table-datetimeformat-components"></emu-xref>, indicating how the component should be presented in the formatted output.</li>
-      <li>[[Hour12]] is a Boolean value indicating whether 12-hour format (*true*) or 24-hour format (*false*) should be used. It is only used when [[Hour]] is not *undefined*.</li>
-      <li>[[HourNo0]] is a Boolean value indicating whether hours from 1 to 12 (*true*) or from 0 to 11 (*false*) should be used. It is only used when [[Hour12]] has the value *true*.</li>
+      <li>[[HourCycle]] is a String value indicating whether 12-hour format (*"h11"*, *"h12"*) or 24-hour format (*"h23"*, *"h24"*) should be used. It is only used when [[Hour]] is not *undefined*.</li>
```
littledan: Should this description explain the difference between the 11/23 and 12/24 values?

zbraniecki:
I think it's all ready to land then :)

spec/datetimeformat.html
littledan: Good fix, seems ready to land.

caridy:
@zbraniecki can you help with the tests needed for this?

littledan:
Sorry that I didn't state this earlier, but I think that in addition to the patch looking good, it's nice if we have a couple things in place before actually merging:
- test262 tests (this is a theoretical process requirement that the committee adopted, but it's not being followed consistently in the main spec either)
- implementation experience in one engine (this is not generally a requirement, but I'd prefer it for non-trivial things so we can know that it's a decent plan, and that it's likely to become web reality)

I guess it's already merged, but if we could keep these requirements in mind for future normative changes, it'd be nice.

zbraniecki:
I'm working on the implementation for SpiderMonkey in bug https://bugzilla.mozilla.org/show_bug.cgi?id=1386146 and will add test262 tests as well in the process.

PR #136, opened 2017-03-17
Fixes #131: avoid reading {minimum,maximum}SignificantDigits twice in SetNumberFormatDigitOptions()

### LINKED ISSUE #131: Should we really be reading {minimum,maximum}SignificantDigits twice?
It looks like, in [SetNumberFormatDigitOptions](https://tc39.github.io/ecma402/#sec-setnfdigitoptions), if either one returns something other than **undefined** for the first Get call, we do a *second* Get to get the real value that we use (from inside GetNumberOption). Is there a reason for this, besides that it was easier to write the spec that way? It would seem more straightforward if we just stuck with the value from the first time we read it out of the object.

caridy:


zbraniecki:
I think this is a regression. You make the further down check if `mxsd`, `mnsd` to never be undefined, which in turn means that we will get default values into slots even when user didn't pass anything.

spec/numberformat.html
```diff
@@ -20,14 +20,12 @@
...
         1. Let _mxfdActualDefault_ be max( _mnfd_, _mxfdDefault_ ).
         1. Let _mxfd_ be ? GetNumberOption(_options_, *"maximumFractionDigits"*, _mnfd_, 20, _mxfdActualDefault_).
-        1. Let _mnsd_ be ? Get(_options_, *"minimumSignificantDigits"*).
-        1. Let _mxsd_ be ? Get(_options_, *"maximumSignificantDigits"*).
+        1. Let _mnsd_ be ? GetNumberOption(_options_, *"minimumSignificantDigits"*, 1, 21, 1).
+        1. Let _mxsd_ be ? GetNumberOption(_options_, *"maximumSignificantDigits"*, _mnsd_, 21, 21).
         1. Set _intlObj_.[[MinimumIntegerDigits]] to mnid.
         1. Set _intlObj_.[[MinimumFractionDigits]] to mnfd.
         1. Set _intlObj_.[[MaximumFractionDigits]] to mxfd.
         1. If _mnsd_ is not *undefined* or _mxsd_ is not *undefined*, then
```
caridy: those two slots are still not present if one of the two resolved options is undefined.

caridy:
@zbraniecki we are not setting the slots if the resolved value is undefined, and it will be `undefined`  based on the `GetNumberOption` logic, which invokes the `Get` anyways. Run the steps, and let me know.

spec/numberformat.html
zbraniecki: If `minimumSignificantDigits` is not passed in `options`, then the variable `mnsd` will be resolved with the fallback value `1`.

That, in turn will mean that it is not `undefined`, so we'll step into the `if` block and assign it to the slot.

Did I miss something?

caridy: oh, I thought we were not passing a default value, you're absolutely right.

caridy: At the moment, we only use `GetNumberOption()` from this method, but that could change in the future. I guess we will have to get the value, and pass the value to the normalizer with the fallback, let me play around with this a little bit, and we will see.

littledan:
Alternative patch here: https://github.com/tc39/ecma402/pull/139

caridy:
closing in favor of #139

PR #137, opened 2017-03-17
Include a note about formatToPart and format length being 1

### LINKED ISSUE #133: Intl.DateTimeFormat.prototype.formatToParts's length should be changed to 1 to follow #76
The length of function returned from `Intl.DateTimeFormat.prototype.format` was changed from 0 to 1 in the 2nd edition (#76). So I think we should change `Intl.DateTimeFormat.prototype.formatToParts`'s length to be also 1, so it's consistent with `Intl.DateTimeFormat.prototype.format`.

/cc @zbraniecki

zbraniecki:
Fixes #133


caridy:
https://github.com/tc39/ecma402/issues/133#issuecomment-287489180

PR #138, opened 2017-03-18
Normative: Intl.DateTimeFormat.prototype.formatToParts.length is 1

### LINKED ISSUE #133: Intl.DateTimeFormat.prototype.formatToParts's length should be changed to 1 to follow #76

littledan:
This is specified by marking the argument as not optional, which
makes its implicit length 1, according to notational conventions.

Closes #133

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -566,17 +566,17 @@
     </emu-clause>
 
     <emu-clause id="sec-Intl.DateTimeFormat.prototype.formatToParts">
-      <h1>Intl.DateTimeFormat.prototype.formatToParts ( [ _date_ ] )</h1>
+      <h1>Intl.DateTimeFormat.prototype.formatToParts ( _date_ )</h1>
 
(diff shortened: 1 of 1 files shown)


caridy:
the reason why we make it optional was to align with `.format()`, where the value is also optional.

littledan:
@caridy Wasn't this changed to 1, as noted in 3acd08b87a3004344b61aa3fe380dae17c4e79f4 ?

caridy:
Ah, I forgot about that! you're right.

littledan:
We need a test262 update, cc @leobalter @zbraniecki

PR #139, opened 2017-03-18
Normative: Avoid reading options twice

### LINKED ISSUE #131: Should we really be reading {minimum,maximum}SignificantDigits twice?

littledan:
This patch refactors NumberFormat's option parsing code to avoid
redundant Get calls for {minimum,maximum}SignificantDigits .

Fixes #131

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -320,6 +320,22 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-defaultnumberoption" aoid="DefaultNumberOption">
+      <h1>DefaultNumberOption ( _value_, _minimum_, _maximum_, _fallback_ )</h1>
+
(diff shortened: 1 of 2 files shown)


caridy:
@zbraniecki waiting for you to look at this before I merge it.

zbraniecki:
lgtm!

littledan:
We need test262 tests for this new logic, cc @leobalter

PR #148, opened 2017-06-28
Normative: Make the prototypes ordinary objects

### LINKED ISSUE #122: Can we change the Intl prototypes to be normal objects?
From https://bugzilla.mozilla.org/show_bug.cgi?id=1332604:

Intl.prototype.{Collator, DateTimeFormat, NumberFormat} are currently spec'ed as being instances of Intl.{Collator, DateTimeFormat, NumberFormat} (cf. 10.3, 11.4, 12.4). We should consider changing them to normal objects similar to how Date.prototype was changed to a normal object in the ECMAScript specification for ES6. 

This change should be web-compatible because V8 already treats the prototype objects as normal objects.

littledan:
Closes #122

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -195,7 +195,7 @@
     <h1>Properties of the Intl.Collator Prototype Object</h1>
 
     <p>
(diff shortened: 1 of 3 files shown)


rwaldron:
@ericf ping

caridy:
@rwaldron I don't think @ericf will get a chance to look at this (he is probably busy with the new gig). Any particular concern?

PR #152, opened 2017-07-27
Minor update README for proposal repos

okuryu:
They seem to have been moved to a new repo.


rxaviers:
In your commit message, do you mean "minor" instead of miner? :)

okuryu:
@rxaviers good catch! force pushed. 😄

PR #155, opened 2017-08-02
Correctly retrieve hourCycle/hour12 from options and Unicode extension value

anba:
c8b9f6e5714c14e4bfd253d02db5d8271d175685
- The "hourCycle" and "hour12" options should always be retrieved and validated from the options object, even when the resolved pattern doesn't contain an "hour" property.
- Otherwise invalid but unused "hourCycle" and "hour12" options aren't rejected,
- and implementations would be forced to resolve the pattern before processing "hourCycle" and "hour12".

4d771c8ef6d09320b6615a22e3a612cbc625b38c
- If an option can be set through a normal option and a Unicode extension value, the normal option should always have precedence over the Unicode extension value (cf. ResolveLocale [1]).
- That means "hourCycle" needs to be retrieved from the options object before calling ResolveLocale (right now https://tc39.github.io/ecma402/#sec-initializedatetimeformat step 12 sets dateTimeFormat.[[HourCycle]] from the Unicode extension value, and step 27.f overrides it from the options value or the locale default value).
- And to ensure we can detect when no hc Unicode extension value was present, the first entry of [[LocaleData]][locale].hc is set to *null*, this is similar to [[SortLocaleData]][locale].co and [[SearchLocaleData]][locale].co. So when we later retrieve dateTimeFormat.[[HourCycle]] again and the value is *null*, we set `hc` to `hcDefault`.

2f681f3005ffc8b7a5b1ea18c9b7be81d3e2d8ab
- To ensure [[HourCycle]] is only present in resolvedOptions() when [[Hour]] is used, we need to reset [[HourCycle]] to *undefined* when [[Hour]] is not present.

14939090f4db878d58735c01ec8e169ba214f654
- Only clarifies to which values hour12 should be set when hourCycle is present.

[1] And better ignore the bugs in ResolveLocale, I really ought to unrot the patches I have laying around for far too long to make ResolveLocale easier to understand. :stuck_out_tongue_closed_eyes:

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -616,7 +616,7 @@
       </p>
 
       <p>
-        For web compatibility reasons, if the property hourCycle is set, the corresponding property hour12 should be set as well.
+        For web compatibility reasons, if the property hourCycle is set, the hour12 property should be set to *true* when hourCycle is *"h11"* or *"h12"*, or to *false* when hourCycle is *"h23"* or *"h24"*.
       </p>
 
       <emu-note>
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -81,6 +81,8 @@
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _hc_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;, *undefined*).
+        1. Set _opt_.[[hc]] to _hc_.
         1. Let _localeData_ be %DateTimeFormat%.[[LocaleData]].
         1. Let _r_ be ResolveLocale( %DateTimeFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %DateTimeFormat%.[[RelevantExtensionKeys]], _localeData_).
         1. Set _dateTimeFormat_.[[Locale]] to _r_.[[locale]].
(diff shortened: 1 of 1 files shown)


spec/datetimeformat.html
```diff
@@ -500,7 +504,10 @@
           The array that is the value of the "nu" property of any locale property of [[LocaleData]] must not include the values "native", "traditio", or "finance".
         </li>
         <li>
-          [[LocaleData]][locale] must have hourCycle property with String values for all locale values.
+          [[LocaleData]][locale].hc must be the array [*null*, *"h11"*, *"h12"*, *"h23"*, or *"h24"*] for all locale values.
```
caridy: Is there a reason for this to be an array?

anba: It's required to be an array per https://tc39.github.io/ecma402/#sec-internal-slots:
> The value of each of these properties must be a non-empty array of those values defined in Unicode Technical Standard 35 for the given key that are supported by the implementation for the given locale, with the first element providing the default value.

(I also have a patch *somewhere* to make this use Lists instead of Arrays, at least I think so...)

caridy: ok, a list makes more sense. thanks @anba, I have added an issue to track that here https://github.com/tc39/ecma402/issues/156

zbraniecki:
yeah, that lgtm!

anba:
> This is a semantic change. I like it, but I think it's important to recognize that the spec did not validate the hour12 option if hour was not specified before hourCycle patch.

The "hour12" option was always validated (ed. 1-4), http://ecma-international.org/ecma-402/4.0/index.html#sec-initializedatetimeformat
> ...
> 26. Let hr12 be ? GetOption(options, "hour12", "boolean", undefined, undefined).
> 27. If dateTimeFormat has an internal slot [[Hour]], then
> ...


> Since the behavior change is not observable (because hour12 was cast to bool, even if you tried to pas hour12: "foo" it would end up as true), I think there's no risk of behavior change either.

But potential getters are executed, e.g. `new Intl.DateTimeFormat(undefined, {year:"numeric", get hour12(){ print("get hour12"); }});` prints "get hour12" in SM/JSC/V8 (didn't test Chakra) even though dateTimeFormat.[[Hour]] is undefined.

spec/datetimeformat.html
```diff
@@ -500,7 +503,10 @@
...
```
littledan: Was the word `or` intended here, or is it a typo?

caridy: Yeah, I haven't seen such wording anywhere in 262 or 402. And reading it the second time, it is true that it is confusing. Does it have to be an array with 1) one of those values? or 2) a combination of those values? or 3) all of those values? I believe it is 1). @anba can you clarify? Maybe we should reword it.

anba: Yes, the "or" is a typo.

I copied parts of the text of [[SearchLocaleData]][locale] (https://tc39.github.io/ecma402/#sec-intl-collator-internal-slots) for the [[LocaleData]][locale] description. And then I copied parts of [[LocaleData]][locale] for [[LocaleData]][locale].hc, but missed to remove the "or".

caridy: I have fixed this issue, no worries!

caridy:
merged with the fix mentioned by @littledan

PR #157, opened 2017-08-02
Editorial and normative changes from an unfinished ES2017-Intl review

anba:
I started these changes for the 2017 release, but I didn't finish it in time, so it is possible that some of the algorithm conventions used are no longer correct. 

And I'm not 100% happy with the new UnicodeExtensionValue operation, because the string index operations still feel a bit too low-level and a more abstract definition is probably easier to understand, but I still think it makes sense to have a separate abstract operation to get the Unicode extension value for a specific key.

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -123,45 +123,38 @@
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-unicodeextensionsubtags" aoid="UnicodeExtensionSubtags">
-      <h1>UnicodeExtensionSubtags ( _extension_ )</h1>
+    <emu-clause id="sec-unicodeextensionvalue" aoid="UnicodeExtensionValue">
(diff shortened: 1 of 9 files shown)


spec/negotiation.html
```diff
@@ -129,53 +123,46 @@
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-unicodeextensionsubtags" aoid="UnicodeExtensionSubtags">
-      <h1>UnicodeExtensionSubtags ( _extension_ )</h1>
+    <emu-clause id="sec-unicodeextensionvalue" aoid="UnicodeExtensionValue">
+      <h1>UnicodeExtensionValue (_extension_, _key_)</h1>
```
caridy: space after ( and before )

caridy:
thanks @anba!

PR #158, opened 2017-08-02
Use lists and records for all internal data

### LINKED ISSUE #156: Use Lists instead of Arrays for Internal slots of Service Constructors
Today, it's required to be an array per https://tc39.github.io/ecma402/#sec-internal-slots:

> The value of each of these properties must be a non-empty array of those values defined in Unicode Technical Standard 35 for the given key that are supported by the implementation for the given locale, with the first element providing the default value.

anba:
Applies on top of #155 (there is a way not to include the commits from #155 here, isn't it?).

fcd6157224e759f73602e03b8112330044c22380 replaces all internal arrays with lists (and objects with records) to fix #156. The commit has a mix of &lt;fieldName&gt; and &amp;lt;fieldName&amp;gt; because I tried to stick with the form which was already used in the algorithms.

And while going through the spec, I couldn't help myself to fix some editorial nits: a9492cb800f6cc5181e51743e61bde9485deb6a7.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -4,7 +4,7 @@
   <emu-clause id="sec-the-intl-collator-constructor">
     <h1>The Intl.Collator Constructor</h1>
     <p>
-      The Intl.Collator constructor is the <dfn>%Collator%</dfn> intrinsic object and a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in 9.1.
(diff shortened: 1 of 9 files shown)


caridy:
there are few `[[<_..._>]]` that requires escaping `<` and `>`, but I can take care of that across the board.

caridy:
merged plus the update on the fieldName to be consistent until we can remove all that (if we ever get to that). thanks @anba!

PR #160, opened 2017-08-08
Re-expose Intl.NumberFormat.formatToParts

stasm:
`Intl.NumberFormat.formatToParts` was first propsed in #30. The spec for it was created in #79 and merged in #100 (with follow-ups). Due to browser implementations not being ready at the time, it was moved back to Stage 3 in #101.  The internal refactoring were kept in master and the user-facing method `formatToParts` was removed from the spec in #102.

As of August 1st, 2017, `Intl.NumberFormat.prototype.formatToParts` has shipped in two engines (behind a flag): [SpiderMonkey](https://bugzilla.mozilla.org/show_bug.cgi?id=1289882) and [V8](https://bugs.chromium.org/p/v8/issues/detail?id=5244).  This PR brings `Intl.NumberFormat.formatToParts` back as Stage 4 proposal.

    > const usd = Intl.NumberFormat('en', { style: 'currency', currency: 'USD' });
    > usd.format(123456.789)
    '$123,456.79'
    > usd.formatToParts(123456.789)
    [ { type: 'currency', value: '$' },
      { type: 'integer', value: '123' },
      { type: 'group', value: ',' },
      { type: 'integer', value: '456' },
      { type: 'decimal', value: '.' },
      { type: 'fraction', value: '79' } ]

    > const pc = Intl.NumberFormat('en', { style: 'percent', minimumFractionDigits: 2 })
    > pc.format(-0.123456)
    '-12.35%'
    > pc.formatToParts(-0.123456)
    [ { type: 'minusSign', value: '-' },
      { type: 'integer', value: '12' },
      { type: 'decimal', value: '.' },
      { type: 'fraction', value: '35' },
      { type: 'literal', value: '%' } ]

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -619,7 +619,7 @@
     </emu-clause>
 
     <emu-clause id="sec-intl.numberformat.prototype.formattoparts">
-      <h1>Intl.NumberFormat.prototype.formatToParts ( [ _value_ ] )</h1>
+      <h1>Intl.NumberFormat.prototype.formatToParts ( _value_ )</h1>
 
       <p>
         When the *Intl.NumberFormat.prototype.formatToParts* is called with an optional argument _value_, the following steps are taken:
@@ -629,7 +629,6 @@
         1. Let _nf_ be *this* value.
         1. If Type(_nf_) is not Object, throw a *TypeError* exception.
         1. If _nf_ does not have an [[initializedNumberFormat]] internal slot, throw a *TypeError* exception.
-        1. If _value_ is not provided, let _value_ be *undefined*.
         1. Let _x_ be ? ToNumber(_value_).
         1. Return ? FormatNumberToParts(_nf_, _x_).
       </emu-alg>
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -618,6 +618,23 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-intl.numberformat.prototype.formattoparts">
+      <h1>Intl.NumberFormat.prototype.formatToParts ( [ _value_ ] )</h1>
+
+      <p>
+        When the *Intl.NumberFormat.prototype.formatToParts* is called with an optional argument _value_, the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Let _nf_ be *this* value.
+        1. If Type(_nf_) is not Object, throw a *TypeError* exception.
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot, throw a *TypeError* exception.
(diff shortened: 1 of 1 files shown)


spec/numberformat.html
```diff
@@ -618,6 +618,23 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-intl.numberformat.prototype.formattoparts">
+      <h1>Intl.NumberFormat.prototype.formatToParts ( [ _value_ ] )</h1>
```
caridy: in the `DateTimeFormat.prototype.formatToParts` we decided to make the argument non-optional. https://tc39.github.io/ecma402/#sec-Intl.DateTimeFormat.prototype.formatToParts, this should do the same.

stasm: Thanks. Should there be an explicit check for the existence of the argument, too?

caridy:
@zbraniecki @stasm can you guys coordinate the tests needed for this? Let's aim for the next meeting to get this to stage 4, by that time, we need to get test262 to include this one.

zbraniecki:
The test have been coordinated back in August :) https://github.com/tc39/test262/tree/master/test/intl402/NumberFormat/prototype/formatToParts

littledan:
Yeah, I think this should be good to land as soon as it gets officially to Stage 4. LGTM at that point.

@domenic has suggested in the past allowing things to reach Stage 4 outside of actual meetings, based on meeting all the requirements and discussion online. However, we didn't reach agreement on doing this.

zbraniecki:
This has reached Stage 4 today. I believe we can merge it now :)

spec/numberformat.html
```diff
@@ -618,6 +618,22 @@
...
+      <h1>Intl.NumberFormat.prototype.formatToParts ( _value_ )</h1>
+
+      <p>
+        When the *Intl.NumberFormat.prototype.formatToParts* is called with an optional argument _value_, the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Let _nf_ be *this* value.
+        1. If Type(_nf_) is not Object, throw a *TypeError* exception.
+        1. If _nf_ does not have an [[initializedNumberFormat]] internal slot, throw a *TypeError* exception.
```
caridy: one minor editorial detail, `initializedNumberFormat` => `InitializedNumberFormat` now that internal slots are camel case. I can probably take care of that directly.

caridy: done https://github.com/tc39/ecma402/commit/675495b49c64bc62d1815106abd9df5afa134128

stasm: Thanks, @caridy!

littledan:
Yay!

PR #167, opened 2017-09-07
Editorial: Remove [[InitializedIntlObject]] internal slot

### LINKED ISSUE #115: initializedIntlObject mechanics still necessary?
In ECMA 402 v1, the internal slot `[[initializedIntlObject]]` provided the function of ensuring that a single object didn't have the constructor called on it multiple times for different Intl constructors. However, with ECMA 402 v2 and greater, you can't initialize an existing object as in intl object. Are the mechanics still needed? cc @rwaldron

littledan:
This internal slot has not done anything since ECMA 402 v1.
This patch simply removes it.

Closes #115

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -51,8 +51,6 @@
       </p>
 
       <emu-alg>
-        1. If _collator_.[[InitializedIntlObject]] is *true*, throw a *TypeError* exception.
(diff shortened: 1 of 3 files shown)


littledan:
Thanks for the merge.

PR #168, opened 2017-09-07
Editorial: Add backticks around strings

### LINKED ISSUE #126: javascript strings should be using backticks
In many different places, we are using "somevalue", including `<th>Property</th>` columns in various tables, those should be wrap with backticks for the proper presentation in spec text.

littledan:
Closes #126


caridy:
@littledan there are some conflicts to be resolved here before we can merge it.

PR #170, opened 2017-09-08
Normative: Use a null prototype for the default options

### LINKED ISSUE #123: Change InitializeCollator and InitializeNumberFormat to use default options which don't inherit from %ObjectPrototype%
InitializeCollator and InitializeNumberFormat currently use `ObjectCreate(%ObjectPrototype%)` to create the default options object. For consistency with InitializeDateTimeFormat (through ToDateTimeOptions) and to enable the common browser fast-paths for Number.prototype.toLocaleString and String.prototype.localeCompare, we should consider changing InitializeCollator and InitializeNumberFormat to use  `ObjectCreate(null)`.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1296233 for why the browser fast paths for Number.prototype.toLocaleString and String.prototype.localeCompare aren't spec-compliant.

littledan:
If an options object is not passed in, then use the default options
rather than querying for possible overrides in Object.prototype.
There are a couple points of motivation here:
- It is simpler to not fall back to Object.prototype, and the fallback
  there feels unexpected.
- WebIDL dictionaries default to not reading Object.prototype; it
  is nice to have consistency here.
- Real engine implementations were not consistently implementing the
  fallback for certain paths, such as Number.prototype.toLocaleString
  and String.prototype.localeCompare, where it is possible to observe
  some invalid caching which would become valid with this patch
- DateTimeFormat already does not look to Object.prototype for defaults.

Closes #123

cc @anba

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -55,7 +55,7 @@
         1. Set _collator_.[[InitializedIntlObject]] to *true*.
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. If _options_ is *undefined*, then
-          1. Let _options_ be ObjectCreate(%ObjectPrototype%).
+          1. Let _options_ be ObjectCreate(*null*).
         1. Else,
           1. Let _options_ be ? ToObject(_options_).
         1. Let _u_ be ? GetOption(_options_, *"usage"*, *"string"*, &laquo; *"sort"*, *"search"* &raquo;, *"sort"*).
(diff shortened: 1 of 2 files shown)


littledan:
I don't think we need to spend committee time on it, but I'd appreciate some more feedback from the group here on this patch. Should we land a change like this?

littledan:
I wouldn't say that, wherever there's an inconsistency, we don't have to go to committee. It's just that this is a pretty narrow change, it's gotten two LGTMs, and has tests and an implementation. So I'll merge it.

PR #172, opened 2017-09-08
Editorial: Describe intentions for compatibility across implementations

littledan:
Articulation of goals as ECMA 402 evolves, meant to address
Microsoft's concerns about making a separate, compatible implementation.


littledan:
We agreed on this change at the September 2017 TC39 meeting.

PR #175, opened 2017-09-13
Normative: Add calendar and numberingSystem options

littledan:
This patch allows calendar and numberingSystem to be specified in
the options bag of the DateTimeFormat and NumberFormat constructors.
One use case for these options would be, when working with locales
which have two numbering systems and calendars in use--the UA default
may be the non-Western one, but in some contexts, it may be appropriate
to use the Western one. Currently, without the patch, it would be
necessary to parse the BCP 47 language tag in the application, but Intl
provides no library to do so. For example, this all occurs in this bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1370086

Related bug: #105 . This patch leaves out "collation" because of a lack
of clear use cases.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -84,10 +84,13 @@
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
         1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, *"string"*, *undefined*, *undefined*).
-        1. If _calendar_ does not match `type`, throw a *RangeError* exception.
+        1. If _calendar_ is not *undefined*, then
+          1. If _calendar_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
         1. Set _opt_.[[ca]] to _calendar_.
         1. Let _numberingSystem_ be ? GetOption(_options_, *"numberingSystem"*, *"string"*, *undefined*, *undefined*).
         1. If _numberingSystem_ does not match `type`, throw a *RangeError* exception.
+        1. If _numberingSystem_ is not *undefined*, then
+          1. If _numberingSystem_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
         1. Set _opt_.[[nu]] to _numberingSystem_.
         1. Let _hour12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
         1. Let _hourCycle_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;, *undefined*).
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -73,12 +73,22 @@
         The abstract operation InitializeDateTimeFormat accepts the arguments _dateTimeFormat_ (which must be an object), _locales_, and _options_. It initializes _dateTimeFormat_ as a DateTimeFormat object. This abstract operation functions as follows:
       </p>
 
+      <p>
+        The following algorithm refers to <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+      </p>
+
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Let _options_ be ? ToDateTimeOptions(_options_, *"any"*, *"date"*).
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, *"string"*, *undefined*, *undefined*).
+        1. If _calendar_ does not match `type`, throw a *RangeError* exception.
+        1. Set _opt_.[[ca]] to _calendar_.
+        1. Let _numberingSystem_ be ? GetOption(_options_, *"numberingSystem"*, *"string"*, *undefined*, *undefined*).
+        1. If _numberingSystem_ does not match `type`, throw a *RangeError* exception.
+        1. Set _opt_.[[nu]] to _numberingSystem_.
         1. Let _hour12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
         1. Let _hourCycle_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;, *undefined*).
         1. If _hour12_ is not *undefined*, then
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -74,7 +74,7 @@
       </p>
 
       <p>
-        The following algorithm refers to <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
       </p>
 
       <emu-alg>
@@ -85,12 +85,13 @@
         1. Set _opt_.[[localeMatcher]] to _matcher_.
         1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, *"string"*, *undefined*, *undefined*).
         1. If _calendar_ is not *undefined*, then
-          1. If _calendar_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
+          1. If _calendar_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
+          1. Set _calendar_ to the result of mapping _calendar_ to lower case as as specified in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
         1. Set _opt_.[[ca]] to _calendar_.
         1. Let _numberingSystem_ be ? GetOption(_options_, *"numberingSystem"*, *"string"*, *undefined*, *undefined*).
-        1. If _numberingSystem_ does not match `type`, throw a *RangeError* exception.
         1. If _numberingSystem_ is not *undefined*, then
-          1. If _numberingSystem_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
+          1. If _numberingSystem_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
+          1. Set _numberingSystem_ to the result of mapping _numberingSystem_ to lower case as as specified in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
         1. Set _opt_.[[nu]] to _numberingSystem_.
         1. Let _hour12_ be ? GetOption(_options_, *"hour12"*, *"boolean"*, *undefined*, *undefined*).
         1. Let _hourCycle_ be ? GetOption(_options_, *"hourCycle"*, *"string"*, &laquo; *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;, *undefined*).
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -41,7 +41,7 @@
       </p>
 
       <p>
-        The following algorithm refers to <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
       </p>
 
       <emu-alg>
@@ -55,7 +55,8 @@
         1. Set _opt_.[[localeMatcher]] to _matcher_.
         1. Let _numberingSystem_ be ? GetOption(_options_, `"numberingSystem"`, `"string"`, *undefined*, *undefined*).
         1. If _numberingSystem_ is not *undefined*, then
-          1. If _numberingSystem_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
+          1. If _numberingSystem_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
+          1. Set _numberingSystem_ to the result of mapping _numberingSystem_ to lower case as as specified in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
         1. Set _opt_.[[nu]] to _numberingSystem_.
         1. Let _localeData_ be %NumberFormat%.[[LocaleData]].
         1. Let _r_ be ResolveLocale(%NumberFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %NumberFormat%.[[RelevantExtensionKeys]], _localeData_).
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -40,6 +40,10 @@
         The abstract operation InitializeNumberFormat accepts the arguments _numberFormat_ (which must be an object), _locales_, and _options_. It initializes _numberFormat_ as a NumberFormat object. The following steps are taken:
       </p>
 
+      <p>
+        The following algorithm refers to <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+      </p>
+
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. If _options_ is *undefined*, then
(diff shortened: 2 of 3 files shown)


littledan:
Test262 tests: https://github.com/tc39/test262/pull/1225

V8 implementation: https://chromium-review.googlesource.com/c/v8/v8/+/664762

littledan:
Any other opinions here? Can this be merged?

littledan:
In particular, do you think we should go to TC39 with this change, or is it minor enough to merge it without that presentation?

caridy:
I think we can do this without any presentation, but let's loop in few more folks to get their feedback:
@zbraniecki @jungshik

zbraniecki:
I agree with @caridy

caridy:
@littledan do you want to merge this?

spec/datetimeformat.html
```diff
@@ -81,6 +81,12 @@
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _calendar_ be ? Get(_options_, *"calendar"*).
+        1. If _calendar_ is not *undefined*, set _calendar_ to ? ToString(_calendar_).
+        1. Set _opt_.[[ca]] to _calendar_.
+        1. Let _numberingSystem_ be ? Get(_options_, *"numberingSystem"*).
+        1. If _numberingSystem_ is not *undefined*, set _numberingSystem_ to ? ToString(_numeringSystem_).
```
anba: Typo: numeringSystem

caridy: @littledan can you update the PR to fix these typos?

littledan: Fixed the typo, apologies for the delay.

littledan:
This PR reached consensus at the November 2017 TC39 meeting, see [notes](https://github.com/tc39/tc39-notes/blob/master/es8/2017-11/nov-28.md#10ij-additional-intl-options-for-addition-by-needs-consensus-pr).

anba:
Nit: Instead of

 1. Let _calendar_ be ? Get(_options_, *"calendar"*).
 1. If _calendar_ is not *undefined*, set _calendar_ to ? ToString(_calendar_).
 1. Set _opt_.[[ca]] to _calendar_.

The `GetOption` operation should be used for consistency with access to other options.

 1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, *"string"*, *undefined*, *undefined*).
 1. Set _opt_.[[ca]] to _calendar_.

(And we should change `Get(options, "timeZone")` in InitializeDateTimeFormat to use `GetOption`, too.)

But probably more important than how to access and convert the new options: Currently all options-bag options seem to be validated in some way and invalid option values are rejected with an error, whereas invalid options from Unicode extensions are simply ignored (e.g. `Intl.Collator("de", {caseFirst: "invalid"})` throws a RangeError, whereas `Intl.Collator("de-u-kf-invalid")` is silently accepted). This change introduces the first options-bag options which don't get any validation at all, which seems a bit unfortunate. And furthermore we should also formally define the allowed values for the "numberingSystem" and "calendar" options (BCP 47 vs. UTS 35 names).

For example let's consider the "calendar" option:
https://tc39.github.io/ecma402/#sec-properties-of-intl-datetimeformat-instances describes [[Calendar]] as:
> [[Calendar]] is a String value with the "type" given in Unicode Technical Standard 35 for the calendar used for formatting.

But which "type" is meant here? The value of the "name" attribute of the `<type>` element for BCP 47 (https://unicode.org/repos/cldr/tags/latest/common/bcp47/calendar.xml) or the value of "type" attribute of the `<calendar>` element (https://unicode.org/repos/cldr/tags/latest/common/main/root.xml)? And is it then `Intl.DateTimeFormat("de", {calendar: "gregory"})` or `Intl.DateTimeFormat("de", {calendar: "gregorian"})` or maybe even both? Also what about canonicalizing the input, for example if the input is "GrEgOrIaN", does it get accepted and if it is accepted, does it get canonicalized to "gregorian"? And if we only accept known calendars per UTS 35, we probably also need to explicitly disallow the name "generic", don't we?

For example `Intl.DateTimeFormat("de-u-ca-gregory").resolvedOptions().calendar` prints "gregory" for me with V8 and SpiderMonkey, but "gregorian" with JSC (on Ubuntu16.04, IIRC JSC uses the system ICU which is ICU55 for Ubuntu16.04; V8 and SpiderMonkey both use ICU60). Right now I'd say only JSC is correct here and "gregorian" is the expected output. But WDYT?

littledan:
Unfortunately I don't think ICU has an API for validating these sorts of things. I was thinking of the validation logic a bit like, things which were based on an open-ended name would be checked through `resolvedOptions`.

Do you think it will be important for users to get a RangeError here? Does anyone remember the original intention with the current design for error checking? cc @NorbertLindenberg

We reviewed this patch at the ECMA 402 meeting. At that time, I'd only skimmed @anba's comment and didn't realize the validation issue. I guess this patch isn't ready to merge until we make a decision on that question.

zbraniecki:
As discussed on the ECMA402 call - we should only validate the options to be well-formed, just like we do for extension key values.

littledan:
OK, the newly uploaded patch is based on validating the grammar, but not contents, of the calendar and numberFormat options. It does not address @anba's other canonicalization question, however. How does this seem?

caridy:
hmm, at first glance, `"match type"` feels weird to me considering that you will have to open the link, and try to get the glimpse of what `type` is. This seems to be simple enough that maybe we can specify it here.

zbraniecki:
SpiderMonkey does not have a bug filled yet, and would be a great first-bug if anyone is interested in a small task to help us push this spec forward. I'm happy to mentor.

FrankYFTang:
@FrankYFTang

spec/datetimeformat.html
```diff
@@ -73,12 +73,22 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
...
+      </p>
+
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Let _options_ be ? ToDateTimeOptions(_options_, `"any"`, `"date"`).
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, `"localeMatcher"`, `"string"`, &laquo; `"lookup"`, `"best fit"` &raquo;, `"best fit"`).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _calendar_ be ? GetOption(_options_, `"calendar"`, `"string"`, *undefined*, *undefined*).
+        1. If _calendar_ does not match `type`, throw a *RangeError* exception.
```
FrankYFTang: "does not match `type`" is not implementable. I propose we follow https://tc39.github.io/proposal-intl-locale/#sec-Intl.Locale and change it to
```
        1. If _calendar_ is not *undefined*, then
          1. If _calendar_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
```

spec/datetimeformat.html
```diff
@@ -73,12 +73,22 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
...
+        1. Set _opt_.[[ca]] to _calendar_.
+        1. Let _numberingSystem_ be ? GetOption(_options_, `"numberingSystem"`, `"string"`, *undefined*, *undefined*).
+        1. If _numberingSystem_ does not match `type`, throw a *RangeError* exception.
```
FrankYFTang: Again, "does not match `type`" is  not implementable. I propose we follow https://tc39.github.io/proposal-intl-locale/#sec-Intl.Locale and change it to

```
          1. If _numberingSystem_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.
```

spec/numberformat.html
```diff
@@ -49,6 +53,9 @@ <h1>InitializeNumberFormat ( _numberFormat_, _locales_, _options_ )</h1>
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, `"localeMatcher"`, `"string"`, &laquo; `"lookup"`, `"best fit"` &raquo;, `"best fit"`).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _numberingSystem_ be ? GetOption(_options_, `"numberingSystem"`, `"string"`, *undefined*, *undefined*).
+        1. If _numberingSystem_ does not match `type`, throw a *RangeError* exception.
```
FrankYFTang: Again, "does not match type" is not implementable. I propose we follow https://tc39.github.io/proposal-intl-locale/#sec-Intl.Locale and change it to
```
          1. If _numberingSystem_ does not match the `(3*8alphanum) *("-" (3*8alphanum))` sequence, throw a *RangeError* exception.

```

FrankYFTang:
Should this also apply to https://github.com/tc39/proposal-intl-relative-time ?

FrankYFTang:
How about "collation" for Intl.Collator ?

FrankYFTang:
v8 implementation in https://chromium-review.googlesource.com/c/v8/v8/+/1575919
Have some issue of the condition of throwing RangeError in the spec.

littledan:
I agree with all of your comments. We should make these changes in this PR. Cc @ms2ger.

littledan:
Collation is omitted with a lack of clear use cases, following discussion in https://github.com/tc39/ecma402/issues/105#issuecomment-253768712 . I agree we should apply this to Intl.RelativeTimeFormat to take a numberingSystem option and plan to follow up with that. Otherwise, I've updated this PR based on the comments.

FrankYFTang:
v8 implementation is checked in under flag --harmony-intl-add-calendar-numbering-system It will be nice if someone can contribute tests to test262.

FrankYFTang:
Could we now consider this PR reach "stage 3" status?

FrankYFTang:
If we all agree this PR reach stage 3 status, I will ask to ship it in v8 w/ m77. The two tests in test262 verify it are in 
test262/intl402/NumberFormat/numbering-system-options
test262/intl402/DateTimeFormat/numbering-system-calendar-options

V8 is now implemented behind the flag --harmony-intl-add-calendar-numbering-system

FrankYFTang:
> I will ask to ship it in v8 w/ m77.

sorry. I missed the deadline for m77. I can only try it again for m78.

FrankYFTang:
I missed the deadline for m78 again. I am trying to make it with #349 and #351 together for Chrome m79. Not sure I can make it in. But still have time.

anba:
@FrankYFTang It seems the current V8 implementation changes the resolved locale when the new options are present. This isn't actually correct per spec, so it may make sense to fix this before shipping the feature.
For example I'm currently seeing the following output:
```js
d8> new Intl.DateTimeFormat("en", {calendar: "gregory"}).resolvedOptions().locale
"en-u-ca-gregory"
```

But the expected output is just `"en"`.

(Keywords which are added through options aren't reflected in the overall resolved locale, see <https://tc39.es/ecma402/#sec-resolvelocale>, step 8.i, where `supportedExtensionAddition` is set to the empty string when the options value overrides the Unicode extension value.)

FrankYFTang:
> @FrankYFTang It seems the current V8 implementation changes the resolved locale when the new options are present. This isn't actually correct per spec, so it may make sense to fix this before shipping the feature.
> For example I'm currently seeing the following output:
> 
> ```js
> d8> new Intl.DateTimeFormat("en", {calendar: "gregory"}).resolvedOptions().locale
> "en-u-ca-gregory"
> ```
> 
> But the expected output is just `"en"`.
> 
> (Keywords which are added through options aren't reflected in the overall resolved locale, see https://tc39.es/ecma402/#sec-resolvelocale, step 8.i, where `supportedExtensionAddition` is set to the empty string when the options value overrides the Unicode extension value.)

Thanks for catching this- track in https://bugs.chromium.org/p/v8/issues/detail?id=9887

jswalden:
SpiderMonkey has a bunch of its own tests in https://hg.mozilla.org/mozilla-central/rev/dd340ac5da49b543871350ff1f7e56e51cae4059 that are possibly fodder for testing here. 
 They definitely encounter the usual problem of depending upon a specific set of locales supported, so they couldn't be just adopted unchanged.

littledan:
I've rebased and updated the PR to follow @anba's suggestions above. The lowercase conversion is a really necessary bug fix IMO. Probably we need to add some additional test262 tests to verify.

bkardell:
Ok, I am having some trouble understanding...  Looking at the spec, the test and using the implementations I have a question.  Here's a matrix of various inputs and qualities.. 

|constructor args|parts has `year`|parts has `yearName`|parts has `relatedYear`|
|---|---|---|---|---|
| "en" | yes | no | no |
| "en",  {year: "numeric"} | yes | no | no |
| "en-u-ca-chinese" | no | no | yes |
| "en-u-ca-chinese",  {year: "numeric"} | no | yes | yes |


as a developer I find it hard to understand the logic behind these results..  why _wouldn't_ you have a  `.year` in some cases? It seems very strange to pass year numeric in the last case and have a `relatedYear`, but not a `.year`?  I'm sure there's probably very good reasons - I'm just wondering if someone can help explain it to me so that I can document it?

bkardell:
@FrankYFTang thanks... sorry, I still don't entirely understand. If there is a better venue to ask you these kinds of questions, let me know.. In the meantime, here are some more clarifying quesitons - sorry if they seem basic or something..  Does this mean that:


1.`"en-u-ca-chinese"` will _never_ have an entry called `year` -  no matter what you put in year option?
2. `"en-u-ca-chinese"` will always have an entry called `yearName``, which is the year, but a string-oriented name and therefore doesn't just belong in `year`?
3. `"en-u-ca-chinese"`  will always have a property called `relatedYear` which will contain.... what is normally in the `year` except is never affected by what you put in the opt for year?

FrankYFTang:
> @FrankYFTang thanks... sorry, I still don't entirely understand. If there is a better venue to ask you these kinds of questions, let me know.. In the meantime, here are some more clarifying quesitons - sorry if they seem basic or something.. Does this mean that:
> 
> 1. `"en-u-ca-chinese"` will _never_ have an entry called `year` -  no matter what you put in year option?

As the current spec, YES

> 2. `"en-u-ca-chinese"` will always have an entry called `` yearName``, which is the year, but a string-oriented name and therefore doesn't just belong in  ``year`?

YES, because the form is neither "numeric" nor "2-digit"

> 3. `"en-u-ca-chinese"`  will always have a property called `relatedYear` which will contain.... what is normally in the `year` except is never affected by what you put in the opt for year?

First, it may not ALWAYS output 'relatedYear'. It MAY output it, depending on the values in the options.
Second, a better way to say it is , if the formatter output 'relatedYear', the 'relatedYear' will contain what is the value in the `year` OF THE Gregorian calendar, not the Chinese calendar. That is why it is called "related"- a year value NOT in that request calendar, but in a different calendar system - the Gregorian calendar. The reason in the modern day, even when user want to see the date output in Chinese calendar (or some other non Gregorian calendar), they ALSO want to see the year of Gregorian calendar next to it. That is why it is also output in the POSITION where the year of that calendar is output.

bkardell:
Thanks @FrankYFTang that's very helpful.  I still find it confusing as someone not yet deeply into this problem, but at least I kind of understand how it winds up here now!

littledan:
@bkardell and I had talked about this before filing the issue, and the case that I found surprising was where we actually had a `relatedYear` but no `yearName`, which you can see in Brian's data above, e.g.,

 var df = new Intl.DateTimeFormat("en-u-ca-chinese")
 var parts = df.formatToParts(new Date())
 console.log(JSON.stringify(parts))

 [{"type":"month","value":"12"},{"type":"literal","value":"/"},{"type":"day","value":"21"},{"type":"literal","value":"/"},{"type":"relatedYear","value":"2019"}]

This works the same in V8 and SpiderMonkey.

@FrankYFTang wrote, 

> > "en-u-ca-chinese" will never have an entry called year - no matter what you put in year option?
>
> As the current spec, YES

This is the part where I didn't quite understand the current spec. I guess this is because in [ToLocalTime](https://tc39.es/ecma402/#sec-tolocaltime), the calendar calculation is independent of the format being output? So, even though we're just using a month/day/year format and there's no yearName, we still use relatedYear. Is that right? The motivation here is that 2019 is not the year in the Chinese calendar? If that's the case, then I'm a bit confused about why this display format exists in the locale database...

FrankYFTang:
> @bkardell and I had talked about this before filing the issue, and the case that I found surprising was where we actually had a `relatedYear` but no `yearName`, which you can see in Brian's data above, e.g.,
> 
> var df = new Intl.DateTimeFormat("en-u-ca-chinese")
> var parts = df.formatToParts(new Date())
> console.log(JSON.stringify(parts))
> 
> [{"type":"month","value":"12"},{"type":"literal","value":"/"},{"type":"day","value":"21"},{"type":"literal","value":"/"},{"type":"relatedYear","value":"2019"}]
> 
> This works the same in V8 and SpiderMonkey.
> 
> @FrankYFTang wrote,
> 
> > > "en-u-ca-chinese" will never have an entry called year - no matter what you put in year option?
> > 
> > 
> > As the current spec, YES
> 
> This is the part where I didn't quite understand the current spec. I guess this is because in [ToLocalTime](https://tc39.es/ecma402/#sec-tolocaltime), the calendar calculation is independent of the format being output? So, even though we're just using a month/day/year format and there's no yearName, we still use relatedYear. Is that right? The motivation here is that 2019 is not the year in the Chinese calendar? If that's the case, then I'm a bit confused about why this display format exists in the locale database...

The skeleton it asked inside ICU is "yMd" for this case, and the pattern is 
                yMd{"M/d/r"}
notice, the 'r' which is the related year
see
https://cs.chromium.org/chromium/src/third_party/icu/source/data/locales/en.txt?q=locales/en.txt&sq=package:chromium&dr&l=269

Feel free to file a CLDR ticket ask CLDR linguist to clarify the rationale. 
In Taiwan, for example, that is now how they print on the news paper. For example 
https://chenminjen07.files.wordpress.com/2015/04/b329a-e59c96e789871.jpg
The top line (above the big bold title of the headline, left next to vertical red banner of the 3 characters title of the newspaper)  is date in both 'roc' calendar and 'chinese' calendar

littledan:
Thanks for explaining, @FrankYFTang

leobalter:
My understanding is that we have tests for this. cc @caiolima, is this correct?

anba:
This PR landed after `Intl.Locale`, but didn't fix the issue outlined in <https://github.com/tc39/proposal-intl-locale/issues/96>. Maybe revert the PR for now?

PR #178, opened 2017-09-18
Editorial: Update Array.prototype.toLocaleString to mirror the algorithm in ECMA-262. (Fixes #173)

### LINKED ISSUE #173: Editorial: Update Array.prototype.toLocaleString's algorithm to ECMA-262's style modulo material differences
Updating the algorithm would remove confusion about the differences between the two and fix errors.

Since the difference is minor, add a note specifying that the only difference is the additional `« locales, options »` argument to `Invoke(nextElement, "toLocaleString")`.

See below for more details.

262:

```
1. Let array be ? ToObject(this value).
2. Let len be ? ToLength(? Get(array, "length")).
3. Let separator be the String value for the list-separator String appropriate for the host environment's current locale (this is derived in an implementation-defined way).
4. Let R be the empty String.
5. Let k be 0.
6. Repeat, while k < len
    a. If k > 0, then
        i. Set R to the string-concatenation of R and separator.
    b. Let nextElement be ? Get(array, ! ToString(k)).
    c. If nextElement is not undefined or null, then
        i. Let S be ? ToString(? Invoke(nextElement, "toLocaleString")).
        ii. Set R to the string-concatenation of R and S.
    d. Increase k by 1.
7. Return R.
```

402:

```
1. Let array be ? ToObject(this value).
2. Let len be ? ToLength(? Get(array, "length")).
3. Let separator be the String value for the list-separator String appropriate for the host environment’s current locale (this is derived in an implementation-defined way).
4. If len is zero, return the empty String.
5. Let firstElement be ? Get(array, "0").
6. If firstElement is undefined or null, then
    a. Let R be the empty String.
7. Else,
    a. Let R be ? ToString(? Invoke(firstElement, "toLocaleString", « locales, options »)).
8. Let k be 1.
9. Repeat, while k < len
    a. Let S be a String value produced by concatenating R and separator.
    b. Let nextElement be ? Get(array, ToString(k)).
    c. If nextElement is undefined or null, then
        i. Let R be the empty String.
    d. Else,
        ii. Let R be ? ToString(? Invoke(nextElement, "toLocaleString", « locales, options »)).
    e. Let R be a String value produced by concatenating S and R.
    f. Increase k by 1.
10. Return R.
```

I believe there are some errors in 402's version. Otherwise, these algorithms appear to have been intended as isomorphisms to each other, modulo the additional `« locales, options »` argument to `Invoke` in 402.

402's `? Get(array, ToString(k))` is missing a `!` before `ToString` as is present in 262's version: `? Get(array, ! ToString(k))`

The handling of the special case for index 0 (not adding a separator before the first element) was handled more elegantly in 262's version (hence the shorter length of the algorithm).

Note that 262's version has a reference to the dfn string-concatenation (`Set R to the string-concatenation of R and S.`) whereas 402 says: `Let R be a String value produced by concatenating S and R.`.

In 402's version:

```
8. Let k be 1.
9. Repeat, while k < len
...
    c. If nextElement is undefined or null, then
        i. Let R be the empty String.
```

9.c. would mean when we see undefined or null then we restart with an empty string. This is clearly an error in copy-paste from step 6. This is correct in 262's version.

/cc @jmdyck

dilijev:
Also added note about the effective difference between this algorithm and the version from ECMA-262.

See https://github.com/tc39/ecma262/pull/662 for the update to ECMA-262's version.

Fixes #173

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -226,27 +226,25 @@
       </p>
 
       <emu-alg>
-        1. Let _A_ be ? ToObject(*this* value).
-        1. Let _len_ be ? ToLength(? Get(_A_, *"length"*)).
-        1. Let _separator_ be the String value for the list-separator String appropriate for the host environment’s current locale (this is derived in an implementation-defined way).
-        1. If _len_ is zero, return the empty String.
-        1. Let _firstElement_ be ? Get(_A_, *"0"*).
-        1. If _firstElement_ is *undefined* or *null*, then
-          1. Let _R_ be the empty String.
-        1. Else,
-          1. Let _R_ be ? ToString(? Invoke(_firstElement_, *"toLocaleString"*, &laquo; _locales_, _options_ &raquo;)).
-        1. Let _k_ be 1.
-        1. Repeat, while _k_ < _len_
-          1. Let _S_ be a String value produced by concatenating _R_ and _separator_.
-          1. Let _nextElement_ be ? Get(_A_, ToString(_k_)).
-          1. If _nextElement_ is *undefined* or *null*, then
-            1. Let _R_ be the empty String.
-          1. Else,
-            1. Let _R_ be ? ToString(? Invoke(_nextElement_, *"toLocaleString"*, &laquo; _locales_, _options_ &raquo;)).
-          1. Let _R_ be a String value produced by concatenating _S_ and _R_.
+        1. Let _array_ be ? ToObject(*this* value).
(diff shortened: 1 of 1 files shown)


dilijev:
@littledan That thought occurred to me, but I think it's because 262 doesn't want to make any assumptions about what those parameters might be used for. Also they would have different names so that difference would still need to be reflected in the two algorithms as written in the spec. In practice, implementations will probably simply pass along the parameters in any case (e.g. Chakra has -Intl- which would fall back on the 262 version but it's actually the same).

littledan:
FWIW I believe V8 will pass them along in any case too.

PR #181, opened 2017-10-03
Remove notion of "objects initialized as Collator/NumberFormat/DateTimeFormat"

anba:
66b872f79f2ccff5259896a5728af47fa351cc41:
- Arbitrary objects can no longer be initialized as Intl object instances, so we can remove the phrases "xyz has been successfully initialized as a Collator/NumberFormat/DateTimeFormat". 
- I've also aligned the type tests for Intl objects to only test for the presence of the `[[Initialized<...>]]` internal slots. Newer code only tested for the presence of these slots, whereas older code still tested that the value of the internal slot is `true`.
- The initial value of `[[Bound<...>]]` internal slots is already `undefined`, so we can also remove the additional assignment. 
- And some drive-by fixes.

22d11efe64383659a63e94a50be059fa1c112d11:
- FormatNumberToString will also be called with Intl.PluralRules objects, so we shouldn't assert that the argument is a NumberFormat object.
- We also need to test for the value of `[[MinimumSignificantDigits]]` and `[[MaximumSignificantDigits]]` internal slots, not if they're present.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -97,8 +97,6 @@
         1. Set _collator_.[[Sensitivity]] to _s_.
         1. Let _ip_ be ? GetOption(_options_, *"ignorePunctuation"*, *"boolean"*, *undefined*, *false*).
         1. Set _collator_.[[IgnorePunctuation]] to _ip_.
-        1. Set _collator_.[[BoundCompare]] to *undefined*.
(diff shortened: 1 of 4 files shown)


littledan:
LGTM, thanks for the fixes and editorial improvements @anba !

caridy:
👍

PR #182, opened 2017-10-03
Add algorithms for resolvedOptions()

### LINKED ISSUE #124: rewrite `resolvedOptions()` to a more algorithmic style to eliminate the usage of `[[<>]]` in spec text

anba:
037a18ed4e4d62fc6249691e5aaedea3744b2008
Because the current algorithms don't explicitly specify in which order the properties are added, the new algorithms are a bit ugly, especially for `Intl.Collator` with its optional internal slots. :disappointed: 
We should probably fix this and add the properties in a specific order.

568e7bdbfd951853ff5f527ba5616b8ccb91e9ca 
No longer needed with the previous commit. Other, non-normative uses of these terms are removed in the next commit.

cf2b87144b953e4b3e31d7e238b6f35c1b99bb2e
So the preamble matches the current style. Also rephrases and moves a note for the DateTimeFormat bound function to the correct position.

a528cd9d4bc6d1608d6895ee2137701388d15994
IMO the `hour12` property note shouldn't be exclusively about web-compatibility, so I rephrased it a bit.

Fixes #124

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -603,16 +603,99 @@
       <h1>Intl.DateTimeFormat.prototype.resolvedOptions ()</h1>
 
       <p>
-        This function provides access to the locale and formatting options computed during initialization of the object. This function initially invokes the internal algorithm UnwrapDateTimeFormat to get the %DateTimeFormat% object on which to operate.
+        This function provides access to the locale and formatting options computed during initialization of the object.
       </p>
 
-      <p>
-        The function returns a new object whose properties and attributes are set as if constructed by an object literal assigning to each of the following properties the value of the corresponding internal slot of this DateTimeFormat object (see <emu-xref href="#sec-properties-of-intl-datetimeformat-instances"></emu-xref>): locale, calendar, numberingSystem, timeZone, hourCycle, weekday, era, year, month, day, hour, minute, second, and timeZoneName. Properties whose corresponding internal slots have the value *undefined* are not assigned.
-      </p>
+      <emu-alg>
+        1. Let _dtf_ be *this* value.
+        1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
+        1. Let _dtf_ be ? UnwrapDateTimeFormat(_dtf_).
+        1. Let _options_ be ! ObjectCreate(%ObjectPrototype%).
+        1. For each row of <emu-xref href="#table-datetimeformat-resolvedoptions-properties"></emu-xref>, except the header row, in any order, do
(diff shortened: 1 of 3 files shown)


littledan:
Great to have this PR; I'll take a detailed look soon.

caridy:
> We should probably fix this and add the properties in a specific order.

I think we can do that in another PR. I'm fine with this one as it is @anba

spec/datetimeformat.html
```diff
@@ -603,20 +599,107 @@ <h1>Intl.DateTimeFormat.prototype.formatToParts ( _date_ )</h1>
...
+        This function provides access to the locale and formatting options computed during initialization of the object.
       </p>
 
-      <p>
-        The function returns a new object whose properties and attributes are set as if constructed by an object literal assigning to each of the following properties the value of the corresponding internal slot of this DateTimeFormat object (see <emu-xref href="#sec-properties-of-intl-datetimeformat-instances"></emu-xref>): locale, calendar, numberingSystem, timeZone, hourCycle, weekday, era, year, month, day, hour, minute, second, and timeZoneName. Properties whose corresponding internal slots have the value *undefined* are not assigned.
-      </p>
+      <emu-alg>
+        1. Let _dtf_ be *this* value.
+        1. If Type(_dtf_) is not Object, throw a *TypeError* exception.
+        1. Let _dtf_ be ? UnwrapDateTimeFormat(_dtf_).
```
littledan: Good catch; the previous specification was wrong here. This is an observable change; would you be interested in upstreaming a test262 test for it? Here's a one-liner assertion: `"string" === typeof Intl.DateTimeFormat.prototype.resolvedOptions.call(Intl.DateTimeFormat.call({})).locale`

anba: In how far was the previous specification wrong?

littledan: In the sense that, previously, there was no call to UnwrapDateTimeFormat, and now there is.

anba: The requirement to call UnwrapDateTimeFormat was placed in the description of `resolvedOptions()`:

https://tc39.github.io/ecma402/#sec-intl.datetimeformat.prototype.resolvedoptions (emphasis mine):
> This function provides access to the locale and formatting options computed during initialization of the object. **This function initially invokes the internal algorithm UnwrapDateTimeFormat to get the %DateTimeFormat% object on which to operate.**

littledan: Oh, right, no change then.

littledan:
This patch seems ready to merge to me.

PR #194, opened 2017-10-24
Normative: Apply TimeClip in PartitionDateTimePattern.

### LINKED ISSUE #192: Apply TimeClip in PartitionDateTimePattern?
https://tc39.github.io/ecma402/#sec-partitiondatetimepattern states:
> [...] interprets x as a time value as specified in ES2015, 20.3.1.1 [...]

but PartitionDateTimePattern doesn't actually apply [20.3.1.15 TimeClip](https://tc39.github.io/ecma262/#sec-timeclip) to ensure `x` is a valid time value. So I think we should apply the following change to PartitionDateTimePattern:

```diff
- 1. If x is not a finite Number, throw a RangeError exception.
+ 1. Let x be TimeClip(x).
+ 2. If x is NaN, throw a RangeError exception.
```

anba:
Fixes #192 

This change will limit the maximum allowed date-time value and aligns the limit with `Date`.

```js
var dtf = Intl.DateTimeFormat("en-US");
print(dtf.format(8640000000000000 - 1)); // 9/13/275760
print(dtf.format(8640000000000000)); // 9/13/275760
print(dtf.format(8640000000000000 + 1)); // <-- Throws a RangeError with the proposed patch!
```

```js
print(new Date(8640000000000000 - 1).getTime()); // 8639999999999999
print(new Date(8640000000000000).getTime()); // 8640000000000000
print(new Date(8640000000000000 + 1).getTime()); // NaN
```

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -257,7 +257,8 @@
       </p>
 
       <emu-alg>
-        1. If _x_ is not a finite Number, throw a *RangeError* exception.
+        1. Let _x_ be TimeClip(_x_).
+        1. If _x_ is *NaN*, throw a *RangeError* exception.
         1. Let _locale_ be _dateTimeFormat_.[[Locale]].
         1. Let _nfOptions_ be ObjectCreate(*null*).
         1. Perform ! CreateDataPropertyOrThrow(_nfOptions_, `"useGrouping"`, *false*).


littledan:
For what it's worth, these semantics differ from what I see supported today in JSC, SpiderMonkey and V8, which only seem to throw if you pass in a non-finite value.

Another way to phrase the specification would be to, in spec-land, call the Date constructor before converting back to a Number. The intuition would be that we're casting the input to a Date, just like we do for all sorts of other different arguments. This would add the behavior that suddenly, strings could be parsed from here. Not sure whether this behavior is desired or not.

anba:
> For what it's worth, these semantics differ from what I see supported today in JSC, SpiderMonkey and V8, which only seem to throw if you pass in a non-finite value.

They all [silently clamp](https://searchfox.org/mozilla-central/rev/1ebd2eff44617df3b82eea7d2f3ca1b60cc591a0/intl/icu/source/i18n/calendar.cpp#1132-1134) the input at ICU's [MAX_MILLIS](https://searchfox.org/mozilla-central/rev/1ebd2eff44617df3b82eea7d2f3ca1b60cc591a0/intl/icu/source/i18n/gregoimp.h#128), which is probably worse than starting to throw a RangeError for too large values. :-)

And FWIW Edge already throws an error if the input is too large for its internal limits, e.g.
```js
new Intl.DateTimeFormat().format(10**15)
```
reports:
```sh
Could not convert Date to Windows.Foundation.DateTime: value outside of valid range
SCRIPT5106: Could not convert Date to Windows.Foundation.DateTime: value outside of valid range
```

anba:
> Another way to phrase the specification would be to, in spec-land, call the Date constructor before converting back to a Number. The intuition would be that we're casting the input to a Date, just like we do for all sorts of other different arguments. This would add the behavior that suddenly, strings could be parsed from here. Not sure whether this behavior is desired or not.

Hmm, I don't think I want to pursue that road right now, because it's larger change than just reporting an error if the input value is too large.

littledan:
> Hmm, I don't think I want to pursue that road right now, because it's larger change than just reporting an error if the input value is too large.

I still feel a little uneasy referencing TimeClip from outside of the Date class. It feels like we're reaching into the internal details of Date (in a way that I don't see us doing anywhere else) to get the smallest observable change. I don't feel too strongly about it though.

littledan:
> They all silently clamp the input at ICU's MAX_MILLIS, which is probably worse than starting to throw a RangeError for too large values. :-)

Oh, I didn't realize that (in my too-quick testing)! We definitely need to make a change. But, when we're changing from not-throwing to throwing, it's still pretty significant. In theory, there's web compatibility risk, but I bet we can do this without breaking the web.

caridy:
Even though the only thing that we use in 402 from 20.3.1* is https://tc39.github.io/ecma262/#sec-time-values-and-time-range, IIRC, I don't see a big issue using `TimeClip` directly. I'm sure we cross those boundaries multiple times in 402 where we something use what is supposed to be an implementation detail in 262, specially because that separation doesn't really exist today.

I'm also ok with the creation of the Date if we feel that's better. So, I don't really have a strong opinion here.

anba:
The more I look into it, the more I get the feeling the omission of `TimeClip` makes it difficult to implement [ToLocalTime](https://tc39.github.io/ecma402/#sec-tolocaltime) in a spec-conformant way, because `ToLocalTime` requires to implement calendrical calculations per [20.3.1](https://tc39.github.io/ecma262/#sec-overview-of-date-objects-and-definitions-of-abstract-operations) for the Gregorian calendar. And if the input value isn't time-clipped, an implementation isn't allowed to impose any limits on the input value, so for the example the clamping the value in ICU makes it non-compliant. :cry:

anba:
> I'm also ok with the creation of the Date if we feel that's better. So, I don't really have a strong opinion here.

One reason I'm hesitant to implicitly call the `Date` constructor here, is that this will add another API entry point with implementation-defined semantics, because the `Date.parse` algorithm is implementation-defined. And that creates lots of issues in browsers: https://bugzilla.mozilla.org/show_bug.cgi?id=1274354

littledan:
You're right; any time we can avoid going into date parsing, it's better. OK, I'm happy with this patch. We should add test262 tests before merging, and consider whether it needs to be presented at TC39.

anba:
Created test262 PR at tc39/test262#1340.

caridy:
@littledan are you OK merging this?

littledan:
I think this change is a good idea, I always prefer to have an implementation for normative changes before merging, though that's not a general prerequisite. With the meeting coming up very soon, maybe we should  mention this PR in the status update before merging, though I don't know if it's productive to debate it further.

anba:
> I think this change is a good idea, I always prefer to have an implementation for normative changes before merging, [...]

https://bugzilla.mozilla.org/show_bug.cgi?id=1413505 updates SpiderMonkey to apply TimeClip for Intl.DateTimeFormat.

>  With the meeting coming up very soon, maybe we should mention this PR in the status update before merging, though I don't know if it's productive to debate it further.

Was this PR presented at the meeting, and if yes, were there any concerns with this change?

littledan:
> Was this PR presented at the meeting, and if yes, were there any concerns with this change?

No, sorry, I forgot to raise this particular issue, though Caridy mentioned the good work you were doing on Intl PRs. However, I asked the committee if it's OK for the ECMA 402 group to make decisions on small normative changes without consulting the committee, and there was consensus on that.

anba:
OK, thanks for the info. I just wanted to make sure there weren't any open issues for this change to avoid updating Firefox and having later to revert it because the PR didn't get consensus.

littledan:
Seems like this change has [landed](https://hg.mozilla.org/mozilla-central/rev/071076af815d) in Firefox 59.

Given that we have an implementation and test262 tests, let's discuss this in the next ECMA 402 phone call and decide whether the patch should land.

littledan:
We reviewed this change at the ECMA 402 VC meeting and seem to be in agreement on it. Given the tests and implementation, this patch seems good to merge to me.

jungshik:
https://bugs.chromium.org/p/v8/issues/detail?id=7471 is a v8 issue to implement this change.

PR #198, opened 2017-11-17
fixes #34: Intl.PluralRules specification

### LINKED ISSUE #34: Intl.PluralRules()
UPDATE: This proposal has advanced to stage 1, details here: https://github.com/caridy/intl-plural-rules-spec

cloned from: https://groups.google.com/forum/#!topic/javascript-globalization/3nFDf5al5hU

It would be very helpful for all l10n libraries to get CLDR plural forms into Intl API.

The proposed API could look like this:

```
var cardinal = new Intl.PluralFormat(‘en’, {style: ‘cardinal’});
console.log(cardinal.format(0)); // “other”
console.log(cardinal.format(1)); // “one”
console.log(cardinal.format(2)); // “other”

var ordinal = new Intl.PluralFormat(‘en’, {style: ‘ordinal’});
console.log(ordinal.format(11)); // “one”
console.log(ordinal.format(22)); // “two”
console.log(ordinal.format(33)); // “few”
console.log(ordinal.format(44)); // “other”
```

caridy:
* Editorial work related to https://github.com/tc39/ecma402/issues/34
* The ground work for this PR was done in https://github.com/tc39/proposal-intl-plural-rules/pull/40


spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      <h1>InitializePluralRules ( _pluralRules_, _locales_, _options_ )</h1>
+
+      <p>
+        The abstract operation InitializePluralRules accepts the arguments _pluralRules_ (which must be an object), _locales_, and _options_. It initializes _pluralRules_ as a PluralRules object. The following steps are taken
+      </p>
+
+      <emu-alg>
+        1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
+        1. If _options_ is *undefined*, then
+          1. Let _options_ be ObjectCreate(*null*).
```
caridy: this is the only point of contention as far as I can tell with respect to what was approved for stage 4 few weeks ago.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+        1. Let _pr_ be *this* value.
+        1. If Type(_pr_) is not Object, throw a *TypeError* exception.
+        1. If _pr_ does not have an [[InitializedPluralRules]] internal slot, throw a *TypeError* exception.
+        1. Let _n_ be ? ToNumber(_value_).
+        1. Return ? ResolvePlural(_pr_, _n_).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-intl.pluralrules.prototype.resolvedoptions">
+      <h1>Intl.PluralRules.prototype.resolvedOptions ()</h1>
```
caridy: @anba I already copied your work from PR #182 so this proposal has this specified from day 1.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      <p>
+        This function provides access to the locale and options computed during initialization of the object.
+      </p>
+
+      <emu-alg>
+        1. Let _pr_ be *this* value.
+        1. If Type(_pr_) is not Object, throw a *TypeError* exception.
+        1. If _pr_ does not have an [[InitializedPluralRules]] internal slot, throw a *TypeError* exception.
+        1. Let _options_ be ! ObjectCreate(%ObjectPrototype%).
+        1. For each row of <emu-xref href="#table-pluralrules-resolvedoptions-properties"></emu-xref>, except the header row, do
```
caridy: @anba @littledan I took the liberty to not add the "in any order" from this algo step to avoid the same confusion that we have with the other algos. Are you guys ok with this?

spec/pluralrules.html
littledan: Added a test at https://github.com/tc39/test262/pull/1350 . This should be all good.

spec/pluralrules.html
littledan: Seems good to me, I'm not sure why you'd put "in any order".

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+         1. If _dp_ = -1, then
+           1. Set _iv_ to _n_.
+           1. Let _f_ be 0.
+           1. Let _v_ be 0.
+         1. Else,
+           1. Let _iv_ be the substring of _s_ from position 0, inclusive, to position _dp_, exclusive.
+           1. Let _fv_ be the substring of _s_ from position _dp_, exclusive.
+           1. Let _f_ be ? ToNumber(_fv_).
+           1. Let _v_ be ? ToLength(? Get(_fv_, `"length"`)).
+         1. Let _i_ be ? abs(? ToNumber(_iv_)).
```
anba: `abs` is a [mathematical operation](https://tc39.github.io/ecma262/#sec-mathematical-operations), so it should neither be preceded by `?` nor by `!`

/? ToNumber/! ToNumber/

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+         1. If _f_ ≠ 0, then
+           1. Let _ft_ be the value of _fv_ stripped of trailing `"0"`.
+           1. Let _w_ be ? ToLength(? Get(_ft_, `"length"`)).
```
anba: See above.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+           1. Let _t_ be ? ToNumber(_ft_).
```
anba: See above

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+         1. Else,
+           1. Let _w_ be 0.
+           1. Let _t_ be 0.
+         1. Return a new Record { [[Number]]: _n_, [[IntegerDigits]]: _i_, [[NumberOfFractionDigits]]: _v_, [[NumberOfFractionDigitsWithoutTrailing]]: _w_, [[FractionDigits]]: _f_, [[FractionDigitsWithoutTrailing]]: _t_ }.
+       </emu-alg>
+
+       <emu-table id="table-plural-operands">
+         <emu-caption>Internal Slots of Plural Rules Operands</emu-caption>
```
anba: The caption doesn't really match, the table describes the **fields** of the **record** returned by GetOperands and not any **internal slots**.

Records and fields: https://tc39.github.io/ecma262/#sec-list-and-record-specification-type
Internal slots: https://tc39.github.io/ecma262/#sec-object-internal-methods-and-internal-slots

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+       </emu-table>
+
+    </emu-clause>
+
+    <emu-clause id="sec-pluralruleselection" aoid="PluralRuleSelection">
+
+      <h1>PluralRuleSelection ( _locale_, _type_, _n_, _operands_ )</h1>
+
+      <p>
+        When the PluralRuleSelection abstract operation is called with four arguments, it performs implementation dependent algorithm to map _n_ to the appropriate plural representation of the Operand Record _operands_ by selecting the rules denoted by _type_ for the corresponding _locale_, or the String value `"other"`.
```
anba: - /performs implementation dependent/performs an implementation-dependent/  (missing indefinite article and hyphen)
- /Operand Record/Plural Rules Operands Record/ (maybe add `<emu-xref>` ?)

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+
+    <emu-clause id="sec-resolveplural" aoid="ResolvePlural">
+      <h1>ResolvePlural ( _pluralRules_, _n_ )</h1>
+      <p>
+        When the ResolvePlural abstract operation is called with arguments _pluralRules_ (which must be an object initialized as a PluralRules) and _n_ (which must be a Number value), it returns a String value representing the plural form of _n_ according to the effective locale and the options of _pluralRules_. The following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Assert: Type(_pluralRules_) is Object.
+        1. Assert: If _pluralRules_ does not have an [[InitializedPluralRules]] internal slot, throw a *TypeError* exception.
```
anba: This assertion doesn't make sense, it's not possible to throw an error here. Instead:
`1. Assert: _pluralRules_ has an [[InitializedPluralRules]] internal slot.`

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+        1. Assert: Type(_n_) is Number.
+        1. If isFinite(_n_) is *false*, then
```
anba: /isFinite(_n_) is *false*/n is not a finite Number/

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+          1. Return `"other"`.
+        1. Let _locale_ be _pluralRules_.[[Locale]].
+        1. Let _type_ be _pluralRules_.[[Type]].
+        1. Let _s_ be ? FormatNumberToString(_pluralRules_, _n_).
```
anba: Similar to above:
If these operations cannot return with an abrupt completion, `!` instead of `?` should be used.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+
+    <p>
+      The PluralRules constructor is the <dfn>%PluralRules%</dfn> intrinsic object and a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in <emu-xref href="#sec-internal-slots"></emu-xref>.
+    </p>
+
+    <emu-clause id="sec-intl.pluralrules">
+      <h1>Intl.PluralRules ( [ _locales_ [ , _options_ ] ] )</h1>
+
+      <p>
+        When the `Intl.PluralRules` function is called with optional arguments, the following steps are taken:
```
anba: Missing parameter names after `optional arguments`.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+
+    </emu-clause>
+
+  </emu-clause>
+
+  <emu-clause id="sec-intl-pluralrules-constructor">
+    <h1>The Intl.PluralRules Constructor</h1>
+
+    <p>
+      The PluralRules constructor is the <dfn>%PluralRules%</dfn> intrinsic object and a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in <emu-xref href="#sec-internal-slots"></emu-xref>.
```
anba: - https://tc39.github.io/ecma402/#sec-internal-slots also needs to be updated to list `Intl.PluralRules`.
- Is the DefaultLocale restriction also okay for PluralRules? That means a certain locale can only be the default locale if this locale is also supported by the PluralRules constructor.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+    <emu-clause id="sec-intl.pluralrules">
+      <h1>Intl.PluralRules ( [ _locales_ [ , _options_ ] ] )</h1>
+
+      <p>
+        When the `Intl.PluralRules` function is called with optional arguments, the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. If NewTarget is *undefined*, throw a *TypeError* exception.
+        1. Let _pluralRules_ be ? OrdinaryCreateFromConstructor(_newTarget_, %PluralRulesPrototype%).
```
anba: Missing list of internal slots in call to OrdinaryCreateFromConstructor.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      <p>
+        The value of the [[AvailableLocales]] internal slot is implementation defined within the constraints described in <emu-xref href="#sec-internal-slots"></emu-xref>.
+      </p>
+
+      <p>
+        The value of the [[RelevantExtensionKeys]] internal slot is *[]*.
+      </p>
+
+      <emu-note>
+        Unicode Technical Standard 35 describes no locale extension keys that are relevant to pluralization process.
```
anba: /to/to the/

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+  </emu-clause>
+
+  <emu-clause id="sec-properties-of-intl-pluralrules-prototype-object">
+    <h1>Properties of the Intl.PluralRules Prototype Object</h1>
+
+    <p>
+      The Intl.PluralRules prototype object is itself an ordinary object. <dfn>%PluralRulesPrototype%</dfn> is not an Intl.PluralRules instance and does not have an [[InitializedPluralRules]] internal slot or any of the other internal slots of Intl.PluralRules instance objects.
+    </p>
+    <p>
+      In the following descriptions of functions that are properties or [[Get]] attributes of properties of %PluralRulesPrototype%, the phrase "this PluralRules object" refers to the object that is the this  value for the invocation of the function; a *TypeError* exception is thrown if the this value is not an object or an object that does not have an [[InitializedPluralRules]] internal slot.
```
anba: The phrase "this PluralRules object" is unused, so we can remove this paragraph.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      <p>
+        This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
+      </p>
+    </emu-clause>
+
+    <emu-clause id="sec-intl.pluralrules.prototype.select">
+      <h1>Intl.PluralRules.prototype.select( _value_ )</h1>
+
+      <p>
+        When the `select` is called with an argument _value_, the following steps are taken:
```
anba: Missing word "method".

spec/pluralrules.html
anba: Yes, I'm definitely okay with that. 😄

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      Intl.PluralRules instances have an [[InitializedPluralRules]] internal slots.
+    </p>
+
+    <p>
+      Intl.PluralRules instances also have several internal slots that are computed by the constructor:
+    </p>
+
+    <ul>
+      <li>[[Locale]] is a String value with the language tag of the locale whose localization is used by the plural rules.</li>
+      <li>[[Type]] is one of the String values `"cardinal"`, or `"ordinal"`, identifying the plural rules used.</li>
```
anba: /, or / or /   (no comma if only two alternatives are present)

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+      <li>[[MinimumIntegerDigits]] is a non-negative integer Number value indicating the minimum integer digits to be used.</li>
+      <li>[[MinimumFractionDigits]] and [[MaximumFractionDigits]] are non-negative integer Number values indicating the minimum and maximum fraction digits to be used. Numbers will be rounded or padded with trailing zeroes if necessary.</li>
+      <li>[[MinimumSignificantDigits]] and [[MaximumSignificantDigits]] are positive integer Number values indicating the minimum and maximum fraction digits to be used. Either none or both of these properties are present; if they are, they override minimum and maximum integer and fraction digits.</li>
+      <li>[[PluralCategories]] is an array of unique string values, from the the list `"zero"`, `"one"`, `"two"`, `"few"`, `"many"` and `"other"`, that are relevant for the locale whose localization is specified in LDML Language Plural Rules.</li>
```
anba: - I'd remove "unique", because it makes it sound like that string values themselves can be unique or non-unique. Instead of that I'd simply say that the array must not contain any duplicates, because I think that's what was meant by "unique" here.
- "LDML Language Plural Rules" is never defined.
- The complete sentence reads a bit strange: "[The] [...] values [...] are relevant for the locale whose localization is specified in LDML Language Plural Rules." This should be reworded to make it more clear where and why these values are used.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+        1. If _options_ is *undefined*, then
+          1. Let _options_ be ObjectCreate(*null*).
+        1. Else
+          1. Let _options_ be ? ToObject(_options_).
+        1. Let _t_ be ? GetOption(_options_, `"type"`, `"string"`, &laquo; `"cardinal"`, `"ordinal"` &raquo;, `"cardinal"`).
+        1. Set _pluralRules_.[[Type]] to _t_.
+        1. Let _opt_ be a new Record.
+        1. Let _matcher_ be ? GetOption(_options_, `"localeMatcher"`, `"string"`, &laquo; `"lookup"`, `"best fit"` &raquo;, `"best fit"`).
+        1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Perform ? SetNumberFormatOptions(_pluralRules_, _options_, *0*, *3*).
```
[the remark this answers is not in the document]
caridy: we have that everywhere. let's do that in another editorial PR all together.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+        1. Let _r_ be ResolveLocale(%PluralRules%.[[AvailableLocales]], _requestedLocales_, _opt_).
```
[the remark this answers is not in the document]
caridy: > ResolveLocale needs five arguments, but only three are present.

I will chat offline with @littledan and @zbraniecki tomorrow about this.

> Also: Maybe move to an earlier step for consistency with the other Intl constructors?

I could not find a good consistency on this one.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+        1. Set _pluralRules_.[[Locale]] to the value of _r_.[[Locale]].
+        1. Let _pluralCategories_ be a List of Strings representing the possible results of plural selection.
```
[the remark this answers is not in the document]
caridy: yes, that's precisely the conversation that I want to have with the other folks tomorrow. I agree this is underspecified ATM.

spec/pluralrules.html
```diff
@@ -0,0 +1,360 @@
...
+
+    <emu-clause id="sec-getoperands" aoid="GetOperands">
+       <h1>GetOperands ( _s_ )</h1>
+       <p>
+         When the GetOperands abstract operation is called with argument _s_. It performs the following steps:
+       </p>
+
+       <emu-alg>
+         1. Assert: Type(_s_) is String.
+         1. Let _n_ be ? ToNumber(_s_).
```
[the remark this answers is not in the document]
caridy: good catch!

spec/pluralrules.html
caridy: yes, DefaultLocale should also apply to PluralRules IMO. I will ask tomorrow for clarification from the champions.

spec/pluralrules.html
caridy: I will ask for clarifications on this points tomorrow.

spec/pluralrules.html
caridy: will ask for some clarification and maybe some description for the table.

spec/pluralrules.html
zbraniecki: If we don't want them in any order, should we sort the properties table alphabetically? It seems like we'll specify an an arbitrary order with it if I read it right.

spec/pluralrules.html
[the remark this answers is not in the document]
anba: > I could not find a good consistency on this one.

I think the normal order is as follows:
1. Call CanonicalizeLocaleList(locales).
2. Retrieve options to select the correct locale data (currently only "usage" for Intl.Collator).
3. Call GetOption(options, "localeMatcher", ...).
4. Retrieve options which are relevant for ResolveLocale, i.e. options which can also be selected through Unicode extension keywords.
5. Call ResolveLocale.
6. Retrieve any remaining options.

caridy:
> Update https://tc39.github.io/ecma402/#sec-api-overview ?

Done

> Add PluralRules to the list in https://tc39.github.io/ecma402/#sec-language-tags ?

Done

> Add PluralRules to https://tc39.github.io/ecma402/#sec-constructor-properties-of-the-intl-object

Done

spec/pluralrules.html
caridy: using "Plural Rules Operands Record Fields". @littledan do you have any opinion about this?

caridy:
> https://tc39.github.io/ecma402/#annex-implementation-dependent-behaviour needs to be updated, too.

added. "List of Strings representing the possible results of plural selection and their corresponding order per locale." seems to be the only implementation dependent-behavior for PluralRules

> Also allow higher values in https://tc39.github.io/ecma402/#conformance ?

I have added: "The options property `type` in the PluralRule constructor." assuming that there might be some other types that might be implementation-defined. But I'm not sure if that's what you're referencing to with "higher values". Can you explain? 

> Add PluralRules to https://tc39.github.io/ecma402/#locale-and-parameter-negotiation ?

Added there for the list that supports `[[LocaleData]]`, and fixed `InitializePluralRules` algo to call `ResolveLocale` with 5 arguments.

caridy:
@littledan @anba @zbraniecki I'm done with the editorial changes! feel free to do  another pass before we finally merge this.

anba:
> Also allow higher values in https://tc39.github.io/ecma402/#conformance ?
> 
> I have added: "The options property type in the PluralRule constructor." assuming that there might be some other types that might be implementation-defined. But I'm not sure if that's what you're referencing to with "higher values". Can you explain?

`Intl.NumberFormat` allows higher values for five options (integer, fraction, and significant digits), do we want to allow larger values for `Intl.PluralRules`, too?

From https://tc39.github.io/ecma402/#conformance:
> The options properties ***minimumIntegerDigits, minimumFractionDigits, maximumFractionDigits, minimumSignificantDigits, and maximumSignificantDigits*** in the NumberFormat constructor, provided that the additional values are interpreted as integer values higher than the specified limits.

spec/pluralrules.html
littledan: AFAICT ICU, for one, supports PluralRules for all locales where CLDR has data, and doesn't actually let you configure which locales you want to ship PluralRules data for (maybe because it's so little data). So I don't think the DefaultLocale restriction should be a problem for ICU-based implementations.

caridy:
Alright, thanks everyone for the help!

PR #204, opened 2017-12-08
Normative: Don't add hour-cycle to resolved locale if hour12 option is present

### LINKED ISSUE #195: Hour-cycle in resolved locale doesn't match actual resolved hour-cycle if hour12 option is present
Test case:
```js
js> print(JSON.stringify(Intl.DateTimeFormat("en-u-hc-h12", {hour:"2-digit", hour12:false}).resolvedOptions(), null, 2))
{
  "locale": "en-u-hc-h12",
  "calendar": "gregory",
  "numberingSystem": "latn",
  "timeZone": "America/Los_Angeles",
  "hour": "2-digit",
  "hourCycle": "h23",
  "hour12": false
}
```

The hour-cycle in the resolved locale "en-u-hc-h12" doesn't match the actual resolved hour-cycle "h23". 
To fix this we probably need to call `GetOption(options, "hour12", ...)` before calling `ResolveLocale(...)` in InitializeDateTimeFormat.

Previous discussion on #jsapi with @zbraniecki: https://mozilla.logbot.info/jsapi/20171102#c13802734-c13802761

anba:
`null` is a valid option for hour-cycle per https://tc39.github.io/ecma402/#sec-intl.datetimeformat-internal-slots and because explicit options override Unicode extension values in [ResolveLocale](https://tc39.github.io/ecma402/#sec-resolvelocale), we just need to set `_opt_.[[hc]]` to `null` before calling ResolveLocale to make sure any hour-cycle Unicode extension value is properly ignored and doesn't show up in the resolved locale. (And we also need to update the assertion in ResolveLocale to accept `null` values.)

Fixes #195


littledan:
What's the motivation for this semantic change? I don't see the harm in including hourCycle in resolvedOptions even if it's not in the input--actually, I thought this was kind of the point: The new interface is hourCycle, and the options system translates from the legacy interface to that when needed.

anba:
> I don't see the harm in including hourCycle in resolvedOptions even if it's not in the input [...]

With the current semantics the `hc` Unicode extension in the resolved locale and the resolved hour-cycle option can be inconsistent:
```js
js> var opt = new Intl.DateTimeFormat("en-u-hc-h12", {hour:"2-digit", hour12:false}).resolvedOptions();
js> opt.locale
"en-u-hc-h12"    // The hc Unicode extension in the resolved locale is set to h12,
js> opt.hourCycle
"h23"            // but the actual hour-cycle is h23.
```

With this patch, the resolved locale will be set to "en" which matches the default behaviour if an option overrides a Unicode extension value:
```js
js> var opt = new Intl.DateTimeFormat("en-u-hc-h12", {hour:"2-digit", hourCycle:"h24"}).resolvedOptions();    
js> opt.locale   // hourCycle overrides hc-h12, so the resolved locale doesn't include hc.
"en"
js> opt.hourCycle
"h24"
```

littledan:
What if we fixed this in the other direction, to make hour12 override hc? (I would've assumed it did, and just missed this in the review.)

anba:
> What if we fixed this in the other direction, to make hour12 override hc?

I don't quite understand what you mean by "in the other direction". Let's recap the current situation to make sure there aren't any misunderstandings to which property overrides the other property.

1. `hour12` and `hourCycle` are present, but the `hc` Unicode extension isn't. This leads to the following result:
 - The resolved locale doesn't contain `hc`.
 - The variable `hr12` in step 28.d ("If hr12 is not undefined, then") is not undefined, therefore the `hour12` option overrides `hourCycle`.
 - The [[HourCycle]] date-time formatter internal slot is only determined by `hour12`.

2. `hourCycle` is present and the `hc` Unicode extension, but `hour12` is present:
 - The resolved locale doesn't contain `hc`, because 9.2.6 ResolveLocale, step 8.i ("If options has a field [[\<key\>]], then") overrides it.
 - The [[HourCycle]] date-time formatter internal slot is only determined by `hourCycle`.

3. `hour12` is present and the `hc` Unicode extension, but `hourCycle` is present:
 - The resolved locale __contains__ `hc`.
 - The variable `hr12` in step 28.d ("If hr12 is not undefined, then") is not undefined, therefore the `hour12` option overrides `hourCycle`.
 - The [[HourCycle]] date-time formatter internal slot is only determined by `hour12`.


With the proposed patch, only the result of case 3 will be changed to ensure the resolved locale doesn't contain `hc`.

This is achieved by setting `opt.[[hc]]` to `null`, because that leads to executing 9.2.6 ResolveLocale, step 8.i which ensures the resolved locale won't contain any `hc` Unicode extension value.

caridy:
oh wow! something very weird is going on then.

> make hour12 override hc? (I would've assumed it did, and just missed this in the review.)

this was always my assumption as well. if you're using `hour12` option, that rules them all, and must be present in the resolved options. If it is not present, it doesn't need to be in the resolved options.

But it was never the intent to preserve the resolved options shape IIRC. In fact, I will say that, over time, we will continue adding more options, and giving them the impression that resolved options is somehow versioned by the provided options is going to be a big problem.

littledan:
Sorry for my confusion; in reviewing this patch I had forgotten about the lengths that ResolveLocale goes through to preserve the shape of the original locale, as in these examples:

```js
> new Intl.DateTimeFormat('en-US-u-ca-gregory').resolvedOptions().locale
en-US-u-ca-gregory
> new Intl.DateTimeFormat('en-US', {calendar: 'gregory'}).resolvedOptions().locale
en-US
```

This behavior differs from Intl.Locale, which puts everything it has in the locale string exposed by its toString() behavior.

Right now, whenever you get a resolved locale string back, any of the -u- options will have been passed in in the original string. In the semantics of the case 3 that @anba listed above, I see three possibilities:
- Current broken semantics: -u-hc-h12
- This patch: No tag at all
- Alternate semantics which I was sort of expecting: -u-uc-h23

I don't really know how we should organize the spec for the last version, though. In particular, step 28 provides some behavior when both hour12 and hc are present, which you'd then want reflected back in the locale. Seems like you'd have to break up ResolveLocale into some smaller pieces.

I'm not sure whether @anba's patch preserves the behavior of step 28 of InitializeDateTimeFormat, which implements some particular behavior based on when both hourCycle and hour12 are included as options in the bag. I'm not sure how important the behavior is, though (cc @zbraniecki ). In this patch, it looks like if they are both present, hourCycle is just ignored.

littledan:
I don't really know the motivation for not just adding all options to the resolved locale, whether they were present in the input or not, by the way, Does anyone know the history here?

ray007:
`hour12` overriding `hourCycle` feels wrong to me - it's only a subset of what `hourCycle` provides.
Just throw an error when they contradict each other?

anba:
> I'm not sure whether @anba's patch preserves the behavior of step 28 of InitializeDateTimeFormat, which implements some particular behavior based on when both hourCycle and hour12 are included as options in the bag.

The patch shouldn't change any in behaviour of step 28, it only ensures the resolved locale will have a consistent `hc` compared to the actually used hour-cycle. 

I'm kind of wondering if we should move `ResolveLocale` in Intl constructors to later steps (that means after getting all options from the `options` object), because the order in which properties are retrieved is observable whereas `ResolveLocale` itself is side-effect free.

littledan:
> The patch shouldn't change any in behaviour of step 28, it only ensures the resolved locale will have a consistent hc compared to the actually used hour-cycle.

Does your patch change the behavior in the case that the options bag contains both `hourCycle` and `hour12` or does it leave that the same?

> I'm kind of wondering if we should move ResolveLocale in Intl constructors to later steps

Sounds good to me. I am wondering if we could actually switch to WebIDL entirely, and let these options bags be WebIDL dictionaries, so they'd be read entirely up-front. cc @domenic @annevk @tobie

anba:
> Does your patch change the behavior in the case that the options bag contains both hourCycle and hour12 or does it leave that the same?

It leaves it the same.

anba:
> Sounds good to me. I am wondering if we could actually switch to WebIDL entirely, and let these options bags be WebIDL dictionaries, so they'd be read entirely up-front.

We should probably file a different issue for that topic to avoid hi-jacking this thread. :-)
And apart from that I'm not sure WebIDL can handle all option cases and pick the correct defaults (SetNumberFormatDigitOptions!). :-)

littledan:
Reading this thread again, I'm happy with @anba's explanation in https://github.com/tc39/ecma402/pull/204#issuecomment-350550029 , and this patch seems good to me. It'd still be good to have a test and implementation.

anba:
> It'd still be good to have a test and implementation.

tc39/test262#1391 and https://bugzilla.mozilla.org/show_bug.cgi?id=1430809

littledan:
We reviewed this patch at the ECMA 402 VC meeting and had consensus on it. Great work on the test and implementation, @anba .

IMO this patch is good to merge now. We can review this result at the January 2018 TC39 meeting and make any adjustments as a follow-on if needed afterwards.

PR #208, opened 2017-12-21
Editorial fixes and clean-ups

anba:
5cb0883def35f5b5a307531fff92fb25d75791ac
- Some variables were already updated in the past to use the more descriptive long names. Update the remaining option variables to use long names, too.

e91e340db485e9e844f4f670adae63558ccdf18a
- Replaces the iteration over Table 2 with explicit steps. This removes more uses of the dynamic record field access syntax and should also make the spec more similar to actual implementations.

8a962e51d0afecd1ad55af3285fcf9fac6653e71
- The dynamic internal slot access syntax isn't used anymore! 

a3e71a4dca57acbc46736ae5916918adc54b412d
- Makes "sec-collator-compare-functions" a proper sub-clause of "sec-intl.collator.prototype.compare".

 688fd6ca8282b2de535961553ace78b7f814dd5c
- Adds a separate emu-clause for `CompareStrings`.

a42253620f8045cc92e6d81d233b7f2d4cb02ac5
- Switch to internal slots to store the Intl instances instead of bound functions. This matches how the main spec stores per-function this-values. Implementations are still free to use actual bound functions for the bound `compare` and `format' functions.

d3a6dc92c9ab0247e5d865d42ba9244fdab48b45
- Use `%Intl%` to be more explicit that these steps refer to the `Intl` object of the current realm.
- Replace a type check which is always true with an assertion.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -194,67 +194,67 @@
       <emu-note>
         The returned function is bound to _collator_ so that it can be passed directly to `Array.prototype.sort` or other functions.
       </emu-note>
-    </emu-clause>
 
-    <emu-clause id="sec-collator-compare-functions">
-      <h1>Collator Compare Functions</h1>
+      <emu-clause id="sec-collator-compare-functions">
+        <h1>Collator Compare Functions</h1>
 
-      <p>A Collator compare function is an anonymous built-in function.</p>
-      <p>When a Collator compare function is called with arguments _x_ and _y_, the following steps are taken:</p>
+        <p>A Collator compare function is an anonymous built-in function.</p>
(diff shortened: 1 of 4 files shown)


caridy:
picked:

> 5cb0883
> e91e340
> 8a962e5
> 688fd6c
> a422536
> d3a6dc9

As for:

>  8a962e5 (The dynamic internal slot access syntax isn't used anymore)

we can probably work toward that as a goal.

As for:

> a3e71a4dca57acbc46736ae5916918adc54b412d (Make 'Collator Compare Functions' a sub-clause of 'get Intl.Collator.prototype.compare')

I have merged, but it is still different from NumberFormat and DateTimeFormat where the Functions are part of the abstract operations clause. We can align that in another PR at some point.

PR #227, opened 2018-04-17
Normative: Improve handling of non-Gregorian calendars

### LINKED ISSUE #225: Clarifying the expected behavior for non-gregorian calendars in DateTimeFormat.formatToParts
I ran into an implementation issue a few days ago in Chakra where calling formatToParts on a date from a non-Gregorian calendar can cause some confusion. Right now, V8 and Chakra agree that `new Intl.DateTimeFormat("en-u-ca-chinese", { year: "numeric" }).format()` produces the string "2018(wu-xu)", while SpiderMonkey returns the string "35". As a result, when formatting to parts, SpiderMonkey returns a single part, `{ type: "year", value: "35" }`... however, what the expected result should be for V8 and Chakra is unclear to me. "2018(wu-xu)" is created at least on our end by passing "y" to udatpg_getBestPattern, which will return "r(U)" for the Chinese calendar, where "r" is the related Gregorian year and "U" is the year name. 

Before continuing, if "2018(wu-xu)" is not the correct string to be producing here, then this may not be an issue. With that being said, it looks like all of the engines that I have access to on Windows (cant test JSC+Intl right now) generally are using udatpg_getBestPattern or something vaguely similar, so even if this string is wrong, this issue may come up again in the future. Additionally, if the ability to ask for short or long date is added, then we will definitely run into issues where basically any LDML specifier _could_ show up in the users preferred short/long date format, so formatToParts would need to be able to map every LDML specifier to a part type. As such, with the current part types, is it expected that there will be two "year" parts coming from this string? That goes against an unwritten rule that I have always assumed about this API where there would only ever be 0 or 1 of each non-literal part, such that you could do `...formatToParts().filter((part) => part.type === "year")[0]` to get the correct year. Perhaps that was a bad assumption on my part, but if there are multiple year fields, how will an application know which is the preferable one for when they just want to display "the year," whatever that may mean? I don't know enough about the Chinese calendar to say if 2018 or wu-xu (or neither) is the best answer, but I don't think the API is currently set up to handle this situation.

If we agree that 2018 and wu-xu should not both be type = year, I would propose adding a new part type, "relatedYear," to line up with the LDML's "r" field.

/cc @zbraniecki, @jefgen

littledan:
In CLDR, the set of patterns is based on the the calendar, not
simply the language-region-script. Previously, however, the set
was based purely on the base name. This patch switches to a two-level
lookup to permit implementations to have more accuracy.

The patch adds an additional field to DateTimeFormat: relatedYear.
This is the related Gregorian year, e.g., as used in some Chinese calendar
formats when the Gregorial year is put next to the year in a 60-year cycle.

The patch makes an editorial change to make the descriptions
of the locale database fully switch to records (there was some object-style
notation remaining).

Closes #225


littledan:
cc @jackhorton @jungshik

anba:
I guess I agree with @jackhorton's comment from above when looking at the description of 'r' at https://unicode.org/repos/cldr/trunk/specs/ldml/tr35-dates.html#dfst-year. That means if it is supported it should not be an input option and the value should not be `[short, long]` (maybe just `numeric`?). Will there be a similar issue with 'U' (cyclic year name)? Or do we just map 'U' to `year`?

littledan:
Is there anything else needed to verify this patch before landing? (I don't see how we can write test262 tests for it, unfortunately.)

littledan:
In the May 2018 [Intl call](https://github.com/tc39/ecma402/blob/master/meetings/notes-2018-05-17.md), we concluded:
- Only permit `relatedYear: "numeric"` and not `"2-digit"`, to match CLDR
- Leave cyclic year for a possible follow-on, based on demand

The patch was discussed in the presentation at the May 2018 TC39 meeting, with no concerns raised, so I believe it has consensus with the committee.

@anba

> The CLDR link shouldn't link to "https://unicode.org/repos/cldr/trunk/specs/ldml/tr35-dates.html#dfst-year", but instead use a stable version.

What's your reasoning here? I thought we were switching to unversioned Unicode references. cc @mathiasbynens 

> PartitionDateTimePattern, ToLocalTime

Yes, looks like we need an update, seems pretty straightforward.

@jackhorton I'm not sure what you mean.

jackhorton:
Sorry, I was trying to answer "Will there be a similar issue for the cyclic year name "U"?" by saying that my understanding was the consensus was that `U` would become type = year and `r` would become type = relatedYear

anba:
> The CLDR link shouldn't link to "https://unicode.org/repos/cldr/trunk/specs/ldml/tr35-dates.html#dfst-year", but instead use a stable version.
> 
> What's your reasoning here? I thought we were switching to unversioned Unicode references. cc @mathiasbynens

As discussed yesterday, I think we rather want to use the link <https://unicode.org/reports/tr35/tr35-dates.html#dfst-year>, which seems to be the stable link to the latest released version of UTR 35.

littledan:
Last time we discussed this issue, we reached consensus on making the change, and were just waiting for an implementation to merge the patch. Has anyone implemented it and written tests, or think they will do so soon? It's now more practical to write tests than previously due to the test262 `locale` tag. If so, I'll merge this patch. If not, I'll separate out the "calendar" elements from the "relativeYear" elements (where calendar represents implementation reality, and relativeYear is a new feature). cc @jungshik @FrankYFTang @jackhorton

FrankYFTang:
@FrankYFTang

FrankYFTang:
I try to implement this PR in v8 and I believe it does not produce meaningful result.
If I only allow the api to take additional relatedYear option as "numeric" and  convert it to 'r' to specify the desired format, it won't produce meaningful result because in step 2 it will call
Let options be ? ToDateTimeOptions(options, "any", "date").
and therefore we will get options as 
{"relatedYear": "numeric", "year": "numeric", "month": "numeric", "day": "numeric" }
as the return of ToDateTimeOptions. The first one is passed in by the user and the year/month/day are added by ToDateTimeOptions.

Now, after step 22 we will also get
{"year": "numeric", "month": "numeric", "day": "numeric", "relatedYear": "numeric"}
as opt.
In the implementation, what should we turn these into?
According to your issue, it will turn relatedYear into "r".
So we will get the value into a skeleton "yMdr" to try to match pattern.
But there are no pattern in CLDR supporting "yMdr". 

It seems 'r' is not for skeleton but pattern
For example I see the following skeleton->pattern mapping in ICU related to 'r'
https://cs.chromium.org/chromium/src/third_party/icu/source/data/locales/en.txt?q=locales/en.txt&sq=package:chromium&dr&l=245

```
        chinese{
            DateTimePatterns{
...
            }
            availableFormats{
 ...
                Gy{"r(U)"}
                GyMMM{"MMM r(U)"}
                GyMMMEd{"E, MMM d, r(U)"}
                GyMMMd{"MMM d, r"}
...
                UM{"M/U"}
                UMMM{"MMM U"}
                UMMMd{"MMM d, U"}
                UMd{"M/d/U"}
 ...
                y{"r(U)"}
                yMd{"M/d/r"}
                yyyy{"r(U)"}
                yyyyM{"M/r"}
                yyyyMEd{"E, M/d/r"}
                yyyyMMM{"MMM r(U)"}
                yyyyMMMEd{"E, MMM d, r(U)"}
                yyyyMMMM{"MMMM r(U)"}
                yyyyMMMd{"MMM d, r"}
                yyyyMd{"M/d/r"}
                yyyyQQQ{"QQQ r(U)"}
                yyyyQQQQ{"QQQQ r(U)"}
            }
```
but notice 'r' is not on the left side (the skeleton) but only appeared in the right side (the pattern)

littledan:
Interesting, thanks for the analysis. Should we remove relatedYear as an option and only permit it in the formatToParts output then? Cc @ms2ger.

littledan:
Pushed a new PR that omits relatedYear as an option and in resolvedOptions. PTAL.

spec/datetimeformat.html
```diff
@@ -303,6 +304,10 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
             1. Else,
               1. Let _fv_ be an implementation and locale dependent String value representing `"ante meridiem"`.
             1. Add new part record { [[Type]]: `"dayPeriod"`, [[Value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal to `"relativeYear"`, then
```
FrankYFTang: Should this be relativeYear or relatedYear ?

FrankYFTang:
I think this current PR need some improvement. r and U are often appear together in the pattern and I think when we deal with r we also need to deal with U.

Here is the example
```
dtf = new Intl.DateTimeFormat("en-u-ca-chinese", {year: "numeric"})
```
in the above code, after considering the calendar for getting the pattern, we got 
```
                y{"r(U)"}
```
so the pattern is "r(U)"
and the following code 
```
d = new Date();
dtf.format(d);
"2019(ji-hai)"
```
notice "2019" is the r:relatedYear, and "ji-hai" is the U "cyclic year name, as in Chinese lunar calendar" 
see http://userguide.icu-project.org/formatparse/datetime

now, for formatToParts, what should the output of dtf.formatToParts(d); ?
Should it be
```
[{type: "relatedYear", value: "2019"}, 
 {type: "literal", value: "("}, 
 {type: "year", value: "ji-hai"}, 
 {type: "literal", value: ")"}]
```
and treat the U the same way as y ? to report the "ji-hai" part as "year"? or should we report it differently?

FrankYFTang:
v8 prototype in https://chromium-review.googlesource.com/c/v8/v8/+/1588401

spec/datetimeformat.html
littledan: Oops! Fixed.

littledan:
@FrankYFTang Do you think we should give it a different part name? What should that part name be?

FrankYFTang:
> @FrankYFTang Do you think we should give it a different part name? What should that part name be?

I think so. How about "yearName"?

FrankYFTang:
Here is my prototype in v8
https://chromium-review.googlesource.com/c/v8/v8/+/1588401
And here is the output
```
ftang@ftang4:~/v8/v8$ out/x64.release/d8 --harmony_intl_other_calendars
V8 version 7.6.0 (candidate)
d8> dtf = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric"});
{}
d8> t = new Date();
Tue May 21 2019 18:45:39 GMT-0700 (Pacific Daylight Time)
d8> dtf.format(t)
"2019己亥年"
d8> dtf.formatToParts(t)
[{type: "relatedYear", value: "2019"}, {type: "yearName", value: "己亥"}, {type: "literal", value: "年"}]
d8>  dtf = new Intl.DateTimeFormat("zh-u-ca-chinese", {year: "numeric", month: "short"})
{}
d8> dtf.format(t)
"2019己亥年四月"
d8> dtf.formatToParts(t)
[{type: "relatedYear", value: "2019"}, {type: "yearName", value: "己亥"}, {type: "literal", value: "年"}, {type: "month", value: "四月"}]
```

This is more than what you spec in the current change, it also add the "yearName" as the output of ToLocalTime.

littledan:
If you want to upload a revised PR, I am happy to review it. Otherwise, I plan to fix this to match those semantics soon.

FrankYFTang:
I think the change in this PR, in particular 
```
-        1. Let _formats_ be _dataLocaleData_.[[formats]].
+        1. Let _formats_ be _dataLocaleData_.[[formats]].[[&lt;_calendar_&gt;]].
```
Need to be part of https://github.com/tc39/ecma402/pull/175
Otherwise, the adding of "calendar" option mean not much, right?
I suggest we fold this PR into https://github.com/tc39/ecma402/pull/175

littledan:
Subsumed by #349, #350, #351

PR #230, opened 2018-04-25
Check in past notes

littledan:
The Intl call attendees agreed to publish the notes as is, unedited. We've been collecting them in Google docs during meetings. This PR checks them in as Markdown files.


littledan:
Does anyone mind if I just land this PR?

PR #234, opened 2018-04-28
Editorial: Make PluralRules point to the right algorithm

littledan:
s/SetNumberFormatOptions/SetNumberFormatDigitOptions/


littledan:
This is a simple fix, and I have heard no concerns, so merging the patch.

PR #235, opened 2018-05-01
Editorial: Use OrdinaryCreateFromConstructor properly

littledan:
Editorial fix for PluralRules; should consider backporting to ES2018


littledan:
This is a simple editorial fix, and I have heard no concerns, so merging the patch.

PR #245, opened 2018-07-15
Normative: Permit "unknown" DateTimeFormat/NumberFormat fields

### LINKED ISSUE #231: Handling unexpected ICU field IDs
I recently learned that ICU users are supposed to be able to handle unexpected values from iterating over fields. Many ECMA-402 implementations are on top of ICU. I can see a few ways to handle this case:
1. Crash with an assertion failure (V8's strategy, apparently incorrect)
1. Leave out that range in formatToParts (but, this risks not actually presenting the whole string)
1. Choose another string for the type

Choice 3 seems the best to me, especially since these assertion failures [are starting to come up in practice](https://bugs.chromium.org/p/chromium/issues/detail?id=826549). Is `"literal"` is the type we want there? Or should we somehow permit a new type, such as `"unknown"`, for these cases?

littledan:
In practice, libraries which underly ECMA-402 implementations may
implement much of algorithms such as PartitionNumberFormat and
PartitionDateTimeFormat. Sometimes, patterns from these algorithms
will include unexpected fields. To handle these cases, this patch
adds logic to represent them literally in the output of `format`
methods. `formatToParts` methods use the "unknown" field type.

This patch includes additional editorial/typo fixups:
- Remove the previous logic to leave unrecognized fields in patterns
  in place (logic which was broken in DateTimeFormat)
- Define ILD and ILND inline, to reduce confusion
- Include some missing words in comparison

Closes #231


littledan:
cc @anba @jackhorton @jungshik

littledan:
@jackhorton reported in [July's ECMA-402 meeting](https://github.com/tc39/ecma402/blob/master/meetings/notes-2018-07-19.md) that this change is implemented in ChakraCore. The change has consensus within the group, and was reported in the July 2018 TC39 meeting with no concerns raised. It's not really possible to write test262 tests. Given all that, it's ready to land.

PR #263, opened 2018-08-29
Mark some variables as such in SetNumberFormatDigitOptions().

Ms2ger:


littledan:
LGTM. Could you make the commit message of editorial patches like these begin with Editorial: ? Then, I'll land this.

PR #267, opened 2018-09-07
Editorial: Add a note to Intl.NumberFormat.prototype.format about the strange binding.

### LINKED ISSUE #265: Add a note to Intl.NumberFormat.prototype.format
Intl.DateTimeFormat.prototype.format has a note:

> The returned function is bound to _dtf_ so that it can be passed directly to **`Array.prototype.map`** or other functions.

Intl.NumberFormat.prototype.format should probably have something similar.

Ms2ger:
Fixes #265.


littledan:
Rather than cargo-culting the existing note, I'd rather the explanation here focus on how this is a historical mistake (and that the other note be updated as well).

PR #278, opened 2018-10-14
Normative: Don't harden supportedLocalesOf Array

### LINKED ISSUE #258: SupportedLocales: returned Array is weird
Steps 5 through 8 of SupportedLocales (https://tc39.github.io/ecma402/#sec-supportedlocales) create an Array from the List of locales that the function wants to return, then iterate over that Array to make all its own properties (i.e. elements and length) non-writable and non-configurable.

I guess this predates the introduction of `Object.freeze`, because notably it leaves the object extensible: while user code can't delete or modify any of the existing array elements, it can add additional properties to the array.

That said, why is it useful to attempt to write-protect the array in the first place? While there's probably no good reason for user code to want to modify the array returned by an `Intl.Something.supportedLocalesOf()` call, I also don't see a good reason to preemptively make this impossible. If a program wanted to change the result array, it could always create a copy first, and then modify that copy at will. So the operation could be simplified by dropping steps 6 and 7.

In summary, I present the following options for your consideration:

(1) Change nothing, keep the spec as is.
(2) Replace the custom loop around `DefinePropertyOrThrow` with a call to `Object.freeze`.
(3) Simply return a regular mutable array (as returned by CreateArrayFromList).

My personal preference would be (3), because simplification is nice.

Implementation complexity is not a concern; we already have the code for (1), changing to (2) would be easy, changing to (3) even easier.

littledan:
This array has nonwritable, nonconfigurable elements to future-proof
for a path that it doesn't seem like we're planning on going down.
Make it an ordinary Array instead, as suggested by @jakobkummerow.

Closes #258

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -269,12 +269,7 @@
           1. Let _supportedLocales_ be BestFitSupportedLocales(_availableLocales_, _requestedLocales_).
         1. Else,
           1. Let _supportedLocales_ be LookupSupportedLocales(_availableLocales_, _requestedLocales_).
-        1. Let _subset_ be CreateArrayFromList(_supportedLocales_).
-        1. Let _keys_ be _subset_.[[OwnPropertyKeys]]().
-        1. For each element _P_ of _keys_ in List order, do
-          1. Let _desc_ be PropertyDescriptor { [[Configurable]]: *false*, [[Writable]]: *false* }.
-          1. Perform ! DefinePropertyOrThrow(_subset_, _P_, _desc_).
-        1. Return _subset_.
+        1. Return CreateArrayFromList(_supportedLocales_).
       </emu-alg>
     </emu-clause>
 


littledan:
Did we get an implementation and tests?

FrankYFTang:
Well... it is chicken-n-egg, I am waiting for the change of spec to implement the change in the v8 code.

littledan:
Will you write test262 tests if I merge this?

FrankYFTang:
> Will you write test262 tests if I merge this?

sure. Maybe some of them are just modify existing test which may fail under the change.

littledan:
OK, that's good enough for me, given the great implementation and testing work you have been doing, and how trivial this change is. Merged.

FrankYFTang:
> > Will you write test262 tests if I merge this?
> 
> sure. Maybe some of them are just modify existing test which may fail under the change.

Here you go
https://github.com/tc39/test262/pull/2031

I do have a problem to make my v8 pass that test. I think the issue is inside 
https://github.com/tc39/test262/blob/master/harness/propertyHelper.js

function isConfigurable(obj, name) {
  try {
    delete obj[name];
  } catch (e) {
    if (!(e instanceof TypeError)) {
      $ERROR("Expected TypeError, got " + e);
    }
  }
  return !Object.prototype.hasOwnProperty.call(obj, name);
}

function isWritable(obj, name, verifyProp, value) {
  var newValue = value || "unlikelyValue";
  var hadValue = Object.prototype.hasOwnProperty.call(obj, name);
  var oldValue = obj[name];
  var writeSucceeded;

  try {
    obj[name] = newValue;
  } catch (e) {
    if (!(e instanceof TypeError)) {
      $ERROR("Expected TypeError, got " + e);
    }
  }


Somehow the exception I got for the array.length is not TypeError but RangeError. 
https://www.ecma-international.org/ecma-262/9.0/index.html#sec-property-attributes
didn't specify it will get TypeError but only stated
"will fail." and "will not succeed." . I think that file has some problem but not 100% sure how to fix it.

PR #279, opened 2018-10-14
Normative: Specify order of table traversal

### LINKED ISSUE #264: Iteration order not specific enough
In [InitializeDateTimeFormat](https://tc39.github.io/ecma402/#sec-initializedatetimeformat):

> 22. For each row of Table 5, except the header row, do

In [Intl.PluralRules.prototype.resolvedOptions](https://tc39.github.io/ecma402/#sec-intl.pluralrules.prototype.resolvedoptions):

> 8. For each row of Table 8, except the header row, do

In all other cases, the spec defines the iteration order.

cc @FrankYFTang

littledan:
Each place that tables are traversed, it is now specified to be in
the order that the table was written, as opposed to being ambiguous
as previously:
- In resolvedOptions, this is observable in the order of keys
- In the DateTimeFormat constructor, this is observable in the order
  of property Get calls

Closes #264

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -274,7 +274,7 @@
         1. If Type(_collator_) is not Object, throw a *TypeError* exception.
         1. If _collator_ does not have an [[InitializedCollator]] internal slot, throw a *TypeError* exception.
         1. Let _options_ be ! ObjectCreate(%ObjectPrototype%).
-        1. For each row of <emu-xref href="#table-collator-resolvedoptions-properties"></emu-xref>, except the header row, in any order, do
+        1. For each row of <emu-xref href="#table-collator-resolvedoptions-properties"></emu-xref>, except the header row, in table order, do
           1. Let _p_ be the Property value of the current row.
           1. Let _v_ be the value of _collator_'s internal slot whose name is the Internal Slot value of the current row.
           1. If the current row has an Extension Key value, then
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -101,7 +101,7 @@
           1. Let _timeZone_ be DefaultTimeZone().
         1. Set _dateTimeFormat_.[[TimeZone]] to _timeZone_.
         1. Let _opt_ be a new Record.
-        1. For each row of <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, do
+        1. For each row of <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, in table order, do
           1. Let _prop_ be the name given in the Property column of the row.
           1. Let _value_ be ? GetOption(_options_, _prop_, `"string"`, &laquo; the strings given in the Values column of the row &raquo;, *undefined*).
           1. Set _opt_.[[&lt;_prop_&gt;]] to _value_.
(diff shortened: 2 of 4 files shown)


littledan:
@anba Can you say anything about why resolvedOptions was previously specified as, "in any order"?

littledan:
I think there are some differences in engines in the enumeration order of resolvedOptions, last time I checked. This patch shouldn't go in without a test. (@Ms2ger ?) Do you think it's better to leave it ambiguous?

Ms2ger:
> In resolvedOptions, this is observable in the order of keys

Is that actually specified at this point? I thought iteration order of plain objects was left undefined.

(Good change in any case, though.)

littledan:
@Ms2ger iteration order is undefined, but Object reflection methods reveal the real underlying order

Ms2ger:
https://github.com/tc39/test262/pull/1866

littledan:
OK, we can come to a conclusion on this at the next ECMA-402 meeting. It's on the agenda.

anba:
> @anba Can you say anything about why resolvedOptions was previously specified as, "in any order"?

Only because the original specification also didn't specify any order. From <https://github.com/tc39/ecma402/pull/182#issue-144485341>:
> Because the current algorithms don't explicitly specify in which order the properties are added, the new algorithms are a bit ugly, especially for Intl.Collator with its optional internal slots. :disappointed:
We should probably fix this and add the properties in a specific order.

Related SpiderMonkey bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1314799

It shouldn't be too hard to implement this in SpiderMonkey:
- Collator is easy, because it doesn't involve optional properties.
- For NumberFormat and PluralRules there'll be a (slight) performance penalty if the optional properties are no longer placed at the end. But I don't think this matters much in practice.
- DateTimeFormat will require some temporary storage when retrieving the resolved options from the date-time pattern, so another performance penalty. It'll be larger than the one for NumberFormat and PluralRules, but probably also not really relevant in practice, because I don't think `resolvedOptions` is called excessively in user code (*).

(*) V8 has an incorrect implementation which makes me wonder how often it's actually called: Literal parts in the resolved pattern are ignored. For example `new Intl.DateTimeFormat("de", {hour:"numeric"}).resolvedOptions().hour12` is `true` in V8, because the resolved pattern is "HH 'Uhr'", which means the search for "h" is successful in https://github.com/v8/v8/blob/db6db6ed8ff617e9d2258fc970224d9a0eff9ea7/src/objects/js-date-time-format.cc#L182.

littledan:
Should we reorder the tables to put the optional ones at the end?

anba:
Hmm, I'm not sure the possible performance regressions (*) are important enough to justify reordering optional properties at the end. I'd rather make sure the ordering makes "sense" (i.e. it's not totally random) and is consistent across the `Intl` objects. 

(*) I've observed a three times longer runtime in some artificial µ-benchmarks after applying the change for SpiderMonkey: Calling `NumberFormat.prototype.resolvedOptions` a million times one a single NumberFormat object will regress from 30ms to 90ms. But (A) this is unlikely to matter in practice and (B) other engines spend seconds on that µ-benchmark, so :man_shrugging:.

littledan:
OK, let's leave things as is then. I agree that that performance regression doesn't sound bad in practice (why would you call this more than once in a single formatter?!).

Ms2ger:
Tests were merged prematurely, btw.

littledan:
Well, we agreed on this in the ECMA-402 meeting, so let's just go ahead with this change then. If it is somehow unimplementable, we should revisit this, but it sounded like anba has a prototype already.

PR #282, opened 2018-10-15
Normative: Handle negative arguments in FormatNumberToString().

### LINKED ISSUE #261: FormatNumberToString doesn't handle negative/non-finite numbers
`FormatNumberToString` is documented as taking any Number value, but it immediately calls     `ToRawPrecision` or `ToRawFixed`, each of which expects a finite non-negative number. 

It turns out there is indeed a path that calls it with (finite) negative numbers: Intl.PluralRules.prototype.select → ResolvePlural → FormatNumberToString.

This is already fixed in the [Intl.NumberFormat Unified API Proposal](https://github.com/sffc/proposal-unified-intl-numberformat/commit/cc2ed3df413b722af1332470a9b81b52718f774a).

Relevant changes: https://github.com/tc39/ecma402/pull/104, https://github.com/tc39/proposal-intl-plural-rules/pull/15.

Ms2ger:
Fixes #261.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -127,10 +127,16 @@
       </p>
 
       <emu-alg>
+        1. Assert: _x_ is not *NaN*, *+&infin;*, or *-&infin;*.
+        1. Let _isNegative_ be _x_ &lt; *0* or _x_ is *-0*.
+        1. If _isNegative_, then
+          1. Let _x_ be -_x_.
         1. If _intlObject_.[[MinimumSignificantDigits]] and _intlObject_.[[MaximumSignificantDigits]] are both not *undefined*, then
           1. Let _result_ be ToRawPrecision(_x_, _intlObject_.[[MinimumSignificantDigits]], _intlObject_.[[MaximumSignificantDigits]]).
         1. Else,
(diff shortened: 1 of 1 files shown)


Ms2ger:
This matches https://github.com/sffc/proposal-unified-intl-numberformat, but maybe we should also check for *x* = **-0**.

littledan:
I agree, we should check for -0.

Ms2ger:
Fixed.

littledan:
Tests? Does anyone implement this?

littledan:
I don't understand this patch. Where is `[[RoundedNumber]]` used?

sffc:
If I understand correctly, this PR is not valid for the current version of the spec.

In the current spec, `ToRawPrecision` and `ToRawFixed` return a String.  My proposal, unified-intl-numberformat, changes those internal methods to return a Record with the String in key "FormattedString" and the rounded number result in key "RoundedNumber".  This PR is the logic to handle RoundedNumber, but that variable does not exist in the current spec, and therefore this PR is invalid.

littledan:
I see. Sounds like we should punt this to the Unified NumberFormat Options change, as we are doing for other parts of it.

PR #290, opened 2018-10-30
Editorial: Note about BiDi concerns

littledan:
This note came out of the Internationalization WG review of ECMA-402
in the 2018 TPAC meeting.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -1,6 +1,10 @@
 <emu-clause id="datetimeformat-objects">
   <h1>DateTimeFormat Objects</h1>
 
+  <emu-note>
+    The output of the the formatting operations from DateTimeFormat may require the proper bidirectional context to be linguistically correct; see <a href="http://unicode.org/reports/tr35/#Text_Directionality">UTS #35"</a> for details.
+  </emu-note>
+
   <emu-clause id="sec-datetimeformat-abstracts">
     <h1>Abstract Operations For DateTimeFormat Objects</h1>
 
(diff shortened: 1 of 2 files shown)


littledan:
cc @aphillips

littledan:
Rather than saying this for specific APIs, we concluded that we should instead make a single high-level note to explain best practices for interpolation and BiDi, as this isn't quite related to specific APIs https://www.w3.org/International/articles/inline-bidi-markup/

littledan:
We discussed this issue in the November 2018 ECMA-402 meeting, and the conclusion was, we should include a note specification-wide. The issue here is not specific to a particular formatter, but rather universal to working with strings in a BiDi context.

Would anyone be interested in writing such a note?

aphillips:
@littledan I'd be interested, in part because I'm writing one in my day job about the same thing.

A tricky thing here is that if all the browsers supported bidi isolation ([glances at mostly at Edge and a bit at Webkit](https://w3c.github.io/i18n-tests/results/the-dir-attribute-isolation)) and the isolating controls ([glances at _everyone_ except Firefox](https://w3c.github.io/i18n-tests/results/bidi-algorithm#rli_etc)), a lot of the complexity would go out of providing support for bidi string assembly. Since we all expect that support to be forthcoming, the note will necessarily have a "legacy browser" and "non-legacy browser" section to it.

littledan:
OK, then I'll close this PR, and I look forward to @aphillips 's new text for this clarification.

zbraniecki:
@aphillips - did you ever get to write a spec-wide version of this note?

PR #291, opened 2018-11-02
Normative: Permit the use of tailorings in CLDR

### LINKED ISSUE #287: toLocale{Upper,Lower}Case does not take into Greek tailoring in CLDR
Unicode 11 3.13 delegates locale-specific tailoring of case mappings to the CLDR (except for Turkic languages).   However, Ecma 402 spec does not take into account that only mentioning UnicodeData.txt and SpecialCasings.txt. 

As a result, Greek uppercasing (context-dependent removal of diacritic marks) as desired by Greek speakers is not supported.  

ICU-based implementations such as v8 does support it.   Spun off from https://crbug.com/v8/8353  

@ftang @littledan @gsathya

littledan:
Apparently there was never a deliberate decision to omit these
tailorings, but the specification was written back when all tailorings
were in SpecialCasings.txt. The new spec text permits looking at CLDR
as well, matching the Unicode Standard.

Closes #287

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -65,27 +65,21 @@
         1. Else,
           1. Let _requestedLocale_ be DefaultLocale().
         1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with all Unicode locale extension sequences (<emu-xref href="#sec-unicode-locale-extension-sequences"></emu-xref>) removed.
-        1. Let _availableLocales_ be a List with the language tags of the languages for which the Unicode character database contains language sensitive case mappings.
+        1. Let _availableLocales_ be a List with the language tags of the languages for which the Unicode Character Database or CLDR contains language sensitive case mappings.
         1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
         1. If _locale_ is *undefined*, let _locale_ be `"und"`.
         1. Let _cpList_ be a List containing in order the code points of _S_ as defined in ES2019, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, starting at the first element of _S_.
-        1. For each code point _c_ in _cpList_, if the Unicode Character Database provides a lower case equivalent of _c_ that is either language insensitive or for the language _locale_, replace _c_ in _cpList_ with that/those equivalent code point(s).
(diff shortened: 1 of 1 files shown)


littledan:
cc @jungshik @anba

anba:
Step 7 also needs to be updated. Apart from that I wonder if it is possible to refer to <https://unicode.org/reports/tr35/tr35-general.html#Transforms> instead of the Unicode character database **and** CLDR? 

So we'd have for step 7:
```
7. Let _availableLocales_ be a List with the language tags of the languages for which CLDR contains language sensitive case mappings.
```

And steps 11-13:
```
11. Let _transform_ be the string-concatenation of _locale_ and "-Lower".
12. Let _cuList_ be a List where the elements are the result of applying the transformation _transform_ per <here some link and details to https://unicode.org/reports/tr35/tr35-general.html#Transforms>.
```

https://unicode.org/reports/tr35/tr35-general.html#Transform_Rules then also takes care of referring to Unicode 3.13. But I'm not quite sure how to interpret `and "Any" can be used instead of "und"` in UTS 35 and if it means it's possible to use "und-Lower" as an alias for "Any-Lower". So maybe step 9 also needs to be changed to:
```
9. If _locale_ is *undefined*, let _locale_ be "Any".
```

littledan:
Thanks for the suggestions. Why would we specifically have to refer to the Lower transform, as opposed to vaguely pointing at CLDR for answers (as the Unicode Standard itself does)?

anba:
Oh, wait a sec. Does this PR aim to support case mapping either per Unicode or per CLDR? We should pick one to guarantee the same result for `toLocale[LowerUpper]Case` on all implementations.

littledan:
Per the Unicode standard, which references CLDR, so both. But, we cannot pick a single data version, as in practice, vendors may have their own local tailorings.

littledan:
@srl295 suggested that we permit any tailoring, and CLDR is just one option (and that this is what the Unicode Standard says; I just misinterpeted it). I'll follow up with a less restrictive patch.

littledan:
Sorry for leaving this patch without an update; if someone else wants to adopt it, that'd be welcome. cc @sffc.

spectranaut:
I adopted this PR in #341, thanks for starting this work, @littledan !

Issue #223, opened 2018-03-16
Why does ResolveLocale drop Unicode extension keys without values?


anba:
```js
var opt1 = new Intl.Collator("de-u-kn").resolvedOptions();
var opt2 = new Intl.Collator("de-u-kn-true").resolvedOptions();

print(`locale=${opt1.locale}, caseFirst=${opt1.numeric}`);
print(`locale=${opt2.locale}, caseFirst=${opt2.numeric}`);
```

The test case should print this per the current spec:
```
locale=de, caseFirst=true
locale=de-u-kn-true, caseFirst=true
```

I don't understand why the resolved locale of the first collator is not set to either `de-u-kn` or `de-u-kn-true`. If the current behaviour is not a spec bug, can we add a comment to clarify why Unicode extensions keys without values are removed from the resolved locale?

littledan:
Huh, that's interesting.

Would fixing this be a matter of adding a simple step as ResolveLocale 8.h.ii.2.b Let _supportedExtensionAddition_ be the concatenation of "-" and _key_ ? If so, that sort of change sounds right to me.

(Aside: Wow, there's a lot of variation among implementations for how the locale presents these tags.)

srl295:
> `caseFirst=${opt1.numeric}`

should be `numeric=`?

littledan:
Rather than just making this one fix (which, in isolation, would seem good), let's see if we can make this part of [a broader change](https://github.com/tc39/proposal-intl-locale/issues/31).

PR #302, opened 2018-12-11
Add step ResolveLocale 8.h.ii.2.b

FrankYFTang:
@littledan @anba @gsathya @Ms2ger 
Close https://github.com/tc39/ecma402/issues/223

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -200,6 +200,7 @@
                   1. Let _supportedExtensionAddition_ be the concatenation of `"-"`, _key_, `"-"`, and _value_.
               1. Else if _keyLocaleData_ contains `"true"`, then
                 1. Let _value_ be `"true"`.
+                1. Let _supportedExtensionAddition_ be the concatenation of `"-"` and _key_.
           1. If _options_ has a field [[&lt;_key_&gt;]], then
             1. Let _optionsValue_ be _options_.[[&lt;_key_&gt;]].
             1. Assert: Type(_optionsValue_) is either String, Undefined, or Null.


FrankYFTang:
Ping

FrankYFTang:
> Let's get tests out for review in test262, and double-check that the committee wants this change.

I am not sure what you are suggesting here.  Currently, with the spec without this PR, the following test in test262
test/intl402/Collator/missing-unicode-ext-value-defaults-to-true.js
expect the collator.resolvedOptions().locale without "-kn" for "en-u-co-phonebk-kn" or "en-u-kn-co-phonebk"

        assert.sameValue(locale.indexOf("-kn"), -1, "\"kn\" is returned in locale, but shouldn't be.");

I don't think that make sense. Since in the same test it also expect collator.resolvedOptions().numeric to be true.

I can work on a PR to change test/intl402/Collator/missing-unicode-ext-value-defaults-to-true.js AFTER we merge this PR first.

littledan:
Yes, I agree the current spec is weird. I would just like to get broader buy-in on this change before merging.

FrankYFTang:
@zbraniecki

FrankYFTang:
ping

FrankYFTang:
so my understand from our meeting today is we will take this one line change in the spec and lead the #311 for future consideration. Can we merge this now?

littledan:
That matches my understanding as well; merging this patch

rkirsling:
Am I correct in understanding that this change was only meant to address `de-u-kn` not losing its `-u-kn`, and was not meant to affect the `de-u-kn-true` case? Because that's what #223 says, and that's what the spec appears to say, but it's not what tc39/test262#1995 says—the test was changed to require that the explicit `-true` of `de-u-kn-true` be dropped, which seems very unexpected.

anba:
The test262 PR may have been conflating #223 with the canonicalisation changes from `Intl.Locale`? But dropping `-true` is now correct according to the current draft spec, because we're now performing [UTS35 canonicalisation](https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers), which removes `"true"` values from all `type` productions.

rkirsling:
Thanks for the clarification, @anba! Looks like #406 got merged in the time since I asked the question above. 😅Glad that it's ultimately an accurate test.

PR #305, opened 2019-01-11
Create agenda-2019-01-17.md

FrankYFTang:


littledan:
This looks like a good start, but it'd be good to list open issues on each of the proposals that we need to work through. @sffc were you working on an agenda?

littledan:
Anyway, I'll merge this and we can work from there as follow-on PRs.

Issue #306, opened 2019-01-12
misleading to return (min|max)imumFractionDigits in resolvedOptions while we also return (min|max)imumSignificantDigits


FrankYFTang:
The ECMA 402 spec in the Intl.NumberFormat section currently said this

https://ecma-international.org/ecma-402/#sec-setnfdigitoptions
12. Set intlObj.[[MinimumFractionDigits]] to mnfd.
13. Set intlObj.[[MaximumFractionDigits]] to mxfd.
14. If mnsd is not undefined or mxsd is not undefined, then
  a. Let mnsd be ? DefaultNumberOption(mnsd, 1, 21, 1).
  b. Let mxsd be ? DefaultNumberOption(mxsd, mnsd, 21, 21).
  c. Set intlObj.[[MinimumSignificantDigits]] to mnsd.
  d. Set intlObj.[[MaximumSignificantDigits]] to mxsd.

So MinimumFractionDigits and MaximumFractionDigits are always stored into intlObj but MinimumSignificantDigits and MaximumSignificantDigits are not always.

and later on in https://ecma-international.org/ecma-402/#sec-formatnumberstring

1. If intlObject.[[MinimumSignificantDigits]] and intlObject.[[MaximumSignificantDigits]] are both not undefined, then
 a. Let result be ToRawPrecision(x, intlObject.[[MinimumSignificantDigits]], intlObject.[[MaximumSignificantDigits]]).
2. Else,
 a. Let result be ToRawFixed(x, intlObject.[[MinimumIntegerDigits]], intlObject.[[MinimumFractionDigits]], intlObject.[[MaximumFractionDigits]]).

so EITHER the xxxFractionDigits OR the xxxSignificantDigits setting will be used to format per object, but NOT BOTH.

But in 
https://ecma-international.org/ecma-402/#sec-intl.numberformat.prototype.resolvedoptions
It will report ALL the field from it, which means MinimumFractionDigits and MaximumFractionDigits will always be returned, but I think that will confused the users and harder for my new implementation to compliance with it. 

Does it make sense for us to change it to

12. If mnsd is not undefined or mxsd is not undefined, then
  a. Let mnsd be ? DefaultNumberOption(mnsd, 1, 21, 1).
  b. Let mxsd be ? DefaultNumberOption(mxsd, mnsd, 21, 21).
  c. Set intlObj.[[MinimumSignificantDigits]] to mnsd.
  d. Set intlObj.[[MaximumSignificantDigits]] to mxsd.
[...]

PR #307, opened 2019-01-12
Skip FractionDigits if we set SignificantDigits

FrankYFTang:
See https://github.com/tc39/ecma402/issues/306
This will impact the resolvedOptions()


FrankYFTang:
During the 2019-1-17 ECMA402 meeting, we discussed and decided this kind of minor tweaks won't get into ECMA 402 except through the Unified Number Format proposal Shane is working on. And this issue is already there. Close this issue therefore.

PR #310, opened 2019-01-12
Revert "Proposal agenda"

littledan:
Reverts tc39/ecma402#308


FrankYFTang:
?

sffc:
I think what happened is that I included #308 in my PR #309 so when both got merged the items were duplicated.

Issue #311, opened 2019-01-12
strip out u-extension in the locale returned by resolvedOptions()


FrankYFTang:
Currently the locale returned by resolvedOptions() of Intl objects has strange information. 
Often time the information on the u-extension of the locale is redundant to other property returned by the  resolvedOptions(). For example, if we create a Intl.Collator with the following 6 conditions
A. "en"
B. "en", {  caseFirst: "upper"}
C. "en-u-kn-upper", {  caseFirst: "upper"}
D. "en-u-kn-lower", {  caseFirst: "upper"} .   <= notice a conflict between the locale and option
E. "en-u-kn-upper",
F. "en-u-kn-lower",

the locale , caseFirst of resolvedOptions, by the current shape of the spec, will be
A. "en", undefined
B. "en", "upper"
C. "en-u-kn-upper", "upper"
D. "en", "upper" .                         <= notice we already strip the "-u-kn-lower" out when we have conflict.
E. "en-u-kn-upper", "upper"
F. "en-u-kn-lower", "lower"

In v8, we found it is hard to implement "EFFICIENTLY" this behavior for resolvedOptions() without increase of memory because it is hard read information from ICU to figure out the distinguishing between them. I would suggest we report locale without the u-extension since these information is already covered by the property. So I would like to see our spec change to make the output became

A. "en", undefined
B. "en", "upper"
C. "en", "upper"
D. "en", "upper"
E. "en", "upper"
F. "en", "lower"
instead, the developer won't miss any important information and it is.
If we have concern of blanking remove all u-extension from the reported locale could be harmful, we could just remove those key in the RelativeExtensionKeys only since those will always have another property in the return object of resolvedOptions()

@gsathya @littledan @zbraniecki @anba

littledan:
We discussed this in the ECMA-402 meeting. It seems like stripping these things out only makes sense once all tags have corresponding options, which we don't have at the moment. We had some reasons to not represent certain options (e.g., currency, timezone) as tags, but I don't think we had any strong objections to build this in the other direction, though. @NorbertLindenberg contacted me recently to explain that he's no longer involved in TC39, so I don't think we can get more information from discussing it with him.

FrankYFTang:
[...]

PR #312, opened 2019-01-12
Add #311 and move # 302 down to group w/ it.

FrankYFTang:
Restructure to add a "locale returned by resolvedOptions()" item and move #311  and #302 under it.


FrankYFTang:
@littledan

PR #318, opened 2019-01-25
Normative: Split out BigInt formatting functions

littledan:
Two new methods are added to Intl.NumberFormat.prototype:
- formatBigInt
- formatBigIntToParts

The earlier format/formatToParts methods are reverted to
accept Number and not BigInt.

Replacement for #236 

---

Closes #236

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -134,7 +134,35 @@
       <emu-alg>
         1. Let _x_ be ? thisNumberValue(*this* value).
         1. Let _numberFormat_ be ? Construct(%NumberFormat%, &laquo; _locales_, _options_ &raquo;).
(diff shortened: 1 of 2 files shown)


FrankYFTang:
If this is a REPLACEMENT of #236 then should we close #236 ?

leobalter:
I edited the top post with the "Closes ..." hint so GH would auto close the referenced PR when this gets merged.

Should we wait to have a decision in the meeting?

littledan:
Yes, I would like to wait for a conclusion in TC39.

sffc:
Is it safe to close this PR now that the committee agreed to move forward with the original idea, #236?

littledan:
@sffc Yes, closing per https://github.com/tc39/ecma402/pull/318#issuecomment-477441201

PR #327, opened 2019-02-28
Add package-lock.json

leobalter:
I'd like to avoid listing this file as untracked after each npm install usage.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


sffc:
I'm not personally much a fan of package-lock.json files except for applications that really need them.  This repository uses package.json only for tooling, so it doesn't benefit much from package-lock.json.  The file itself is large and adding it to version control will result in large diffs whenever anyone installs new versions of packages.

How about if you [disable package-lock.json](https://docs.npmjs.com/misc/config#package-lock) via .npmrc?  Create a .npmrc file with the following contents:

```
package-lock=false
```

ljharb:
Another alternative is adding `package-lock=false` to `.npmrc`; then npm won't try to create one. It won't really serve much of a purpose here except to make ecmarkup updates not show up for free.

leobalter:
Works for me. I can do it in my own .npmrc but here is a last question: should we share the .npmrc file here?

ljharb:
Definitely put it in `.npmrc` inside the repo; each npmrc file is read and merged upwards until it gets to `/`.

leobalter:
> You'll also want to gitignore `package-lock.json`

`.npmrc` should avoid it already, right? I'd like to hold on to these others for now unless they appear as issues.

ljharb:
It will, but sometimes they still slip into PRs. Doing both is a more certain practice.

PR #329, opened 2019-03-01
Editorial: master is now ES2020

leobalter:


leobalter:
> Hmm, I wonder if it would make sense to make some (most?) of these references unversioned.

It makes sense but for now I'll just unblock this. While a search for "2019" was not hard over the project, along with some observation where the changes are necessary, it's good to minimize this over some template job in the html files.

PR #332, opened 2019-03-14
Create agenda page for March 2019

leobalter:
cc @sffc @littledan 

Not sure who to set as contact in this agenda, so far I just did a copy and paste from the previous agenda.


sffc:
This is fine, I was going to make it also but I'll update it as needed.

littledan:
Thanks for making this. Going forward, are one of you interested in being the contact person for invitations?

sffc:
I was thinking about that.  I don't really like having a single point of failure.  Could we set up an alias, like ecma402-admin@chromium.org, and have it send to all three of us and anyone else who's interested?

littledan:
@sffc Sounds like a great idea. Go for it!

Issue #330, opened 2019-03-13
Various Unicode BCP 47 locale identifiers issues


anba:
[6.2 Language Tags](https://tc39.github.io/ecma402/#sec-language-tags)
- The link to UTS 35 should use https instead of http.
- "[...] identifies locales using language tags as defined by IETF BCP 47 (RFCs 5646 and 4647 or their successors)" should be changed to refer to Unicode BCP 47 locale identifiers exclusively.
- "Unicode BCP 47 Locale Identifiers that meet those validity criteria of Unicode Technical Standard 35, section 3.2 [....]" needs to be reworded, because "validity" can now be misunderstood to mean "validity" as specified in UTS 35 (cf. the "Validity / Comments" column in UTS 35). 
  - IIUC "structurally valid" in ECMA-402 maps to "syntactically well-formed" in UTS 35.
- "[...] without reference to the IANA Language Subtag Registry" may no longer be needed resp. it should be said, that ECMA-402 considers those languages tags as valid which match the syntax of Unicode BCP 47 locale identifiers, but that it is not required to validate them according to the Unicode validation data. (For example "aaj" is a valid language tag in ECMA-402 even though "aaj" is not included in <https://unicode.org/repos/cldr/tags/latest/common/validity/language.xml>.)

-------------------------

[6.2.1 Unicode Locale Extension Sequences](https://tc39.github.io/ecma402/#sec-unicode-locale-extension-sequences)
- The definition should be changed to refer to `unicode_locale_extensions` from <https://unicode.org/reports/tr35/#Unicode_locale_identifier>.

-------------------------

[6.2.2 IsStructurallyValidLanguageTag](https://tc39.github.io/ecma402/#sec-isstructurallyvalidlanguagetag)
- The next revision of UTS 35 will remove the ABNF grammar, so `IsStructurallyValidLanguageTag` will need to refer to the EBNF grammar.
 Ref: http://unicode.org/repos/cldr/trunk/specs/ldml/tr35.html#Unicode_language_identifier
- It should be clarified whether or not "does not include duplicate variant subtags" also applies to variant subtags in `tlang`. For example is "en-t-en-emodeng-emodeng" valid or not?

---------------------------

[6.2.3 CanonicalizeLanguageTag](https://tc39.github.io/ecma402/#sec-canonicalizelanguagetag)
[...]

PR #333, opened 2019-03-14
Normative: Apply recommended updates for BCP 47 Locale Identifiers

leobalter:
Ref #330 

cc @anba @sffc @FrankYFTang 

This is a tentative to update the immediate parts of #330, please let me know WDYT?

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -21,19 +21,18 @@
     <h1>Language Tags</h1>
 
     <p>
-      The ECMAScript 2020 Internationalization API Specification identifies locales using language tags as defined by IETF BCP 47 (RFCs 5646 and 4647 or their successors), which may include extensions such as those registered through RFC 6067. Their canonical form is that of a Unicode BCP 47 Locale Identifier, as specified in <a href="http://unicode.org/reports/tr35/#BCP_47_Conformance">Unicode Technical Standard #35 LDML § 3.3 BCP 47 Conformance</a>.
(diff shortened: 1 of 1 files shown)


spec/locales-currencies-tz.html
```diff
@@ -51,7 +50,7 @@ <h1>IsStructurallyValidLanguageTag ( _locale_ )</h1>
       </ul>
 
       <p>
-        The abstract operation returns true if _locale_ can be generated from the ABNF grammar in section 3.2 of the Unicode Technical Standard 35, or successor, starting with unicode_locale_id, and does not contain duplicate variant or singleton subtags (other than as a private use subtag). It returns false otherwise. Terminal value characters in the grammar are interpreted as the Unicode equivalents of the ASCII octet values given.
+        The abstract operation returns true if _locale_ can be generated from the EBNF grammar in section 3.2 of the Unicode Technical Standard 35, or successor, starting with unicode_locale_id, and does not contain duplicate variant or singleton subtags (other than as a private use subtag). It returns false otherwise. Terminal value characters in the grammar are interpreted as the Unicode equivalents of the ASCII octet values given.
```
littledan: I guess ABNF or EBNF both work here, as Unicode defines both grammars for some reason.

FrankYFTang: NO, we should surely switch to EBNF because Mark Davis is deleting ABNF from UTS35 now. The update will be pushed out very soon. See http://unicode.org/repos/cldr/trunk/specs/ldml/tr35.html#Unicode_Language_and_Locale_Identifiers
the ABNF was very buggy (see http://unicode.org/cldr/trac/ticket/11746 and more)
The EBNF on that latest version still have some minor problems and I am reviewing for Mark Davis now. He will fix them shortly.

FrankYFTang:
LGTM

Issue #299, opened 2018-11-26
Permit the use of tailorings NOT in CLDR


srl295:
- Followon to https://github.com/tc39/ecma402/pull/291
- Rather than just permitting looking at CLDR, we should allow implementations to have their own tailorings— with CLDR being recommended, of course, but not restricting implementations.  Look at how CLDR itself and Unicode structure conformance clauses for example text.
- Consider some front (or back matter, https://tc39.github.io/ecma402/#annex-implementation-dependent-behaviour ) to discuss this kind of tailoring, "CLDR or local tailioring" so that it can be referred to or at least not repeated many times in the spec.

srl295:
Discussed during https://github.com/tc39/ecma402/blob/master/meetings/agenda-2018-11-26.md

littledan:
Thanks for filing the issue. Yes, we'll need a PR for this. Any takers?

anba:
Do we plan to require that at least the tailorings defined in SpecialCasing.txt must be supported? Because if any implementation-defined tailoring is allowed, implementations could choose to not support any locale-specific tailoring, which would mean the tests in <https://github.com/tc39/test262/tree/master/test/intl402/String/prototype/toLocaleLowerCase> and <https://github.com/tc39/test262/tree/master/test/intl402/String/prototype/toLocaleUpperCase> need to be removed, because they expect that the locale-specific tailorings from SpecialCasing.txt are implemented.

littledan:
@anba About removing tests: We could instead use the test262 `locale:` metadata to indicate that they contain assumptions about the locale data, rather than removing the tests. We should probably do this whether or not we require SpecialCasing.txt, since the test will be effectively checking that there are no *additional* tailorings.

spectranaut:
fixed in #341

PR #341, opened 2019-04-11
Normative: Permit use of implementation-defined tailoring for case mapping

spectranaut:
This is a new PR after feedback from @littledan's original fix #291 

The feedback in #299 and discussed in the [2018-11-26 meeting](https://github.com/tc39/ecma402/blob/b7ace4ce6d6fe6598f3f16536d42f7c8fd95258e/meetings/notes-2018-11-26.md) was to:
1. Allow any custom tailoring, not only the Unicode Database.
2. Mandate the implementations consistently use the same case mapping for the same locale.

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -65,27 +65,21 @@
         1. Else,
           1. Let _requestedLocale_ be DefaultLocale().
         1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with all Unicode locale extension sequences (<emu-xref href="#sec-unicode-locale-extension-sequences"></emu-xref>) removed.
-        1. Let _availableLocales_ be a List with the language tags of the languages for which the Unicode character database contains language sensitive case mappings.
+        1. Let _availableLocales_ be a List with the language tags of the languages for which the Unicode Character Database or CLDR contains language sensitive case mappings.
         1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
         1. If _locale_ is *undefined*, let _locale_ be `"und"`.
         1. Let _cpList_ be a List containing in order the code points of _S_ as defined in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, starting at the first element of _S_.
-        1. For each code point _c_ in _cpList_, if the Unicode Character Database provides a lower case equivalent of _c_ that is either language insensitive or for the language _locale_, replace _c_ in _cpList_ with that/those equivalent code point(s).
-        1. Let _cuList_ be a new empty List.
-        1. For each code point _c_ in _cpList_, in order, append to _cuList_ the elements of the UTF-16 Encoding (defined in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) of _c_.
+        1. Let _cuList_ be a List where the elements are the result of toLowercase(_cpList_), according to a tailored version of the Unicode Default Case Conversion algorithm, with tailorings based on _locale_ take into account based on data in SpecialCasings.txt and/or CLDR, per the Unicode Standard 3.13.
         1. Let _L_ be a String whose elements are, in order, the elements of _cuList_.
         1. Return _L_.
       </emu-alg>
 
       <p>
-        The result must be derived according to the case mappings in the Unicode character database (this explicitly includes not only the UnicodeData.txt file, but also the SpecialCasings.txt file that accompanies it).
+        The result must be derived according to the case mappings in the Unicode Standard. (This explicitly includes not only the UnicodeData.txt file, but also the SpecialCasings.txt file that accompanies it, as well as any tailorings from CLDR.)
       </p>
 
       <emu-note>
-        As of Unicode 10.0, the _availableLocales_ list contains the elements `"az"`, `"lt"`, and `"tr"`.
-      </emu-note>
-
-      <emu-note>
-        The case mapping of some code points may produce multiple code points. In this case the result String may not be the same length as the source String. Because both *toLocaleUpperCase* and *toLocaleLowerCase* have context-sensitive behaviour, the functions are not symmetrical. In other words, *s.toLocaleUpperCase().toLocaleLowerCase()* is not necessarily equal to *s.toLocaleLowerCase()*.
+        The case mapping of some code points may produce multiple code points. In this case the result String may not be the same length as the source String. Because both `toLocaleUpperCase` and `toLocaleLowerCase` have context-sensitive behaviour, the functions are not symmetrical. In other words, `s.toLocaleUpperCase().toLocaleLowerCase()` is not necessarily equal to `s.toLocaleLowerCase()`.
       </emu-note>
 
       <emu-note>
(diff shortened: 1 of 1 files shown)


spectranaut:
@anba @srl295 @littledan -- can you all take a look?

spec/locale-sensitive-functions.html
```diff
@@ -65,27 +65,23 @@ <h1>String.prototype.toLocaleLowerCase ( [ _locales_ ] )</h1>
         1. Else,
           1. Let _requestedLocale_ be DefaultLocale().
         1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with all Unicode locale extension sequences (<emu-xref href="#sec-unicode-locale-extension-sequences"></emu-xref>) removed.
-        1. Let _availableLocales_ be a List with the language tags of the languages for which the Unicode character database contains language sensitive case mappings.
+        1. Let _availableLocales_ be a List with language tags that includes the languages for which the Unicode Character Database contains language sensitive case mappings. Implementations may add additional language tags if they support custom tailoring.
```
littledan: I'd delete " if they support custom tailoring". The phrasing I'd suggest, and which much of the spec tends towards, implementations maintain their own locale database, rather than talking about "tailoring".

spec/locale-sensitive-functions.html
```diff
@@ -65,27 +65,23 @@ <h1>String.prototype.toLocaleLowerCase ( [ _locales_ ] )</h1>
...
         1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
         1. If _locale_ is *undefined*, let _locale_ be `"und"`.
         1. Let _cpList_ be a List containing in order the code points of _S_ as defined in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>, starting at the first element of _S_.
-        1. For each code point _c_ in _cpList_, if the Unicode Character Database provides a lower case equivalent of _c_ that is either language insensitive or for the language _locale_, replace _c_ in _cpList_ with that/those equivalent code point(s).
+        1. For each code point _c_ in _cpList_, replace _c_ with a lower case equivalent of _c_ that is either language insensitive or defined according to the _locale_ specific code point mapping. Implementations must always be consistent in the case mapping they use for a specified locale.
```
littledan: This "consistent" wording is interesting, but it could be a little confusing. Could you use phrasing that indicates, the output of this method is a function of the locale data and its inputs? E.g., "In a particular implementation, the transformation of _cpList_ always yields the same result, for any fixed _cpList_ and locale." (Maybe you have ideas for making that clearer.)

Note, the "for each code point" wording here is misleading; I fixed it in ECMA-262 in https://github.com/tc39/ecma262/pull/932/files , and we should probably apply the same fix in ECMA-402. (Could be a separate PR, or the same one, as you're already touching this line.)

spec/locale-sensitive-functions.html
```diff
@@ -65,27 +65,23 @@ <h1>String.prototype.toLocaleLowerCase ( [ _locales_ ] )</h1>
...
         1. Let _cuList_ be a new empty List.
         1. For each code point _c_ in _cpList_, in order, append to _cuList_ the elements of the UTF-16 Encoding (defined in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>) of _c_.
         1. Let _L_ be a String whose elements are, in order, the elements of _cuList_.
         1. Return _L_.
       </emu-alg>
 
       <p>
-        The result must be derived according to the case mappings in the Unicode character database (this explicitly includes not only the UnicodeData.txt file, but also the SpecialCasings.txt file that accompanies it).
+        Lower case code point mappings may be derived according to a tailored version of the Default Case Conversion Algorithms of the Unicode Standard. Implementaitons may use locale specific tailoring defined in SpecialCasings.txt and/or CLDR and/or any other custom tailoring.
```
littledan: Nit: Implementations

(I feel like it's OK to use the word "tailoring" in this section, but I'm not sure if SpecialCasings.txt is considered a tailoring or not.)

spec/locale-sensitive-functions.html
```diff
@@ -65,27 +65,23 @@ <h1>String.prototype.toLocaleLowerCase ( [ _locales_ ] )</h1>
...
       </p>
 
       <emu-note>
-        As of Unicode 10.0, the _availableLocales_ list contains the elements `"az"`, `"lt"`, and `"tr"`.
```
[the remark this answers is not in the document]
spectranaut: There are more locales in unicode that have case mapping than those three -- it seems misleading to only mention them? Do you think I should include a list of all of them?

spec/locale-sensitive-functions.html
spectranaut: @leobalter also thought I should use language more like what you just suggested, @littledan -- but I only see the word "database" used in the spec when referring to "Unicode Character Database" or "IANA Time Zone Database". I feel like I would be confused if I saw a reference suddenly to a "implementation maintained locale database". But then again, I am new to reading this spec :)

What about not using the phrase "custom tailoring":
`Implementations may add additional language tags if they support case mapping for additional locales.`

spec/locale-sensitive-functions.html
spectranaut: I like your replacement for "consistent" - I'll add it.

Do you think code points is used clearly in many places in this spec? If so can you file and issue? :)

spec/locale-sensitive-functions.html
spectranaut: The Default Case Conversion Algorithm (unicode 3.13) calls it a "tailoring" so I think we are ok there.

spec/locale-sensitive-functions.html
littledan: I am not aware of other instances of this error outside of case mapping.

spec/locale-sensitive-functions.html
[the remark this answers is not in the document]
littledan: Good to know. I am fine with removing this list.

spec/locale-sensitive-functions.html
spectranaut: I think I didn't understand what was misleading about the "for each code point" until now, but I see -- the mapping isn't always directly from on code point to another.

spectranaut:
@leobalter can you also give one more look?

spec/locale-sensitive-functions.html
```diff
@@ -102,7 +96,7 @@ <h1>String.prototype.toLocaleUpperCase ( [ _locales_ ] )</h1>
       </p>
 
       <p>
-        This function interprets a string value as a sequence of code points, as described in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. This function behaves in exactly the same way as `String.prototype.toLocaleLowerCase`, except that characters are mapped to their _uppercase_ equivalents as specified in the Unicode character database.
+        This function interprets a string value as a sequence of code points, as described in ES2020, <emu-xref href="#sec-ecmascript-language-types-string-type"></emu-xref>. This function behaves in exactly the same way as `String.prototype.toLocaleLowerCase`, except that characters are mapped to their _uppercase_ equivalents. Implementations must always be consistent in the case mapping they use for a specified locale.
```
littledan: I like the consistency wording you adopted above. Would it make sense to use something similar here as well?

spectranaut: done!

leobalter:
I'm not sure we can author tests for this so I'm happy to merge it after a rebase to organize the commits.

PR #345, opened 2019-05-20
Normative: Add quarter option

FrankYFTang:


FrankYFTang:
@zbraniecki @anba

FrankYFTang:
```js
let d = new Date();
var dtf = new Intl.DateTimeFormat("en", {year: "numeric", quarter: "short"});
dtf.format(d);  // "Q2 2019"
dtf.formatToParts(d);
// [{type: "quarter", value: "Q2"}, 
//  {type: "literal", value: " "}, 
//  {type: "year", value: "2019"}]
dtf = new Intl.DateTimeFormat("en", {year: "numeric", quarter: "long"});
dtf.format(d);  // "2nd quarter 2019"
dtf.formatToParts(d);
// [{type: "quarter", value: "2nd quarter"}, 
//  {type: "literal", value: " "}, 
//  {type: "year", value: "2019"}]
dtf = new Intl.DateTimeFormat("zh-Hant", {year: "numeric", quarter: "short"});
dtf.format(d);  // "2019年第2季"
dtf.formatToParts(d);
// [{type: "year", value: "2019"}, 
//  {type: "literal", value: "年"}, 
//  {type: "quarter", value: "第2季"}]

**Comment by sffc** [comment]
LGTM; let's discuss in the next meeting to seek consensus.

**Comment by littledan** [comment]
LGTM as well. Is the idea here to pursue this feature as a one-off, or did you have more Intl.DateTimeFormat feature proposals as well?

**Comment by FrankYFTang** [comment]
> LGTM as well. Is the idea here to pursue this feature as a one-off, or did you have more Intl.DateTimeFormat feature proposals as well?

That is the question I am not sure. There are 3 thing (or 4 if we count the relativeYear / yearName you proposed) we should address, all of them are kind of independent from each other:
2) another one is the "dayPeriod", which will have very similar change as this one. 
3) an option "millisecondDigits" (0 to 3 ) to control the output of how many digits for millisecondDigits (default 0 as now. )

 I wonder should I file one by one as PR or group them together as a proposal (see https://github.com/FrankYFTang/proposal-intl-datetimeformat-more-options/blob/master/README.md ). Should I also include the "yearName" into such proposal?

**Comment by FrankYFTang** [comment]
How about I just file all the changes one by one as PR here so we will have a sense of what the changes look like. And if we all feel it need to be group into a bigger proposal I can just merge those changes there.

**Comment by FrankYFTang** [comment]
v8 prototype in https://chromium-review.googlesource.com/c/v8/v8/+/1621445

**Comment by FrankYFTang** [comment]
We have discussed this PR in Jun 5 2019 TC39 and decided this is small enough and align with pre-existing API already and therefore does not need to form a staged proposal but do require consensus with in ECMA402 committee + tests in test262 + 1 js engine implementation and no concern from other js implementation to land into ECMC402 as stage 4.

**Comment by FrankYFTang** [comment]
Initial set of tests in https://github.com/tc39/test262/pull/2194
V8 is now having this behind a flag --harmony_intl_dateformat_quarter

**Comment by leobalter** [comment]
Question from the Intl meeting: 

> How does this API behave in non gregorian calendars?

**Comment by FrankYFTang** [comment]
> How does this API behave in non gregorian calendars?

Quarter, regardless which calendar system we are using, is defined as (month / 3) in zero-based quarter and zero-based month
month 0 (Jan) -> quarter 0 (Q1)
month 2 (March) -> quarter 0 (Q1)
month 3 (April) -> quarter 1 (Q2)
month 11 (Dec) -> quarter 3 (Q4)

I will make a spec change to clarify that quarter is calculated from month / 3.

> How do we handle extension to trimesters and more general periods in the future?
Other periods is not considered, neither will conflict with this proposal. For example, if there are strong demand, we can propose to add to format "decade" (e.g "80") , "century" ("21 century"), "semester" ("first semester"), or "trimester" and won't conflict with this. The scope of this PR is only tight to a frequently use period in the modern world: 
1)  in the financial world (2019 Q1 Financial report of Apple Computer Inc - Ex https://www.boxinvestorrelations.com/financial-information/financial-reports/default.aspx )
2) in company planning ( https://corvisio.com/wp-content/uploads/2019/04/OKR-Software-3.png ) 
maybe in other usage.

It is not intend to be a general abstract period that cover other concept in different calendar system.
The quarter is supported in CLDR but not other more general abstract period.

**Comment by FrankYFTang** [comment]
API behave in non gregorian calendars is now fully specified by https://github.com/tc39/ecma402/pull/345/commits/6c60ab7935585066ec0ca4970124d5b85c692d67

**Comment by srl295** [comment]
So we got feedback that in the UK there are ['financial years' which start April 1st](https://en.wikipedia.org/wiki/Fiscal_year#United_Kingdom), and companies use those quarters in their reporting, and so 'quarters' need to be calculated from a different start date than the calendar.

If the primary use for Quarters is financial, I think this may need some more thought.  I know this is a longstanding ICU/CLDR and Java feature. But if it is not sufficient for a large number of cases, then we should probably rethink whether this feature should be simply named `quarter: "short"`.

**Comment by srl295** [comment]
I've opened [CLDR-13109](https://unicode-org.atlassian.net/browse/CLDR-13109) to discuss these questions on the CLDR side.

**Comment by srl295** [comment]
> Quarter, regardless which calendar system we are using, is defined as (month / 3) in zero-based quarter and zero-based month

OK, but for say Hebrew calendar in a leap year, there are 13 months. so would that result in "5th Quarter"? (Note: There's no CLDR data for `<quarter type="5" />`) And actually, the numbering of the months internally (ICU fields + CLDR data) is something like 0,1,2,3,4,6,7,8,9,10,11,12 in non-leap years (skipping 5) and 0,1,2,3,4,5,6,7,8,9,10,11,12 in leap years.  How are you proposing to handle this?

**Comment by ashemedai** [comment]
FYI: In the Netherlands we have a normal [fiscal year](https://nl.wikipedia.org/wiki/Boekjaar) (_boekjaar_; 1 Jan – 31 Dec), but also a "broken" fiscal year, which can start on 1 Apr, 1 Jul. In addition a fiscal year might be shorter than 1 year but also longer, to a maximum of 24 months minus 1 day.

File: spec/datetimeformat.html [hunk 1]
```diff
@@ -281,7 +286,10 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
           1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
           1. If _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then
             1. Let _f_ be the value of _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the matching row.
-            1. Let _v_ be the value of _tm_'s field whose name is the Internal Slot column of the matching row.
+            1. If _p_ is `"quarter"`, then
+              1. Let _v_ be ( _tm_.[[month]] / 3 ) + 1.
```
srl295: assumes only 12 months

srl295:
@FrankYFTang can you comment on the use cases behind the proposal? I know it was added to CLDR 13 years ago, but it would be helpful to get the expected usage. If it is financial reports and company planning (above) I think that argues for at least a settable start date.

spec/datetimeformat.html
FrankYFTang: Not really, it assume every 3 months is a quarter and won't prevent month 13 to be 5th quarter in the calendar have more than 12 months.

Also according to 
https://www.ecma-international.org/ecma-262/9.0/index.html#sec-month-number
"Months are identified by an integer in the range 0 to 11, inclusive."

The ICU logic is in i18n/smpdtfmt.cpp
```
    case UDAT_QUARTER_FIELD:
        if (count >= 4)
            _appendSymbol(appendTo, value/3, fSymbols->fQuarters,
                          fSymbols->fQuartersCount);
        else if (count == 3)
            _appendSymbol(appendTo, value/3, fSymbols->fShortQuarters,
                          fSymbols->fShortQuartersCount);
        else
            zeroPaddingNumber(currentNumberFormat,appendTo, (value/3) + 1, count, maxIntCount);
        break;
```

FrankYFTang:
> So we got feedback that in the UK there are ['financial years' which start April 1st](https://en.wikipedia.org/wiki/Fiscal_year#United_Kingdom), and companies use those quarters in their reporting, and so 'quarters' need to be calculated from a different start date than the calendar.
> 
> > in the financial world (2019 Q1 Financial report of Apple Computer Inc )
> 
> Yes, but in the UK, here is a "1st Quarter" report that ends July 1st:
> 
> * [MARKS AND SPENCER GROUP PLC QUARTER 1 2017/18 TRADING STATEMENT 13 WEEKS TO 1 JULY 2017](https://corporate.marksandspencer.com/media/press-releases/2017/trading-statement-13-weeks-to-1-july-2017)
> 
> If the primary use for Quarters is financial, I think this may need some more thought. I know this is a longstanding ICU/CLDR and Java feature. But if it is not sufficient for a large number of cases, then we should probably rethink whether this feature should be simply named `quarter: "short"`.

If the year start with April 1st, instead of Jan 1st, then it is an issue of a different "calendar" rather than an issue with "quarter", right?

spec/datetimeformat.html
srl295: A quarter means a fourth of something. So a 5th quarter is not a quarter.

20.3.1.4 Month Number is also wrong for non-Gregorian calendars, then. because 0 to 11 inclusive is not sufficient. Perhaps you should spec this just for gregorian calendars?

Note my node (v8) example referenced above showing numeric formats from 1 to 13 inclusive:  https://gist.github.com/srl295/4ce6860667770b5cf563c383897d1b20

FrankYFTang:
> @FrankYFTang can you comment on the use cases behind the proposal? I know it was added to CLDR 13 years ago, but it would be helpful to get the expected usage. If it is financial reports and company planning (above) I think that argues for at least a settable start date. There's a 2013 CLDR ticket that notes that Japan also has an Apr 1 fiscal start date used with quarters, and also a [2008 ICU ticket that I filed](https://unicode-org.atlassian.net/browse/ICU-6158) (which I only wrote a test case for now) that shows that formatting is broken for ICU4J

I think it is very common for people to think 2019 Q1 as Jan-March of 2019 and this is what this feature intend to support. Of course, there are other demand for different kind of "calendar system" to be support, but that is beyond what I intend to change here. All what you talk about fiscal year is related to a different calendar system, since it also change the week number and day number start from that year. That is beyond this feature is about.

srl295:
> I think it is very common for people to think 2019 Q1 as Jan-March of 2019 and this is what this feature intend to support. 

Which people, and for what example usage? Have you surveyed different regions? As an international standard, we need to consider more than specific regional usage.  I think CLDR did not do this analysis in v1.4 when quarters were added, and should have.

> If the year start with April 1st, instead of Jan 1st, then it is an issue of a different "calendar" rather than an issue with "quarter", right? 

Possibly. But would month numbering be affected (i.e. April be month 0(1st)?) Or week numbering?

spec/datetimeformat.html
srl295: yes, the ICU logic for Java for quarter is broken for the Hebrew calendar, it produces garbage output.

srl295: Is the Ecma262 reference above binding on Ecma402 date formats? I wouldn't think so, otherwise non-12 month calendars would not be supported at all.

Crissov:
> Quarter, regardless which calendar system we are using, is defined as (month / 3) in zero-based quarter and zero-based month
> month 0 (Jan) -> quarter 0 (Q1)
> month 2 (March) -> quarter 0 (Q1)
> month 3 (April) -> quarter 1 (Q2)
> month 11 (Dec) -> quarter 3 (Q4)

No, a _quarter(-year)_ is a (solar) year divided into 4 time spans of almost equal length, usually rounded to an integer multiple of another time unit: this does not have to be (3) months, but can also be (91+) days or (13+) weeks!

> > How do we handle extension to trimesters and more general periods in the future?

A _trimester_, by its very name, is 3 months long (which confusingly is about a third of the length of a human pregnancy), whereas a _semester_ (**not** related to _semi_ as often assumed) is 6 months. _Bimester_  (2 months) and _quadmester_ (4 months) are analogous terms that are used less frequently. Therefore, the proposed definition `month / 3` is appropriate for _trimester_, but not for _quarter(-year)_. 

1. calendar season: _first trimester_ = Jan+Feb+Mar (90 or 91 days)
2. calendar season: _second trimester_ = Apr+May+Jun (91 days)
3. calendar season: _third trimester_ = Jul+Aug+Sep (92 days)
4. calendar season: _fourth trimester_ = Oct+Nov+Dec (92 days)

Alas, the first _&hellip;mester_ does not need to start with January (or whatever the first month is called), because these terms are most frequently used with fiscal and academic years that not necessarily equate to the calendar year. Often, a (summer) break, despite having the same length as teaching terms, is not considered a trimester, which adds to the confusion about the number of trimesters per year and hence their length. In much of Europe, the academic year starts either with September or with October, and the terms may have traditional or local names:

1. academic term: _Michaelmas, Fall, Autumn_ = Oct+Nov+Dec (92 days) or Sep+Oct+Nov (91 days)
2. academic term: _Hilary, Epiphany, Lent, Winter_ = Jan+Feb+Mar or Dec+Jan+Feb (both 90 or 91 days)
3. academic term: _Easter, Spring_ = Apr+May+Jun (91 days) or Mar+Apr+May (92 days)
4. academic term: _Trinity, Summer_ = Jul+Aug+Sep or Jun+Jul+Aug (both 92 days)

European _meteorological seasons_ are another example of offset trimesters:

1. meteo season: _winter_ = Dec+Jan+Feb (90 or 91 days)
2. meteo season: _spring_ = Mar+Apr+May (92 days)
3. meteo season: _summer_ = Jun+Jul+Aug (92 days)
4. meteo season: _autumn_ = Sep+Oct+Nov (91 days)

_Astronomic seasons_, lasting from an equinox to a solstice or vice versa (which may also be known by their pagan holidays *Ostara, Litha* (i.e. *Midsummer*), *Mabon* and *Yule*, have a similar length and contain about 3 zodiac signs (but not their eponymous constellations) each:

1. classical, tropical season: _spring_ = ~20 March Equinox – June Solstice (~92.75 ≤ 93 days) = ♈+♉+♊
2. classical, tropical season: _summer_ = 20–21 June Solstice – September Equinox (~93.65 ≥ 93 days) = ♋+♌+♍
3. classical, tropical season: _autumn_ = 22–23 September Equinox – December Solstice (~89.85 ≤ 90 days) = ♎+♏+♐
4. classical, tropical season: _winter_ = 21–22 December Solstice – March Equinox (~88.99 ≤ 89 days) = ♑+♒+♓

These _tropical seasons_ must not be confused with _wet and dry seasons_ or _wet, hot and cold seasons_ in the tropics. While the tropical year is commonly used in Western astrology as well, sometimes _astrological seasons_ are defined as parts of the sidereal year as in South Asia:

1. visible, sidereal season: _spring_ = mid-Apr – mid-Jul
2. visible, sidereal season: _summer_ = mid-Jul – mid-Oct
3. visible, sidereal season:  _autumn_ = mid-Oct – mid-Jan
4. visible, sidereal season:  _winter_ = mid-Jan – mid-Apr

Traditional (and sometimes still Julian) English _quarter days_, which are the origin of academic trimester names, usually are Christian holidays that lay between a solstice or equinox and the end of the month:

1. quarter day: _Lady Day_ (25 Mar)
2. quarter day: _Midsummer_ (24 Jun)
3. quarter day: _Michaelmas_ (29 Sep)
4. quarter day: _Christmas_ (25 Dec)

or in areas with stronger Celtic heritage, e.g. Scotland:

1. (cross-)quarter day: _Candlemas_ (2 Feb, *28 Feb)
2. (cross-)quarter day: _May Day / Roodmas_ (1 May) or nominal _Whitsunday_ (15 May, *28 May)
3. (cross-)quarter day: _Lammas_ (1 Aug, *28 Aug)
4. (cross-)quarter day: _All Hallows_ (1 Nov) or _Martinmas_ (11 Nov, *28 Nov)

_Solar seasons_, best known from Gaelic or Wiccan changeover feasts *Imbolc, Beltane, Lammas* and *Samhain* (i.e. *Halloween*), historically started at the midpoints between equinoxes and solstices – as they do in East Asia:

1. solar season: _spring_, _chūn_ (春) ≈ 4 Feb – 4 May
2. solar season: _summer_, _xià_ (夏) ≈ 5 May – 5 Aug
3. solar season: _autumn_, _qiū_ (秋) ≈ 6 Aug – 5 Nov
4. solar season: _winter_, _dōng_ (冬) ≈ 6 Nov – 3 Feb

They have since moved by less than a week to cover 3 complete Gregorian months each:

1. Gaelic season: _spring_ = Feb+Mar+Apr
2. Gaelic season: _summer_ = May+Jun+Jul
3. Gaelic season: _autumn_ = Aug+Sep+Oct
4. Gaelic season: _winter_ = Nov+Dec+Jan

> The scope of this PR is only tight to a frequently use period in the modern world:
> 
> 1. in the financial world (2019 Q1 Financial report of Apple Computer Inc - ...)
> 2. in company planning (...) maybe in other usage.

Apple's reporting quarters are infamously week-based (i.e. usually 13 weeks in each quarter, but every 5 to 6 years there is a 14th week in Q4), see e.g. [Slate](https://slate.com/business/2014/02/14-week-quarters-financial-markets-get-confused.html). The actual start of a country's or company's fiscal year varies, as does the definition of week start.

1. fiscal quarter = W01–W13
2. fiscal quarter = W14–W26
3. fiscal quarter = W27–W39
4. fiscal quarter = W40–W52 or W53

Old German _montane or mining quarters_, each named for a Christian holiday contained within itself and thus closely related to Catholic _embertides_, over time changed from being week-based (starting on a Sunday) to last 3 Gregorian calendar months each.

1. embertide: _Lent / Reminiscere_ (6/5 weeks before Easter: 15 February – 21 March, W07–W12) → Jan+Feb+Mar (~W01–W13)
2. embertide: _Whitsun / Trinitatis_ (7/8 weeks after Easter:  17 May – 20 June, W20–W25) → Apr+May+Jun (~W14–W26)
3. embertide: _Holyrood, Crucis_ (14 September, W37–W38) → Jul+Aug+Sep (~W27–W39)
4. embertide: _Luci(a)e_ (13 December, W50–W51) → Oct+Nov+Dec (~W40–W52)

In conclusion, it probably makes sense to specify *at least* three terms, e.g.:

- _trimester_: 3 (out of 12) consecutive calendar months, starting with the first of the calendar year
- _quarter_: a fourth of a (solar) calendar year, usually 91 days = 13 weeks, exceptionally 92 days or 14 weeks (= 98 days)
- _season_: 3 (out of 12) consecutive zodiac signs, starting with the first of the astronomical year

FrankYFTang:
> In conclusion, it probably makes sense to specify _at least_ three terms, e.g.:
> 
> * _trimester_: 3 (out of 12) consecutive calendar months, starting with the first of the calendar year
> * _quarter_: a fourth of a (solar) calendar year, usually 91 days = 13 weeks, exceptionally 92 days or 14 weeks (= 98 days)
> * _season_: 3 (out of 12) consecutive zodiac signs, starting with the first of the astronomical year

Thank you for your information. It is very informative. 
Could we agree of the following?
A. The supporting for "quarter" should not be mixed with the supporting for trimester and season. In other words, we should not introduce a concept/term to represent any of these three different concept but should instead, keep them separated if we add the support.
B. The calculation of when such "quarter" start/end depending on each calendar system.
C. The calculation of when such "quarter" start/en may also require customization within the same calendar system for some usage. 
D. There are no need to support all the three concepts (quarter/trimester/season) in the same time (same revision of spec or vai same PR). In other words, we could choose to add the support to format one of them first and others later.
E. The current PR need to be changed to ensure B and C above won't be restricted to a particular calendar and allows future customization of start/end time of quarter in a particular calendar system.

sffc:
Since this PR is not actionable until the open questions are addressed, I am converting it to an issue, as discussed in ECMA-402 on September 5.  Please direct future discussion to #374.

PR #348, opened 2019-05-23
Editorial: clarification how to calculate gregorian date and time values

spectranaut:
This PR should not change the outcome of this abstract operation. The intent of these changes is only to make clear which algorithms to use for which values from [ECMA-262's date-values-from-time algorithms](https://tc39.github.io/ecma262/#sec-overview-of-date-objects-and-definitions-of-abstract-operations).

Here is the html for easy reading:
https://spectranaut.github.io/ecma402/#sec-tolocaltime

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -372,17 +372,19 @@
 
       <emu-alg>
         1. Assert: Type(_t_) is Number.
-        1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
-        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>. In these calculations, use the value _tz_, which is the time value _t_ + _timeZoneOffset_. The value _timeZoneOffset_ must be calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xre> where `local time zone` is replaced with timezone `timeZone`.
-          <ul>
-            <li>The value for weekday must be the result of `WeekDay(tz)` specified in ES2020's <emu-xref href="#sec-week-day">Week Day</emu-xref>.</li>
-            <li>The value for year must be the result of `YearFromTime(tz)` specified in ES2020's <emu-xref href="#sec-year-number">Year Number</emu-xref>.</li>
-            <li>The value for month must be the result of `MonthFromTime(tz)` specified in ES2020's <emu-xref href="#sec-month-number">Month Number</emu-xref>.</li>
-            <li>The value for date must be the result of `DateFromTime(tz)` specified in ES2020's <emu-xref href="#sec-date-number">Date Number</emu-xref>.</li>
-            <li>The value for hour must be the result of `HourFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
-            <li>The value for minute must be the result of `MinuteFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
-            <li>The value for second must be the result of `SecondFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
-          </ul>
+        1. If _calendar_ is `"gregory"`,
+          1. Apply calendrical calculations on _t_ to produce values for weekday, year, month, day, hour, minute and second according to the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>. In these calculations, use the value _tz_, which is the time value _t_ + _timeZoneOffset_. The value _timeZoneOffset_ must be calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xref> where the local time zone is replaced with timezone _timeZone_.
+            <ul>
+              <li>The value for weekday must be the result of `WeekDay(tz)` specified in ES2020's <emu-xref href="#sec-week-day">Week Day</emu-xref>.</li>
+              <li>The value for year must be the result of `YearFromTime(tz)` specified in ES2020's <emu-xref href="#sec-year-number">Year Number</emu-xref>.</li>
+              <li>The value for month must be the result of `MonthFromTime(tz)` specified in ES2020's <emu-xref href="#sec-month-number">Month Number</emu-xref>.</li>
+              <li>The value for date must be the result of `DateFromTime(tz)` specified in ES2020's <emu-xref href="#sec-date-number">Date Number</emu-xref>.</li>
+              <li>The value for hour must be the result of `HourFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
+              <li>The value for minute must be the result of `MinuteFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
+              <li>The value for second must be the result of `SecondFromTime(tz)` specified in ES2020's <emu-xref href="#sec-hours-minutes-second-and-milliseconds">Hours, Minutes, Second, and Milliseconds</emu-xref>.</li>
+            </ul>
+        1. Else,
+          1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
         1. Return a Record with fields [[weekday]], [[era]], [[year]], [[month]], [[day]], [[hour]], [[minute]], [[second]], and [[inDST]], each with the corresponding calculated value.
       </emu-alg>
 
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -373,15 +373,15 @@
       <emu-alg>
         1. Assert: Type(_t_) is Number.
         1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
-        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>:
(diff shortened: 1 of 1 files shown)


spectranaut:
@FrankYFTang can you take a look because of your familiarity with this part of the spec?

spec/datetimeformat.html
```diff
@@ -364,14 +364,25 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+        1. Assert: Type(_t_) is Number.
+        1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
+        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>:
```
FrankYFTang: If you want to further link the calculation for Gregory calendar to those abstract operation in ES2020, then you must also first adjust the _t_ based on the offset of the timezone and the inDST . Otherwise, the result will always be in UTC and that is not what ToLocalTime should do.

littledan: Just to clarify, will "gregory" always be the proleptic Gregorian calendar, or should we allow for the local transition from the Julian calendar?

spec/datetimeformat.html
```diff
@@ -373,15 +373,15 @@ <h1>ToLocalTime ( _t_, _calendar_, _timeZone_ )</h1>
...
-        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>:
+        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>. In these calculations, use the value _tz_, which is the time value _t_ + _timeZoneOffset_. The value _timeZoneOffset_ must be calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xre> where `local time zone` is replaced with timezone `timeZone`.
```
spectranaut: @FrankYFTang can you review this prose? Unfortunately all timezone offset calculations only refer to "local time" in ECMA-262, which makes it a little awkward to use.

leobalter: pinging @FrankYFTang. Also @sffc for potentially more feedback

spec/datetimeformat.html
```diff
@@ -364,14 +364,25 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+        1. Assert: Type(_t_) is Number.
+        1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
+        1. If the _calendar_ is `"gregory"`, then the calculations of weekday, year, month, day, hour, minute and second must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>. In these calculations, use the value _tz_, which is the time value _t_ + _timeZoneOffset_. The value _timeZoneOffset_ must be calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xre> where `local time zone` is replaced with timezone `timeZone`.
+          <ul>
```
sffc: Why mixing HTML with Markdown?

spec/datetimeformat.html
```diff
@@ -364,14 +364,25 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+        When the ToLocalTime abstract operation is called with arguments _t_, _calendar_, and _timeZone_, the following steps are taken:
       </p>
 
       <emu-alg>
-        1. Apply calendrical calculations on _date_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules. If the _calendar_ is `"gregory"`, then the calculations must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>.
+        1. Assert: Type(_t_) is Number.
+        1. Apply calendrical calculations on _t_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.
```
sffc: How about something more like,

```
1. Assert: Type(_t_) is Number.
1. If _calendar_ is `"gregory"`,
  1. Return a Record with fields calculated according to <reference to table>
1. Else,
  1. Return a Record with the fields in Column 1 of <reference to table>, calculated by (insert hand-wavy prose here).
```

spec/datetimeformat.html
leobalter: if I'm not incorrect, these are parsed by ecmarkup, right?

spec/datetimeformat.html
leobalter: LGTM, @spectranaut ?

spectranaut: Looks good to me -- I'll use this structure, but I'm confused what table you mean to be referencing, @sffc ? There is no table with just the entries of this record we are constructing that I can find..

sffc: I'm suggesting that you make a table similar to https://github.com/tc39/proposal-intl-displaynames/pull/44/files and https://github.com/tc39/proposal-unified-intl-numberformat/pull/58.  Something like,

| Field | Algorithm for calendar == `"gregory"` |
|---|---|
| `[[weekday]]` | The result of `WeekDay(tz)` specified in ES2020's <emu-xref href="#sec-week-day">Week Day</emu-xref>. |

and so forth.

spectranaut: nice. will do!

spec/datetimeformat.html
```diff
@@ -364,17 +364,70 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+              <th>Value Calculation for Gregorian Calendar</th>
+            </tr>
+          </thead>
+          <tr>
+            <td>[[weekday]]</td>
+            <td>`WeekDay(tz)` specified in ES2020's <emu-xref href="#sec-week-day">Week Day</emu-xref></td>
+          </tr>
+          <tr>
+            <td>[[era]]</td>
+            <td></td>
```
sffc: Era?

spec/datetimeformat.html
```diff
@@ -364,17 +364,70 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+          </tr>
+          <tr>
+            <td>[[year]]</td>
+            <td>`YearFromTime(tz)` specified in ES2020's <emu-xref href="#sec-year-number">Year Number</emu-xref></td>
```
sffc: I didn't check that these section links are valid.  I'll trust that they are valid.

spec/datetimeformat.html
spectranaut: I actually am a bit confused about what to put for era here and wanted to reach out, @sffc and @leobalter -- "era" is not discussed in 262, and [PartitionDateTimePattern](https://tc39.es/ecma402/#sec-partitiondatetimepattern) does not seem to expect a particular format. Should we just choose "AD" and "BC" and return based on the year? :/

sffc: Yes, I would like Gregorian calculation to be very well defined.  FYI, the CLDR data that ICU uses to determine era is here:

https://github.com/unicode-org/cldr/blob/release-35-1/common/supplemental/supplementalData.xml#L4289

I think it is very reasonable to write that out here.

PartitionDateTimePattern is also really hand-wavy and needs some work.  @Ms2ger is working on that a bit in tc39/proposal-temporal#140.

spec/datetimeformat.html
```diff
@@ -364,17 +364,70 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
       </p>
 
       <emu-alg>
-        1. Apply calendrical calculations on _date_ for the given _calendar_ and _timeZone_ to produce weekday, era, year, month, day, hour, minute, second, and inDST values. The calculations should use best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules. If the _calendar_ is `"gregory"`, then the calculations must match the algorithms specified in ES2020, <emu-xref href="#sec-overview-of-date-objects-and-definitions-of-abstract-operations"></emu-xref>.
-        1. Return a Record with fields [[weekday]], [[era]], [[year]], [[month]], [[day]], [[hour]], [[minute]], [[second]], and [[inDST]], each with the corresponding calculated value.
+        1. Assert: Type(_t_) is Number.
+        1. If _calendar_ is `"gregory"`,
+          1. Let _timeZoneOffset_ be the value calculated according to <emu-xref href="#sec-local-time-zone-adjustment">LocalTZA(_t_, *true*)</emu-xref> where the local time zone is replaced with timezone _timeZone_.
```
Ms2ger: "where the local time zone is replaced with timezone _timeZone_" feels really strange to me. Can we update 262 to split LocalTZA into two parts?

spectranaut: I thought about this -- but was unsure if it was worth addressing the vagueness of "local time" and "local time zone" references in 262's implementation defined date time algorithms.

Here is an example implementation to consider: https://github.com/tc39/ecma262/compare/master...bocoup:timezoneadjustment?expand=1#diff-3540caefa502006d8a33cb1385720803

The link above adds an abstract operation called `TimeZoneAdjustment`, which is `LocalTZA` with the additional third argument `timeZone`. `LocalTZA` now calls `TimeZoneAdjustment` after setting `timeZone` to the mysterious `local time zone`.

@ljharb, do you think it's reasonable to break up ecma262's `LocalTZA` for this specific use case in ecma402? Specifically, so we can call `TimeZoneAdjustment(_t_, *true*, timeZone)` in the line @Ms2ger commented on above.

spec/datetimeformat.html
sffc: This string isn't exposed to the user, so I don't think it matters if you use AD/BC or 0/1 (or, for that matter, CE/BCE).

spec/datetimeformat.html
```diff
@@ -364,17 +364,70 @@ <h1>FormatDateTimeToParts ( _dateTimeFormat_, _x_ )</h1>
...
+        <table class="real-table">
+          <thead>
+            <tr>
+              <th>Field</th>
+              <th>Value Calculation for Gregorian Calendar</th>
+            </tr>
+          </thead>
+          <tr>
+            <td>[[weekday]]</td>
+            <td>`WeekDay(tz)` specified in ES2020's <emu-xref href="#sec-week-day">Week Day</emu-xref></td>
```
gibson042: I think we should just reference "ECMAScript" rather than a specific edition.

spec/datetimeformat.html
gibson042: I think we should just reference "ECMAScript" rather than a specific edition, and link directly to the operation rather than the containing section (e.g., "#sec-year-number" → "#eqn-YearFromTime").

spec/datetimeformat.html
spectranaut: We talked about it in an ecma402 meeting, going with the prose as is.

spec/datetimeformat.html
spectranaut: I just made these changes and built the HTML, and got the error: `Warning: can't find clause, production, note or example with id #eqn-YearFromTime` (similarly for the other operations)

I can try to dig into this later -- but does anyone here know the rules for linking cross specs for `eqn` and why `sec` references work? :/

spec/datetimeformat.html
ljharb: I think it’d be fine to split up the abstract ops in 262, if that’d help here, for future reference.

PR #349, opened 2019-05-27
Normative: Allow calendar to determine choice of pattern

littledan:
In CLDR, the set of patterns is based on the the calendar, not
simply the language-region-script. Previously, however, the set
was based purely on the base name. This patch switches to a two-level
lookup to permit implementations to have more accuracy.

Splitting off part of #227. Fixes part of #225.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -87,7 +87,8 @@
         1. Let _localeData_ be %DateTimeFormat%.[[LocaleData]].
         1. Let _r_ be ResolveLocale( %DateTimeFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %DateTimeFormat%.[[RelevantExtensionKeys]], _localeData_).
         1. Set _dateTimeFormat_.[[Locale]] to _r_.[[locale]].
-        1. Set _dateTimeFormat_.[[Calendar]] to _r_.[[ca]].
+        1. Let _calendar_ be _r_.[[ca]].
+        1. Set _dateTimeFormat_.[[Calendar]] to _calendar_.
         1. Set _dateTimeFormat_.[[HourCycle]] to _r_.[[hc]].
         1. Set _dateTimeFormat_.[[NumberingSystem]] to _r_.[[nu]].
         1. Let _dataLocale_ be _r_.[[dataLocale]].
@@ -106,7 +107,7 @@
           1. Let _value_ be ? GetOption(_options_, _prop_, *"string"*, &laquo; the strings given in the Values column of the row &raquo;, *undefined*).
           1. Set _opt_.[[&lt;_prop_&gt;]] to _value_.
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
-        1. Let _formats_ be _dataLocaleData_.[[formats]].
+        1. Let _formats_ be _dataLocaleData_.[[formats]].[[&lt;_calendar_&gt;]].
         1. Let _matcher_ be ? GetOption(_options_, *"formatMatcher"*, *"string"*, &laquo; *"basic"*, *"best fit"* &raquo;, *"best fit"*).
         1. If _matcher_ is *"basic"*, then
           1. Let _bestFormat_ be BasicFormatMatcher(_opt_, _formats_).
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@anba @zbraniecki  @caridy

littledan:
I agree. We have also discussed it in TC39 and the ECMA-402 meeting in the past, so I don't think there is any formal need to request further consensus.

FrankYFTang:
what do we need to do to merge this PR? I would like to launch my implementation of https://github.com/tc39/ecma402/pull/175 in v8 soon but I believe I need to adjust the code align with this PR first before launching that. This PR should be consider as a bug fix, right?
@sffc @leobalter

FrankYFTang:
hum... it looks like this PR is intertwine with https://github.com/tc39/ecma402/pull/351
because once we select the pattern based on the calendar, patterns for calendar other than non-gregorian may have some fields we need to defined in https://github.com/tc39/ecma402/pull/351 . So this PR cannot be merged without adding those field names (in the formatToParts) in these non-gregorian patterns.

leobalter:
> I agree. We have also discussed it in TC39 and the ECMA-402 meeting in the past, so I don't think there is any formal need to request further consensus.

If possible can you help finding a reference to link from this PR? Just for clarity.

leobalter:
@FrankYFTang I believe it's fine to add this PR before #351, but I'm fine to wait if you prefer. 

@littledan maybe you'd like to bring #351 to the TC39 meeting considering it complements unblocks this?

in the meantime, @FrankYFTang we need to get tests for #351 as well.

FrankYFTang:
The reason this PR is depending on #351 is because of the following. 
Once we choose the pattern based on the calendar, then the pattern of non Gregorian calendar may contain relatedYear and yearName, therefore, the formatToParts will need to output these parts, but unless we also merge #351, we cannot output those parts. So this PR is blocked by #351 from my point of view.

littledan:
I think it's only intertwined in implementations that have a locale database that includes those patterns. IMO we can land this patch separately, and the spec will continue to imply that those patterns can't be used yet.

FrankYFTang:
> I think it's only intertwined in implementations that have a locale database that includes those patterns. IMO we can land this patch separately, and the spec will continue to imply that those patterns can't be used yet.

agree. It should be a product launching dependency, not specification dependency. I think it is LGTM in this shape.

leobalter:
@littledan would you mind rebasing this one as well? I believe it's best we have your input here rather than I picking a preference without coordination.

sffc:
@srl295 You commented in the September 26 meeting that it may be possible to write a test for this PR using the Hebrew calendar.  Were you able to follow up with this?

srl295:
OK, i've just been around and around on this because of v8's "harmony_intl_other_calendars" flag. What is this flag and why is it off by default? @FrankYFTang

srl295:
> @srl295 You commented in the September 26 meeting that it may be possible to write a test for this PR using the Hebrew calendar. Were you able to follow up with this?

So the original issue, now that I have found the right v8 flag, has a good test case. Examples shown below:

```
> new Date(0).toLocaleString("en-u-ca-chinese", {year: "numeric"})
'1969(ji-you)'
> new Date(0).toLocaleString("en-u-ca-gregory", {year: "numeric"})
'1969'
> new Date(0).toLocaleString("en-u-ca-japanese", {year: "numeric"})
'44 Shōwa'
```

The Chinese and Japanese calendars best match changes the pattern. You can use the ICU web demo to see this:  (input pattern `y`, 3 different calendars).

- http://demo.icu-project.org/icu4jweb/flexTest.jsp?pat=y&_=en%40calendar%3Dchinese
- http://demo.icu-project.org/icu4jweb/flexTest.jsp?pat=y&_=en%40calendar%3Dgregorian
- http://demo.icu-project.org/icu4jweb/flexTest.jsp?pat=y&_=en%40calendar%3Djapanese

littledan:
Now that this is rebased, and given the tests that @FrankYFTang wrote at https://github.com/tc39/test262/pull/2379 , is this PR good to land?

leobalter:
@littledan please help figuring out the missing parts here? https://github.com/tc39/ecma402/wiki/Proposal-and-PR-Progress-Tracking#ecma-402-prs

Do we have a clear signal from SM?

I can merge to make it follow the same train as #351. Pinging @sffc for feedback

sffc:
From last week ECMA-402:

> SFC: Do we need a SpiderMonkey implementation for this one?
>
> JSW: Off the top of my head I don't think we do (choose locales in this way), but I'm not sure.

@jswalden can you elaborate?  Do we need a SM implementation for this PR?  (Also #351)

anba:
#351 and #349 have been implemented in <https://bugzilla.mozilla.org/show_bug.cgi?id=1573439>, but are currently restricted to Firefox Nightly. We're currently in [our soft freeze period](https://groups.google.com/forum/?_escaped_fragment_=topic/mozilla.dev.platform/f4171vFyaHg), so we can't remove the Nightly-only restriction before next week. But when removed after this week, Firefox should be able to provide both changes for the [Firefox 73 release](https://wiki.mozilla.org/RapidRelease/Calendar).

jswalden:
FWIW we removed the nightly-only restriction on #351 and #349 a couple days ago for SpiderMonkey.

sffc:
> FWIW we removed the nightly-only restriction on #351 and #349 a couple days ago for SpiderMonkey.

Great!  Can you update the status wiki as appropriate?

jswalden:
> Can you update the status wiki as appropriate?

Done.

littledan:
Rebased again; I believe this should be ready to land.

PR #350, opened 2019-05-27
Editorial: Consistently use records  with [[]]

littledan:
The patch makes an editorial change to make the descriptions
of the locale database fully switch to records (there was some object-style
notation remaining).

Split off from #227


leobalter:
I like this, thanks!

PR #351, opened 2019-05-27
Normative: Permit relatedYear and yearName in output

littledan:
These fields are present in, e.g., some Chinese calendars.

Broken out from #227. Fixes part of #225.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -296,6 +296,13 @@
             1. Else,
               1. Let _fv_ be an implementation and locale dependent String value representing `"ante meridiem"`.
             1. Add new part record { [[Type]]: `"dayPeriod"`, [[Value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal to `"relatedYear"`, then
+            1. Let _v_ be _tm_.[[relatedYear]].
+            1. Let _fv_ be FormatNumber(_nf_, _v_).
+            1. Add new part record { [[Type]]: `"relatedYear"`, [[Value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal to `"yearName"`, then
+            1. Let _v_ be _tm_.[[yearName]].
+            1. Add new part record { [[Type]]: `"yearName"`, [[Value]]: _v_ } as a new element of the list _result_.
           1. Else,
             1. Let _unknown_ be an implementation-, locale-, and numbering system-dependent String based on _x_ and _p_.
             1. Append a new Record { [[Type]]: `"unknown"`, [[Value]]: _unknown_ } as the last element of _result_.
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@anba @zbraniecki @caridy

littledan:
Thanks for the review!

spec/datetimeformat.html
```diff
@@ -303,6 +303,13 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
             1. Else,
               1. Let _fv_ be an implementation and locale dependent String value representing `"ante meridiem"`.
             1. Add new part record { [[Type]]: `"dayPeriod"`, [[Value]]: _fv_ } as a new element of the list _result_.
+          1. Else if _p_ is equal to `"relatedYear"`, then
+            1. Let _v_ be _tm_.[[relatedYear]].
+            1. Let _fv_ be FormatNumber(_nf_, _v_).
+            1. Add new part record { [[Type]]: `"relatedYear"`, [[Value]]: _fv_ } as a new element of the list _result_.
```
FrankYFTang: if we are using kebab-case, should this be "related-year" instead?

spec/datetimeformat.html
```diff
@@ -303,6 +303,13 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
...
+          1. Else if _p_ is equal to `"yearName"`, then
```
FrankYFTang: same as above, should this be "year-name" instead?

spec/datetimeformat.html
leobalter: Those are option keys (property names) but not the enumeration values. So camelCase is fine here. I had the same question and reviewed this part yesterday.

spec/datetimeformat.html
leobalter: same answer

spec/datetimeformat.html
leobalter: the key to quickly differentiate here what is a property name or a value, is distinguishing `_p_` (property name) and `_f_` (values)

littledan:
Thanks for your reviews. Seems like no changes are needed at the moment, right? Are we missing anything besides test262 tests (cc @Ms2ger) and TC39 review to land this PR?

leobalter:
Resolving the conflicts manually before merging. We have the tests landed.

leobalter:
Actually I'd like to defer this to @littledan so you can pick the best description for relatedYear and yearName in their [respective table](https://tc39.es/ecma402/#table-datetimeformat-tolocaltime-record).

I pushed a separate branch with a rebased commit if you wanna reuse it, but not required to.

https://github.com/tc39/ecma402/compare/master...leobalter:relatedyear-yearname?expand=1#diff-1f2e1333fe08618e8af2d7344b031dc1R400-R405

leobalter:
I believe we can mention ICU for the necessary references in that table, e.g. https://github.com/tc39/ecma402/pull/227#issuecomment-488109525

sffc:
The status wiki for this PR indicates that there is still work to be done on Tests, MDN, SM, and V8.  Is that still the case?  It could be that the wiki page is out of date.

https://github.com/tc39/ecma402/wiki/Proposal-and-PR-Progress-Tracking

leobalter:
My understanding is that we already have tests. I need to double check the other parts.

leobalter:
So this is not passing on SM unflagged. In this case, I can revert the PR, @jswalden do you have update to avoid a revert?

sffc:
I don't think we need to revert the PR as long as the outstanding issues get taken care of in a timely fashion (by the end of Q4).

PR #352, opened 2019-05-28
Normative: Align currency handling in NumberFormat constructor with the Unified API proposal.

Ms2ger:
This contains a few minor but script-observable changes to the NumberFormat constructor from the Unified API proposal.


sffc:
Related: #307

There is a comment there:

> During the 2019-1-17 ECMA402 meeting, we discussed and decided this kind of minor tweaks won't get into ECMA 402 except through the Unified Number Format proposal Shane is working on. And this issue is already there. Close this issue therefore.

sffc:
That being said, I'm not personally opposed to PRs like this *if they help someone*.  It's just that I don't think we should go out of our way and spend a lot of time pulling apart the Unified proposal.

leobalter:
As a normative change: should we discuss it in this TC39 meeting or should we wait for the next ecma402 meeting?

leobalter:
As discussed in the Intl meeting from June 13th 2019, we are closing the PR in favor of the unified numberformat.

PR #354, opened 2019-05-30
Editorial: Update editors lists with emails

leobalter:


spec/metablock.html
```diff
@@ -7,9 +7,11 @@ <h1>Contributing to this Specification</h1>
     <li>Issues: <a href="https://github.com/tc39/ecma402/issues">All Issues</a>, <a href="https://github.com/tc39/ecma402/issues/new">File a New Issue</a></li>
     <li>Pull Requests: <a href="https://github.com/tc39/ecma402/pulls">All Pull Requests</a>, <a href="https://github.com/tc39/ecma402/pulls/new">Create a New Pull Request</a></li>
     <li>Test Suite: <a href="https://github.com/tc39/test262">Test262</a></li>
-    <li>Editor:
+    <li>Editors:
       <ul>
-        <li><a href="mailto:leonardo.balter at gmail dot com">Leo Balter</a> (<a href="https://twitter.com/leobalter">@leobalter</a>)</li>
+        <li><a href="mailto:ecma402+leo@bocoup.com">Leo Balter</a> (<a href="https://twitter.com/leobalter">@leobalter</a>)</li>
+        <li><a href="mailto:ecma402+val@bocoup.com">Valerie Young</a> (<a href="https://twitter.com/spectranaut">@spectranaut</a>)</li>
+        <li><a href="mailto:ecma402+isaac@bocoup.com">Isaac Durazo</a> (<a href="https://twitter.com/isaacdurazo">@isaacdurazo</a>)</li>
```
leobalter: @sffc I created this email to be fully open and publicly available for immediate response from the editors.

I believe we should add a new part here listing you and Daniel respectively as TC39-TG2's convener and chief advisor. To make it possible I need to confirm with you what email address we can publicly use.

I also wonder if we should say at some point about the TC39-TG2 in this document.

sffc: Thanks Leo!  My "public" email is shane@unicode.org.

Maybe also include ecma402-admin@chromium.org?

leobalter:
@sffc sorry for the delay, it's updated now. WDYT?

PR #359, opened 2019-06-20
Normative: Apply kebab-case for best-fit

leobalter:
This is an initial effort to apply kebab-case for enumeration string values in ECMA-402.

This style follow convention set at the W3C Tag Design Principles. Ref:
https://w3ctag.github.io/design-principles/#casing-rules

The goal is to illustrate a change in an experimental way, it first needs discussion from
the Intl team to confirm if 'best fit' should also be included in a possible set of changes
to match style convention.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -27,7 +27,8 @@
         1. Else,
           1. Let _localeData_ be %Collator%.[[SearchLocaleData]].
         1. Let _opt_ be a new Record.
-        1. Let _matcher_ be ? GetOption(_options_, `"localeMatcher"`, `"string"`, &laquo; `"lookup"`, `"best fit"` &raquo;, `"best fit"`).
+        1. Let _matcher_ be ? GetOption(_options_, `"localeMatcher"`, `"string"`, &laquo; `"lookup"`, `"best-fit"`, `"best fit"` &raquo;, `"best-fit"`).
+        1. If _matcher_ is `"best fit"`, let _matcher_ be `"best-fit"`.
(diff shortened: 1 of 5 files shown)


leobalter:
some considerations: I'm not sure we should change "best fit", it's a thing of its own sort like timezone names. I'm happy to close this PR if necessary.

leobalter:
Also, this reference: https://github.com/tc39/proposal-intl-displaynames/issues/32

cc @annevk @sffc

littledan:
I don't think it makes sense to go back and apply these kinds of conventions to existing APIs. It might not be web-compatible. Let's just look into making good use of conventions when it makes sense going forward.

leobalter:
> It might not be web-compatible.

well, the work here is just extending and not removing any old parts. I understand if we don't find enough traction to ship this PR's specific change but I don't believe it might reveal as non-web-compatible.

I think we should hold the discussion here anyway and continue the casing discussion for the new apis. For here, let's wait until a new meeting call?

littledan:
I'm not going to be in the calls for the next couple months. You can have as my input to that discussion: As a general principle, I don't think we should go and change the names of things that are already shipped to the web.

leobalter:
I think it’s reasonable to say we won’t have consensus for this specific PR, then. It’s not necessary to keep it open by the same reason and I don’t want to force a discussion where you’re not present and voiced a valid objection.

For the record, we should move the convo back to DisplayNames and I’d like to know if there are parts there you consider as already shipping. Would you help me pointing to these cases?

PR #360, opened 2019-07-04
Editorial: More consistent wording for the %DateTimeFormat%.[[LocaleData]] requirements

Ms2ger:


sffc:
Old wording:

> [[LocaleData]].[[&lt;_locale_&gt;]] must have an [[hourCycle]] field with a String value equal to `"h11"`, `"h12"`, `"h23"`, or `"h24"` for all locale values _locale_.

New wording:

> [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle]] must be a String value equal to `"h11"`, `"h12"`, `"h23"`, or `"h24"`.

That is, the old wording is "X must have Y field which is Z", whereas the new wording is "X[Y] must be Z".  Does the new wording imply that X[Y] exists?  Are there examples elsewhere in ECMA 262 or 402?

leobalter:
@sffc It seems we've got the same sentiment. AFAIK, both specs are pretty explicit when a field is required.

We can ping @ljharb and @zenparsing to weight in here. hHopefully, it may not take a long time from them.

ljharb:
The dot notation implicitly yields undefined, i believe, if the LHS of the dot doesn’t have the field.

spec/datetimeformat.html
```diff
@@ -491,21 +491,21 @@ <h1>Internal slots</h1>
...
 
       <p>
-        The value of the [[LocaleData]] internal slot is implementation defined within the constraints described in <emu-xref href="#sec-internal-slots"></emu-xref> and the following additional constraints:
+        The value of the [[LocaleData]] internal slot is implementation defined within the constraints described in <emu-xref href="#sec-internal-slots"></emu-xref> and the following additional constraints, for all locale values _locale_:
       </p>
 
       <ul>
         <li>
-          The list that is the value of the `"nu"` field of any locale field of [[LocaleData]] must not include the values `"native"`, `"traditio"`, or `"finance"`.
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[nu]] must be a list that does not include the values `"native"`, `"traditio"`, or `"finance"`.
```
ljharb:
```suggestion
          [[LocaleData]].[[&lt;_locale_&gt;]].[[nu]] must be a List that does not include the values `"native"`, `"traditio"`, or `"finance"`.
```
should this really be “traditio” or is it meant to be “traditional” or similar? Right now it looks like a typo/truncation.

spec/datetimeformat.html
```diff
@@ -491,21 +491,21 @@ <h1>Internal slots</h1>
...
         </li>
         <li>
-          [[LocaleData]][locale] must have a formats field for all locale values. The value of this field must be a list of records, each of which has a subset of the fields shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, where each field must have one of the values specified for the field in <emu-xref href="#table-datetimeformat-components"></emu-xref>. Multiple records in a list may use the same subset of the fields as long as they have different values for the fields. The following subsets must be available for each locale:
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[formats]] must be a list of records, each of which has a subset of the fields shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, where each field must have one of the values specified for the field in <emu-xref href="#table-datetimeformat-components"></emu-xref>. Multiple records in a list may use the same subset of the fields as long as they have different values for the fields. The following subsets must be available for each locale:
```
ljharb:
```suggestion
          [[LocaleData]].[[&lt;_locale_&gt;]].[[formats]] must be a List of records, each of which has a subset of the fields shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, where each field must have one of the values specified for the field in <emu-xref href="#table-datetimeformat-components"></emu-xref>. Multiple records in a List may use the same subset of the fields as long as they have different values for the fields. The following subsets must be available for each locale:
```

leobalter:
ping @Ms2ger

Ms2ger:
Fixed the List capitalization. I think it's clear that "X[Y] must be Z" implies that X[Y] exists. Please let me know if you have any other questions.

PR #364, opened 2019-07-18
Add notes for 2019-06 and 2019-07

leobalter:


sffc:
Thanks, this is my responsibility but thanks for helping out :)

There are some things that should be cleaned up in the notes doc before publishing:

1. There are some lists in the Google Doc that Google converted to "fancy" formatted lists, and when you copy them into Markdown, the bullets go away.  For example, most of the header sections in the proposal updates have "-" lists.  The auto-format should be removed such that when you copy the Google Doc, you get valid Markdown.  There may also be links worth fixing; for example, links like "issues/32" don't get copied over correctly to Markdown.  In the future we might want to disable any kind of fancy formatting in Google Docs, such that the Markdown process is easier.
2. There was one TODO in the July notes doc, about the HTMLElement.style property.  I filled it in on the Google Doc.  Please replace it here.

sffc:
I addressed my feedback and pushed to your branch.  I'll merge the PR.

PR #366, opened 2019-07-29
Editorial: Extract pattern string handling into an abstract method

Ms2ger:
The PartitionNumberPattern and PartitionDateTimePattern algorithms contain
substantively similar prose to iterate over a string of the form
"aaa{property}bbb". Extracting this from from the domain-specific prose
makes all three algorithms more readable.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -266,19 +266,12 @@
         1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, `"useGrouping"`, *false*).
         1. Let _nf2_ be ? Construct(%NumberFormat%, &laquo; _locale_, _nf2Options_ &raquo;).
         1. Let _tm_ be ToLocalTime(_x_, _dateTimeFormat_.[[Calendar]], _dateTimeFormat_.[[TimeZone]]).
-        1. Let _pattern_ be _dateTimeFormat_.[[Pattern]].
         1. Let _result_ be a new empty List.
-        1. Let _beginIndex_ be Call(%StringProto_indexOf%, _pattern_, &laquo; `"{"`, 0 &raquo;).
-        1. Let _endIndex_ be 0.
-        1. Let _nextIndex_ be 0.
-        1. Let _length_ be the number of code units in _pattern_.
-        1. Repeat, while _beginIndex_ is an integer index into _pattern_
-          1. Set _endIndex_ to Call(%StringProto_indexOf%, _pattern_, &laquo; `"}"`, _beginIndex_ &raquo;).
-          1. Assert: _endIndex_ is greater than _beginIndex_.
-          1. If _beginIndex_ is greater than _nextIndex_, then
-            1. Let _literal_ be a substring of _pattern_ from position _nextIndex_, inclusive, to position _beginIndex_, exclusive.
-            1. Add new part record { [[Type]]: `"literal"`, [[Value]]: _literal_ } as a new element of the list _result_.
-          1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
+        1. Let _patternParts_ be PartitionPattern(_dateTimeFormat_.[[Pattern]]).
+        1. For each element _patternPart_ of _patternParts_, in List order, do
+          1. Let _p_ be _patternPart_.[[Type]].
+          1. If _p_ is `"literal"`, then
+            1. Add new part record { [[Type]]: `"literal"`, [[Value]]: _patternPart_.[[Value]] } as a new element of the list _result_.
           1. If _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then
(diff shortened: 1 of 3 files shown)


spec/negotiation.html
```diff
@@ -325,5 +325,35 @@ <h1>GetNumberOption ( _options_, _property_, _minimum_, _maximum_, _fallback_ )<
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-partitionpattern" aoid="PartitionPattern">
+      <h1>PartitionPattern ( _pattern_ )</h1>
+
+      <p>
+        The PartitionPattern abstract operation is called with argument _pattern_. The following steps are taken:
```
Ms2ger: I could add some more details about the return value if it would be helpful.

sffc: This is a good opportunity to say something like,

> This abstract operation parses an abstract pattern string into a list of Records with two fields, `[[Type]]` and `[[Value]]`.  The syntax of the abstract pattern strings is an implementation detail and is not exposed to users of ECMA-402.

spec/negotiation.html
```diff
@@ -325,5 +325,39 @@ <h1>GetNumberOption ( _options_, _property_, _minimum_, _maximum_, _fallback_ )<
...
+        1. Repeat, while _beginIndex_ is an integer index into _pattern_
+          1. Set _endIndex_ to Call(%StringProto_indexOf%, _pattern_, &laquo; `"}"`, _beginIndex_ &raquo;).
+          1. Assert: _endIndex_ is greater than _beginIndex_.
+          1. If _beginIndex_ is greater than _nextIndex_, then
+            1. Let _literal_ be a substring of _pattern_ from position _nextIndex_, inclusive, to position _beginIndex_, exclusive.
+            1. Add new part record { [[Type]]: `"literal"`, [[Value]]: _literal_ } as a new element of the list _result_.
+          1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
+          1. Add new part record { [[Type]]: _p_, [[Value]]: *undefined* } as a new element of the list _result_.
+          1. Set _nextIndex_ to _endIndex_ + 1.
+          1. Set _beginIndex_ to Call(%StringProto_indexOf%, _pattern_, &laquo; `"{"`, _nextIndex_ &raquo;).
```
leobalter: same here

spec/negotiation.html
```diff
@@ -325,5 +325,35 @@ <h1>GetNumberOption ( _options_, _property_, _minimum_, _maximum_, _fallback_ )<
...
+    <emu-clause id="sec-partitionpattern" aoid="PartitionPattern">
+      <h1>PartitionPattern ( _pattern_ )</h1>
+
+      <p>
+        The PartitionPattern abstract operation is called with argument _pattern_. The following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Let _result_ be a new empty List.
+        1. Let _beginIndex_ be Call(%StringProto_indexOf%, _pattern_, &laquo; `"{"`, 0 &raquo;).
```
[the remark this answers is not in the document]
ljharb: This can now use the new intrinsics notation.
```suggestion
        1. Let _beginIndex_ be Call(%String.prototype.indexOf%, _pattern_, &laquo; `"{"`, 0 &raquo;).
```

spec/negotiation.html
ljharb:
```suggestion
          1. Set _beginIndex_ to Call(%String.prototype.indexOf%, _pattern_, &laquo; `"{"`, _nextIndex_ &raquo;).
```

spec/datetimeformat.html
```diff
@@ -266,19 +266,12 @@ <h1>PartitionDateTimePattern ( _dateTimeFormat_, _x_ )</h1>
...
-          1. Assert: _endIndex_ is greater than _beginIndex_.
-          1. If _beginIndex_ is greater than _nextIndex_, then
-            1. Let _literal_ be a substring of _pattern_ from position _nextIndex_, inclusive, to position _beginIndex_, exclusive.
-            1. Add new part record { [[Type]]: `"literal"`, [[Value]]: _literal_ } as a new element of the list _result_.
-          1. Let _p_ be the substring of _pattern_ from position _beginIndex_, exclusive, to position _endIndex_, exclusive.
+        1. Let _patternParts_ be PartitionPattern(_dateTimeFormat_.[[Pattern]]).
+        1. For each element _patternPart_ of _patternParts_, in List order, do
+          1. Let _p_ be _patternPart_.[[Type]].
+          1. If _p_ is `"literal"`, then
+            1. Add new part record { [[Type]]: `"literal"`, [[Value]]: _patternPart_.[[Value]] } as a new element of the list _result_.
```
ljharb: is “new part record” a thing? id expect `append new Record { ... } to list _result_`

Ms2ger: It was a thing before this change, so I'm not planning to change that.

Ms2ger:
Any more comments?

Issue #363, opened 2019-07-15
Add casing recommendations to style guide


sffc:
We agreed in https://github.com/tc39/proposal-intl-displaynames/issues/32 to recommend camel case for both keys and enum strings.  We should document this decision in a style guide somewhere so that this discussion doesn't get lost in time.

@leobalter @spectranaut

spectranaut:
this has been added by @sffc : https://github.com/tc39/ecma402/blob/master/docs/style-guide.md

PR #377, opened 2019-09-21
Editorial: Adding style-guide.md with casing conventions explainer.

sffc:
Follow-up to https://github.com/tc39/proposal-intl-displaynames/issues/32.

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -165,12 +165,12 @@ ECMA-402 could allow users to choose between kebab case and camel case for relev
 
 #### Examples
 
-Locale identifiers are defined in [UTS #35 Section 3](https://www.unicode.org/reports/tr35/#Identifiers).  Examples:
+[BCP47](http://www.rfc-editor.org/rfc/bcp/bcp47.txt) identifiers, using registered extensions as described in [UTS#35§3](https://www.unicode.org/reports/tr35/#Identifiers).  Examples:
 
 - `"en-US"`
 - `"ar-EG-u-nu-latn"`
 
-Unit identifiers are defined in [UTS #35 Part 2 Section 6](https://www.unicode.org/reports/tr35/tr35-general.html#Unit_Elements).  Examples:
+Unit identifiers, defined in [UTS #35 Part 2 Section 6](https://www.unicode.org/reports/tr35/tr35-general.html#Unit_Elements).  Examples:
 
 - `"square-kilometer"`
 - `"part-per-million"`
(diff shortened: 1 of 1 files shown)


sffc:
@waldemarhorwat @wycats @domenic @gesa @littledan @leobalter 

I would like to adopt this style guide for ECMA-402.  It recommends Option 3 from my presentation in July, to use camel case everywhere.  As @littledan suggested, these guidelines would be scoped only to ECMA-402 and would not restrict what ECMA-262 could do in the future.

I am adding this to the October agenda in https://github.com/tc39/agendas/pull/622.

sffc:
Backref: Fixes #363

docs/style-guide.md
```diff
@@ -0,0 +1,201 @@
...
+            - [Decision](#decision-2)
+    - [Identifiers defined outside ECMA-402](#identifiers-defined-outside-ecma-402)
+        - [Examples](#examples-1)
+        - [Alternative: Convert identifiers to camel case](#alternative-convert-identifiers-to-camel-case)
+            - [Pros](#pros-3)
+            - [Cons](#cons-3)
+            - [Decision](#decision-3)
+
+
+*Table of Contents generated using https://magnetikonline.github.io/markdown-toc-generate/*
```
leobalter: I'm fine adding this, but I wonder if we should move it to the end of the file.

docs/style-guide.md
```diff
@@ -0,0 +1,201 @@
...
+- It is difficult to take this option seriously since it would add debt to the specification.
+- If strongly desired, the spec could be changed later to allow the kebab-cased strings.
+
+### Identifiers defined outside ECMA-402
+
+:star2: *ECMA-402 sits between ECMA-262 and internationalization standards such as UTS #35. For identifiers whose syntax comes from a different specification, follow the convention in that specification.* :star2:
+
+#### Examples
+
+Locale identifiers are defined in [UTS #35 Section 3](https://www.unicode.org/reports/tr35/#Identifiers).  Examples:
```
srl295: I'd prefer something such as 
> [BCP47](http://www.rfc-editor.org/rfc/bcp/bcp47.txt) identifiers, using registered extensions as described in [UTS#35§3](https://www.unicode.org/reports/tr35/#Identifiers)

The identifiers are and must be valid BCP47.

leobalter:
TC39 approved this document during the meeting in 2019-10-02

sffc:
@waldemarhorwat had some suggestions from TC39; could you leave comments here?  Thanks!

leobalter:
ping @waldemarhorwat

docs/style-guide.md
sffc: There isn't really anywhere to put it down there.  I like having it close to the actual ToC so it's easy to find how to update it.

docs/style-guide.md
sffc: Done.

sffc:
I'm going to merge this.  Feedback from @waldemarhorwat can be integrated in a follow-up PR.

PR #385, opened 2019-11-17
Editorial: Remove reference to undeclared variable in PluralRules constructor

gibson042:


spec/pluralrules.html
```diff
@@ -156,7 +156,7 @@ <h1>Intl.PluralRules ( [ _locales_ [ , _options_ ] ] )</h1>
 
       <emu-alg>
         1. If NewTarget is *undefined*, throw a *TypeError* exception.
-        1. Let _pluralRules_ be ? OrdinaryCreateFromConstructor(_newTarget_, "%PluralRulesPrototype%", &laquo; [[InitializedPluralRules]], [[Locale]], [[Type]], [[MinimumIntegerDigits]], [[MinimumFractionDigits]], [[MaximumFractionDigits]], [[MinimumSignificantDigits]], [[MaximumSignificantDigits]] &raquo;).
+        1. Let _pluralRules_ be ? OrdinaryCreateFromConstructor(NewTarget, `"%PluralRulesPrototype%"`, &laquo; [[InitializedPluralRules]], [[Locale]], [[Type]], [[MinimumIntegerDigits]], [[MinimumFractionDigits]], [[MaximumFractionDigits]], [[MinimumSignificantDigits]], [[MaximumSignificantDigits]] &raquo;).
```
leobalter: This is correct, but out of curiosity:

In ECMA-262 there are usages of both `NewTarget` and `|NewTarget|`. The approach here is similar to passing it as a parameter but I wonder if that's something that should be normalized (first in ECMA-262). 

cc @ljharb

ljharb: Where is `|NewTarget|` used in 262?

PR #386, opened 2019-11-18
Editorial: Update spec text to be more consistent across interfaces

gibson042:


anba:
I think ECMA-262 normally uses "*Thing* instances are ordinary objects that inherit properties from ...", so this sentence should be left as-is:

> Intl.Collator instances are ordinary objects that inherit properties from %CollatorPrototype%.

And instead the remaining sections should be updated to include "are ordinary objects".

leobalter:
That's accurate and I skipped this one, fixing it in a moment

PR #387, opened 2019-11-20
Editorial: Reference ECMA-262 RequireInternalSlot and string-concatenation

gibson042:
A bunch of algorithms were duplicating [RequireInternalSlot](https://tc39.es/ecma262/#sec-requireinternalslot), but I think it makes more sense to just use it. However, I don't know if it should be copied in (as I have here) or referenced with a new biblio entry.

Also, should we do the same for [_string-concatenation_](https://tc39.es/ecma262/#sec-ecmascript-language-types-string-type)?


ljharb:
Ideally it would be referenced and not duplicated, I’d think.

gibson042:
Are there rules for which parts of ECMA-262 are safe to reference? I just don't want a target to be pulled out from under ECMA-402.

ljharb:
I'd say every part - while we don't have a programmatic way to enforce it, renaming or removal of any abstract op generally requires checking downstream specs, including 402 and HTML.

leobalter:
I'm in favor of referencing and reusing the most we can from 262. The idea of the abstractions in 262 are also to be reused by extensions.

On the other side, we need to find a better tracking for referenced abstractions. I believe we might achieve that adding tests here in 402 for the parts we add into the glossary. WDYT?

ljharb:
What would probably be most effective is adding tests in 262 that check 402 and html, that fail if any referenced ops no longer exist.

leobalter:
I believe the tests should be in ECMA-402, but adding tests in ECMA-262 for known usages is fine too.

My raw idea:

- ECMA-402: Every build will check references in the glossary, **failing** if not found in ECMA-262.
- ECMA-262: Every build will import the references from glossary files we mark (checkout repo's default branch and lookup). Instead of a failure, **warn** if new changes affect references.

This avoids a vicious cycle if we actually need to change abstraction names.

ljharb:
Sounds great.

gibson042:
Updated to reference rather than duplicate.

sffc:
In numberformat.html, you changed `.formatToParts`, but not `.format` or `.resolvedOptions`.  Is that intentional?

sffc:
Note that `.format` uses `UnwrapNumberFormat`, which checks for `[[InitializedNumberFormat]]`, whereas `.formatToParts` checks for that slot directly.

Issue #54, opened 2015-11-13
All string literals must be code formatted


rwaldron:
@bterlson why have you changed the convention for string literals?

![](https://i.gyazo.com/8c3dbe348104be74bfa9f7cacba94ff6.png)

to: 

![](https://i.gyazo.com/7a667c2287d85fb9d08c718acd7a6cf9.png)

bterlson:
Caridy just pointed me to this, sorry. It was changed as it was annoying - some quotes aren't the start of strings. In general, \*'s are used for values, so it makes sense to use them for string values as well.

spectranaut:
It looks like Ecma-262 has two conventions for string literals now.

In [String.prototype.normalize](https://tc39.es/ecma262/#sec-string.prototype.normalize) we see ``` `"string"` ```:
 ```If _f_ is not one of `"NFC"`, `"NFD"`, `"NFKC"`, or `"NFKD"`, throw a *RangeError* exception.```

Then in [String.prototype.padStart](https://tc39.es/ecma262/#sec-string.prototype.padstart) we see ``` *"string"* ```:
```Return ? StringPad(_O_, _maxLength_, _fillString_, *"start"*).```

Which is correct? Ecma-402 only uses ``` `"string"` ```. Should we change to ``` *"string"* ``` as seems to be @bterlson's preference from 3 years ago? :) @leobalter @sffc @ljharb

ljharb:
cc @rkirsling, who i believe has an open PR to unify these conventions?

rkirsling:
It's true that updating <code>\`"foo"\`</code> → <code>\*"foo"\*</code> is meant to follow https://github.com/tc39/ecma262/pull/1302, which mainly addresses the current inconsistency in <code>\`foo\`</code> vs. <code>\`"foo"\`</code> for properties, but that's one already been open for a year, so it's completely taken the wind out of my sails... :sob:

ljharb:
I plan to merge that PR within the next few weeks, for the record, just after merging a few others that need to be prioritized (and after rebasing 1302), with perhaps minimal adjustments - so I think that perhaps 402 should hold off making any changes until after that lands, at which point the overarching convention in 262 will be `*"string"*`, and then 402 could make that change?

rkirsling:
Yay! Sounds like a plan to me.

ljharb:
https://github.com/tc39/ecma262/pull/1302 has been merged.

sffc:
[...]

PR #396, opened 2019-12-19
Editorial: Stylize strings as values.

spectranaut:
This follows the convention set in ECMA-262 to format strings as literals (`*"string*"` instead of ``` `"string"` ```, see PR: https://github.com/tc39/ecma262/pull/1733) and closes https://github.com/tc39/ecma402/issues/54.

There is also one commit that adds a link in [5. Notational Conventions](https://tc39.es/ecma402/#conventions) to ECMA-262.

While reviewing, consider the following conventions from ECMA-262 that I'm not positive apply here:

- [Tilde for spec-interal values](https://github.com/tc39/ecma262/pull/1725): I couldn't find an example of a spec internal value in ECMA-402 that is not also a string specified somewhere else, such as the  IANA Time Zone Database or Unicode. Because the string is not an internal arbitrary value it doesn't seem like we should use `~`.
- [Notation for code points](https://github.com/tc39/ecma262/pull/1724/files): I don't have an opinion on whether we should stylize characters the way ECMA-262 now stylizing code points, I'm curious if others do.

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -9,11 +9,11 @@
     <h1>Case Sensitivity and Case Mapping</h1>
 
     <p>
-      The String values used to identify locales, currencies, and time zones are interpreted in a case-insensitive manner, treating the Unicode Basic Latin characters `"A"` to `"Z"` (U+0041 to U+005A) as equivalent to the corresponding Basic Latin characters `"a"` to `"z"` (U+0061 to U+007A). No other case folding equivalences are applied. When mapping to upper case, a mapping shall be used that maps characters in the range `"a"` to `"z"` (U+0061 to U+007A) to the corresponding characters in the range `"A"` to `"Z"` (U+0041 to U+005A) and maps no other characters to the latter range.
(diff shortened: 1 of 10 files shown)


spec/locales-currencies-tz.html
```diff
@@ -9,11 +9,11 @@ <h1>Identification of Locales, Currencies, and Time Zones</h1>
...
+      The String values used to identify locales, currencies, and time zones are interpreted in a case-insensitive manner, treating the Unicode Basic Latin characters *"A"* to *"Z"* (U+0041 to U+005A) as equivalent to the corresponding Basic Latin characters *"a"* to *"z"* (U+0061 to U+007A). No other case folding equivalences are applied. When mapping to upper case, a mapping shall be used that maps characters in the range *"a"* to *"z"* (U+0061 to U+007A) to the corresponding characters in the range *"A"* to *"Z"* (U+0041 to U+005A) and maps no other characters to the latter range.
     </p>
 
     <p>
-      EXAMPLES `"ß"` (U+00DF) must not match or be mapped to `"SS"` (U+0053, U+0053). `"ı"` (U+0131) must not match or be mapped to `"I"` (U+0049).
+      EXAMPLES *"ß"* (U+00DF) must not match or be mapped to *"SS"* (U+0053, U+0053). *"ı"* (U+0131) must not match or be mapped to *"I"* (U+0049).
```
rkirsling: The only concern I see is that we may wish to change cases like <code>\*"ß"\* (U+00DF)</code> to <code>\`ß\` (U+00DF)</code> throughout this file, in alignment with, say, Note 4 of [21.2.2.8.2](https://tc39.es/ecma262/#sec-runtime-semantics-canonicalize-ch) in 262.

I don't have a strong opinion on this though, particularly since we're referring to these as "characters". (@jmdyck, what do you think?)

jmdyck: It seems to me that it's incorrect for the preceding para to talk about ES String values as if they consist of Unicode code points.

leobalter: I believe this can be done through a separate patch if necessary.

PR #400, opened 2020-01-10
Editorial: Use string notation for properties

spectranaut:
To be consistence internally and with all the changes in ECMA-262.

### DIFF
--- a/spec/annexes.html
+++ b/spec/annexes.html
@@ -36,7 +36,7 @@
       In Collator:
       <ul>
         <li>
-          Support for the Unicode extensions keys kn, kf and the parallel options properties numeric, caseFirst (<emu-xref href="#sec-initializecollator"></emu-xref>)
+          Support for the Unicode extensions keys *"kn"*, *"kf"* and the parallel options properties *"numeric"*, *"caseFirst"* (<emu-xref href="#sec-initializecollator"></emu-xref>)
         </li>
         <li>
           The set of supported *"co"* key values (collations) per locale beyond a default collation (<emu-xref href="#sec-intl-collator-internal-slots"></emu-xref>)
@@ -125,7 +125,7 @@
       <emu-xref href="#sec-the-intl-collator-constructor"></emu-xref>, <emu-xref href="#sec-intl-numberformat-constructor"></emu-xref>, <emu-xref href="#sec-intl-datetimeformat-constructor"></emu-xref> In ECMA-402, 1st Edition, constructors could be used to create Intl objects from arbitrary objects. This is no longer possible in 2nd Edition.
     </li>
     <li>
-      <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the _length_ property of the function object _F_ was set to 0. In 2nd Edition, _length_ is set to 1.
+      <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the *"length"* property of the function object _F_ was set to *0*. In 2nd Edition, *"length"* is set to *1*.
     </li>
   </ul>
 </emu-annex>
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -114,7 +114,7 @@
       </emu-alg>
 
       <p>
-        The value of the `length` property of the *supportedLocalesOf* method is 1.
+        The value of the *"length"* property of the *supportedLocalesOf* method is 1.
       </p>
     </emu-clause>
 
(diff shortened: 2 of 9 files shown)


spec/annexes.html
```diff
@@ -125,7 +125,7 @@ <h1>Additions and Changes That Introduce Incompatibilities with Prior Editions</
       <emu-xref href="#sec-the-intl-collator-constructor"></emu-xref>, <emu-xref href="#sec-intl-numberformat-constructor"></emu-xref>, <emu-xref href="#sec-intl-datetimeformat-constructor"></emu-xref> In ECMA-402, 1st Edition, constructors could be used to create Intl objects from arbitrary objects. This is no longer possible in 2nd Edition.
     </li>
     <li>
-      <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the _length_ property of the function object _F_ was set to 0. In 2nd Edition, _length_ is set to 1.
+      <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the *"length"* property of the function object `F` was set to *0*. In 2nd Edition, *"length"* is set to *1*.
```
rkirsling: I think `function object _F_` can stay as-is here, since it's a spec variable in the linked algorithm.

spec/collator.html
```diff
@@ -114,7 +114,7 @@ <h1>Intl.Collator.supportedLocalesOf ( _locales_ [ , _options_ ] )</h1>
       </emu-alg>
 
       <p>
-        The value of the `length` property of the *supportedLocalesOf* method is 1.
+        The value of the *"length"* property of the *supportedLocalesOf* method is 1.
```
rkirsling: Looks like `*supportedLocalesOf*` → <code>\`supportedLocalesOf\`</code> got changed in a few spots and missed in a few spots. (Here's one of the missed spots.)

spec/locale-sensitive-functions.html
```diff
@@ -32,7 +32,7 @@ <h1>String.prototype.localeCompare ( _that_ [ , _locales_ [ , _options_ ] ] )</h
       </emu-alg>
 
       <p>
-        The value of the `length` property of the *localeCompare* method is 1.
+        The value of the *"length"* property of the *localeCompare* method is 1.
```
rkirsling: Here's another method that could be updated to backticks (though I guess that may go beyond the title of this PR).

spec/datetimeformat.html
```diff
@@ -754,11 +754,11 @@ <h1>Intl.DateTimeFormat.prototype.resolvedOptions ()</h1>
       </emu-table>
 
       <p>
-        For web compatibility reasons, if the property hourCycle is set, the hour12 property should be set to *true* when hourCycle is *"h11"* or *"h12"*, or to *false* when hourCycle is *"h23"* or *"h24"*.
+        For web compatibility reasons, if the property *"hourCycle"* is set, the *"hour12"* property should be set to *true* when *"hourCycle"* is *"h11"* or *"h12"*, or to *false* when *"hourCycle"* is *"h23"* or *"h24"*.
       </p>
 
       <emu-note>
-        In this version of the ECMAScript 2020 Internationalization API, the timeZone property will be the name of the default time zone if no timeZone property was provided in the options object provided to the Intl.DateTimeFormat constructor. The first edition left the timeZone property *undefined* in this case.
+        In this version of the ECMAScript 2020 Internationalization API, the *"timeZone"* property will be the name of the default time zone if no *"timeZone"* property was provided in the options object provided to the Intl.DateTimeFormat constructor. The first edition left the *"timeZone"* property *undefined* in this case.
```
rkirsling: I'm not certain how far you want to go in this PR, but I suppose <code>\`Intl.DateTimeFormat\` constructor</code> might be ideal in the end.

spectranaut: I think it would have to be a different PR -- because I intentionally didn't change it, actually. Many (if not all...?) global objects in ECMA-262 are not styled, like `Array` here for example: https://tc39.es/ecma262/#sec-array-objects

Maybe we should change it in both locations.

spec/collator.html
spectranaut: I'll make a new PR for these, good eye!

spec/locale-sensitive-functions.html
spectranaut: Making a new PR for these.

spec/datetimeformat.html
rkirsling: Oh wow, you're quite right, it really is quite inconsistent.
I'll ask about this on IRC and maybe create a 262 issue for it!

rkirsling: Just for visibility: https://github.com/tc39/ecma262/pull/1860

spectranaut: Cool! If this goes through on 262 I'll make a PR on 402.

PR #401, opened 2020-01-11
Editorial: Capitalize field names for recorded returned by ToLocalTime

spectranaut:
The algorithm [PartitionDateTimePattern](https://tc39.es/ecma402/#sec-partitiondatetimepattern) actually expects these field names to be capitalized, re: the step:
```
Let v be the value of tm's field whose name is the Internal Slot column of the matching row.
```
`tm` is the record returned to [ToLocalTime](https://tc39.es/ecma402/#sec-tolocaltime), and the Internal Slot column has all fields capitalized.


leobalter:
Proceeded with a manual merge (resolving a small git conflict) here: 5a542c

PR #414, opened 2020-03-02
Revert changes from calendar and numberingSystem options

leobalter:
This is a temporary revert commit for #175 until a canonicalization for Unicode extensions are
added through ResolveLocale.

Ref https://github.com/tc39/proposal-intl-locale/issues/96

This commit should be reverted along the proposed solution. Changes are simplified in this single commit to avoid reverting all the 7 original commits from #175.

cc @anba @jswalden @littledan @caiolima.

---

Considering the changes from #175 are already on the master branch and we need to have the "release candidate cut" set before the next TC39 meeting, this PR will help us setting the ideal path, we can pick one of these:

1. Don't merge this PR, the release candidate for ECMA-402 2020 will ship without the changes proposed in tc39/proposal-intl-locale#96. The next ECMA-402 2021 will ideally ship with the canonicalization fixes.
2. Merge this PR, the ECMA-402 2020 RC will ship without the proposed fixes. Right after the Release Candidate Cut, this PR can be reverted. Fixes will follow along. ECMA-402 2021 would ship with everything.

Unfortunately I don't see a way to push everything (including canonicalization fixes) for the ECMA-402 2020, technically we need to communicate and confirm consensus for the changes in the next TC39 meeting and I'd rather not rush things.

---

My preference is option 1, but I'd go with number 2 if anyone opposes to it. This PR is to keep both options equally available.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -73,26 +73,12 @@
         The abstract operation InitializeDateTimeFormat accepts the arguments _dateTimeFormat_ (which must be an object), _locales_, and _options_. It initializes _dateTimeFormat_ as a DateTimeFormat object. This abstract operation functions as follows:
       </p>
 
-      <p>
-        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
-      </p>
-
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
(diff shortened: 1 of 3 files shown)


sffc:
Thanks for preparing this.  We achieved consensus last meeting on the fix to https://github.com/tc39/proposal-intl-locale/issues/96, with @jswalden as the champion to fix it.  I'd say to set a deadline (say, March 13, which is 1 week before the 10-day TC39 deadline) for the PR, such that we can review and merge it.  If we don't have the fix by that time, then we can submit this PR.

leobalter:
I'm closing this down due to #418, there is no need to revert these changes for now.

Issue #413, opened 2020-02-27
Canonicalise Unicode extensions options added through ResolveLocale


leobalter:
The PR #175 landed after `Intl.Locale`, but didn't fix the issue outlined in <https://github.com/tc39/proposal-intl-locale/issues/96>. Maybe revert the PR for now?

_Originally posted by @anba in https://github.com/tc39/ecma402/pull/175#issuecomment-591879408_

leobalter:
@anba as we are currently at the Intl meeting I'm taking the opportunity to see if we can make a decision here so we could avoid reverting the PR.

If it requires extra time, I'll need to postpone #175 for after the ES2020 cut anyway.

leobalter:
and thanks for reminding me of this issue.

sffc:
IIRC, @jswalden was going to help write up some spec for this.

PR #418, opened 2020-03-16
Various fixes for ECMA-402 2020

anba:
I've started with fixing #413, but things to fix just piled up. :man_shrugging: 

Other potential issues I've noticed while working on this PR:

1. There doesn't seem to be a good CLDR replacement for RFC 4647, so we're still stuck with some references to "BCP 47" which don't quite mean "Unicode BCP 47 locale identifiers" (cf. "BCP 47 language priority list").
2. The number formatter is a bit loose with performing mathematical operations on mixed Number/BigInt inputs. I think ECMA-262 is nowadays a bit more strict in this regard.


leobalter:
> things to fix just piled up. 🤷‍♂

I never liked things to fix just piling up the way I like it here.

I like everything I see here and honestly this looks good enough for me to get a candidate version for 2020. I'd have a rollback version in case it becomes necessary.

---

@sffc let's get this for the TC39-TG2 meeting next week?

sffc:
Thanks for all the detail work!  I think we should definitely get this in before we cut 2020.  (I haven't completely reviewed it yet)

spec/numberformat.html
```diff
@@ -175,7 +175,7 @@ <h1>PartitionNumberPattern ( _numberFormat_, _x_ )</h1>
         1. Let _exponent_ be 0.
         1. If _x_ is *NaN*, then
           1. Let _n_ be an implementation- and locale-dependent (ILD) String value indicating the *NaN* value.
-        1. Else if _x_ is not a finite Number or BigInt,
+        1. Else if _x_ is not a finite Number,
```
jswalden: "is a Number that is not finite", or "is a non-finite Number", because by one reading a BigInt is not a finite Number?

spec/negotiation.html
```diff
@@ -207,6 +207,11 @@ <h1>ResolveLocale ( _availableLocales_, _requestedLocales_, _options_, _relevant
           1. If _options_ has a field [[&lt;_key_&gt;]], then
             1. Let _optionsValue_ be _options_.[[&lt;_key_&gt;]].
             1. Assert: Type(_optionsValue_) is either String, Undefined, or Null.
+            1. If Type(_optionsValue_) is String, then
+              1. Let _optionsValue_ be the string _optionsValue_ after performing the algorithm steps to transform Unicode extension values to canonical syntax per <a href="https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers">Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode Locale Identifiers</a>.
+              1. Let _optionsValue_ be the string _optionsValue_ after performing the algorithm steps to replace Unicode extension values with their canonical form per <a href="https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers">Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode Locale Identifiers</a>.
```
jswalden: This seems like it needs to be done with respect to _key_, after all "ca-islamicc" should be transformed but "nu-islamicc" shouldn't.  Maybe something like

```
1. Treating _key_ as a <a href="https://unicode.org/reports/tr35/#unicode_locale_extensions">`key`</a> and _optionsValue_ as its associated <a href="https://unicode.org/reports/tr35/#unicode_locale_extensions">`type`</a> in a Unicode locale extension extension sequence,
  1. Transform _optionsValue_ to have canonical syntax per <a href="https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers">Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode Locale Identifiers</a>.  (This will convert to lowercase.)
  1. Transform _optionsValue_ to be in canonical form per <a href="https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers">Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode Locale Identifiers</a>.  (This will replace an alias with its canonical form.)
```

instead of the two steps here?  Best I've got here at this point...

anba: Added "treating _key_ as `ukey' and _optionsValue_ as `uvalue`" to both steps, because I didn't want to add a new algorithm step notation.

PR #430, opened 2020-04-28
Normative: Make @@toStringTag values more meaningful

### LINKED ISSUE #176: Why are the toStringTag values "Object"?
It would seem more intuitive that, for example, `Intl.Collator[Symbol.toStringTag]` would be `"Intl.Collator"`. Is this done for compatibility reasons, where we're aware of a site that depends on `toString()` on Intl objects returning `"[Object object]"`? For new APIs, should we use `"Object"` or the name of the class, if we're not constrained by compatibility?

shvaikalesh:
Resolves #176.

If this reaches consensus, I will proceed with `test262` coverage and WebKit implementation.

_cc_ @littledan @Ms2ger

### DIFF
--- a/spec/annexes.html
+++ b/spec/annexes.html
@@ -155,5 +155,17 @@
     <li>
       <emu-xref href="#sec-intl.datetimeformat.prototype.format"></emu-xref> In ECMA-402, 1st Edition, the *"length"* property of the function object _F_ was set to *0*. In 2nd Edition, *"length"* is set to *1*.
     </li>
+    <li>
+      <emu-xref href="#sec-intl.collator.prototype-@@tostringtag"></emu-xref> In ECMA-402, 7th Edition, the @@toStringTag property of `Intl.Collator.prototype` was set to *"Object"*. In 8th Edition, @@toStringTag is set to *"Intl.Collator"*.
+    </li>
+    <li>
+      <emu-xref href="#sec-intl.datetimeformat.prototype-@@tostringtag"></emu-xref> In ECMA-402, 7th Edition, the @@toStringTag property of `Intl.DateTimeFormat.prototype` was set to *"Object"*. In 8th Edition, @@toStringTag is set to *"Intl.DateTimeFormat"*.
+    </li>
+    <li>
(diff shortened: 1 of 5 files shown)


Ms2ger:
This seems fine to me in theory, but the main question is if implementers will implement it.

shvaikalesh:
V8, JSC, and SpiderMonkey follow the current spec (except `Intl.PluralRules[Symbol.toStringTag]` is not defined in Firefox 75).

I am sure that with a few failing tests that are easy to fix, implementations will come.

sffc:
Thanks for this; I think browsers will implement it.  I added it to the agenda to discuss at the next TG2 call, and then we can have it approved at the June TC39.

littledan:
I'm in favor of this; I bet it'll be web-compatible, but we could roll it back if needed.

leobalter:
@shvaikalesh I believe this will be well accepted. It would be very helpful if you could add the respective coverage in Test262. This way we have a good status of matching conformance and perhaps a vision of (non) breaking challenges.

FrankYFTang:
We need to update the following test262 tests right after we land this PR

intl402/Collator/instance-class.js
intl402/Collator/prototype/builtin.js
intl402/DateTimeFormat/prototype/builtin.js
intl402/DateTimeFormat/instance-class.js
intl402/NumberFormat/prototype/builtin.js
intl402/NumberFormat/instance-class.js
intl402/NumberFormat/prototype/toStringTag/configurable.js
intl402/NumberFormat/prototype/toStringTag/prop-desc.js
intl402/PluralRules/prototype/builtins.js

shvaikalesh:
@FrankYFTang Thank you for extensive list, I will make sure to update those in https://github.com/tc39/test262/pull/2634.

FrankYFTang:
FYI - bug for v8 to sync with this PR https://bugs.chromium.org/p/v8/issues/detail?id=10554

leobalter:
> Is this a thing we want to document in [Annex B - Additions and Changes That Introduce Incompatibilities with Prior Editions](https://tc39.es/ecma402/#annex-incompatibilities)?

yes, we need to add that.

@shvaikalesh would you mind adding it?

FrankYFTang:
TC39 reach consensus this morning.

Issue #425, opened 2020-04-05
Confusing description of IsStructurallyValidLanguageTag() operation


iamstolis:
`IsStructurallyValidLanguageTag()` operation [states](https://tc39.es/ecma402/#sec-isstructurallyvalidlanguagetag) that it

> verifies that the locale argument represents a well-formed "Unicode BCP 47 locale identifier" as specified in Unicode Technical Standard 35 section 3.2,

It also says that

> The abstract operation returns true if locale can be generated from the EBNF grammar in section 3.2 of the Unicode Technical Standard 35, starting with unicode_locale_id, and does not contain duplicate variant or singleton subtags (other than as a private use subtag). It returns false otherwise.

These requirements are inconsistent. The mentioned grammar does not describe "Unicode BCP 47 locale identifier". It describes "Unicode CLDR locale identifier", i.e. it describes identifiers that support some backward compatibility syntax (`root` subtag, underscores as separator, tags starting with script subtag) that is not allowed in "Unicode BCP 47 locale identifier", see [Unicode Technical Standard 35 section 3.3](https://unicode.org/reports/tr35/#BCP_47_Conformance). Please, fix/improve the description of this operation such that it is clear what this operation is supposed to do.

sffc:
@anba @jswalden @FrankYFTang

zbraniecki:
It should refer to `Unicode BCP 47 locale identifier` - not the CLDR one.

littledan:
I don't really see the inconsistency. The grammar describes Unicode locale identifiers in general, and Unicode BCP 47 locale identifiers are a subset of this, as described in section 3.3. But the grammar is in section 3.2; there's no other grammar we could refer to. Should we link directly to section 3.3 *in addition* for clarity?

iamstolis:
> I don't really see the inconsistency.

I find it hard not to see the inconsistency when one sentence says that the operation verifies that the argument

> represents a well-formed "Unicode BCP 47 locale identifier"

and the following sentence says the equivalent of: The operation returns `true` if the argument is "Unicode CLDR locale identifier" (and returns `false` otherwise).

[...]

PR #431, opened 2020-05-02
Editorial: Add note for IsStructurallyValidLanguageTag

ryzokuken:
Add a note for IsStructurallyValidLanguageTag, clarifying the exact
variant of locale indentifer that's expected and point to a resource
that clarifies the differences between the two.

Fixes: https://github.com/tc39/ecma402/issues/425

/cc @littledan

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -52,6 +52,10 @@
       <p>
         The abstract operation returns true if _locale_ can be generated from the EBNF grammar in section 3.2 of the Unicode Technical Standard 35, starting with `unicode_locale_id`, and does not contain duplicate variant or singleton subtags (other than as a private use subtag). It returns false otherwise. Terminal value characters in the grammar are interpreted as the Unicode equivalents of the ASCII octet values given.
       </p>
+
+      <emu-note>
+        This function <em>specifically</em> accepts a "Unicode BCP 47 locale identifier", i.e. the backwards compatible syntax from "Unicode CLDR locale identifier" is not accepted. The difference between the two syntaxes is specified in <a href="https://unicode.org/reports/tr35/#BCP_47_Conformance">Unicode Technical Standard 35 section 3.3</a>.
+      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-canonicalizeunicodelocaleid" aoid="CanonicalizeUnicodeLocaleId">


littledan:
Rereading @iamstolis's comments, I guess the change we need to make is in the immediately preceding paragraphs, adding something right after

>  can be generated from the EBNF grammar in section 3.2 of the Unicode Technical Standard 35

to explain that we're taking into account the restrictions for section 3.3, and not accepting everything within the grammar.

I'm kinda wondering why we have that paragraph at all. It's really redundant. Maybe that whole thing should be a note? It's not great to have normative text that's duplicated like that. @anba do you know?

spec/locales-currencies-tz.html
```diff
@@ -52,6 +52,10 @@ <h1>IsStructurallyValidLanguageTag ( _locale_ )</h1>
       <p>
         The abstract operation returns true if _locale_ can be generated from the EBNF grammar in section 3.2 of the Unicode Technical Standard 35, starting with `unicode_locale_id`, and does not contain duplicate variant or singleton subtags (other than as a private use subtag). It returns false otherwise. Terminal value characters in the grammar are interpreted as the Unicode equivalents of the ASCII octet values given.
       </p>
+
+      <emu-note>
+        This function <em>specifically</em> accepts a "Unicode BCP 47 locale identifier", i.e. the backwards compatible syntax from "Unicode CLDR locale identifier" is not accepted. The difference between the two syntaxes is specified in <a href="https://unicode.org/reports/tr35/#BCP_47_Conformance">Unicode Technical Standard 35 section 3.3</a>.
```
jswalden, line 57: I suppose this gets the job done, sort of.  But -- it seems like the crux of the confusion is that the bullet points give one explanation for when the function returns *true*; then the paragraph just above this gives a second, and inconsistent, explanation for when the function returns *true*.  The bullets describe a narrower syntax; the paragraph describes a broader syntax.  Why do we have (arguably) two explanations?

Or, stepping back from those nitty-gritty details: if the bullets already say this only recognizes a "Unicode BCP 47 locale identifier", adding a note that has to _repeat_ that claim indicates something has already gone awry.

Or to put it the way I previously noted, https://github.com/tc39/ecma402/pull/429 would address this concern by listing the requirements of the algorithm, then leaving the _implications_ of those requirements to a note that neither reiterates what those requirements say, nor (even implicitly) contradicts them.

anba:
> I'm kinda wondering why we have that paragraph at all. It's really redundant. Maybe that whole thing should be a note? It's not great to have normative text that's duplicated like that. @anba do you know?

The duplicate information was already present in the [first edition](https://www.ecma-international.org/ecma-402/1.0/index.html#sec-6.2.2). Except that at that time there wasn't any (semantic) difference between the bullet points and the text. 

If I had to guess, I'd say the text was only added to explicitly define the return values of the function. And possibly to state that the duplicate singleton `x` in `und-x-x` isn't an error. 

It probably doesn't hurt to keep:
> Terminal value characters in the grammar are interpreted as the Unicode equivalents of the ASCII octet values given.

but even without that sentence the intent should be clear.

ryzokuken:
@anba @jswalden do we still need this?

jswalden:
I don't think we do, but then most of the [`IsStructurallyValidLanguageTag`](https://tc39.es/ecma402/#sec-isstructurallyvalidlanguagetag) spec text is my fault at this point, so I'm a little biased.  :-)

The current text is clear that backwards compatibility syntax is not accepted.

The current text _does_ clarify that any string it accepts must be a "Unicode BCP 47 locale identifier", in a manner that -- if you work through it -- also restricts beyond what that term presently means.

And the current text links to relevant UTS35 text.

This covers everything the proposed patch did (except for it not including #429's additional restrictions).

ryzokuken:
@jswalden that sounds good to me, I'll close this PR in that case.

PR #433, opened 2020-05-03
Editorial: Alphabetize extension keys.

### LINKED ISSUE #427: Should [[RelevantExtensionKeys]] be pre-alphabetized?
If UTS 35 expects a canonical locale identifier to have its extensions alphabetized, should we update the [[RelevantExtensionKeys]] lists for [Collator](https://tc39.es/ecma402/#sec-intl-collator-internal-slots) and [DateTimeFormat](https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots) to have `kf` before `kn` and `hc` before `nu`, respectively?

There's no alphabetization to do if the list is already in order, and I don't believe this has any impact on the initialization of either class, since we access the options by name.

rkirsling:
Resolves #427.

The first commit alphabetizes all [[RelevantExtensionKeys]] slots, which has no normative impact due to UTS 35 canonicalization. It does not change the order of internal slots (though I suppose it could) or resolved options (which would make this normative).

The second commit here alphabetizes non-normative mentions of extension keys throughout 402.


sffc:
@jswalden @FrankYFTang

spec/datetimeformat.html
```diff
@@ -551,11 +551,11 @@ <h1>Internal slots</h1>
       </p>
 
       <p>
-        The value of the [[RelevantExtensionKeys]] internal slot is &laquo; *"ca"*, *"nu"*, *"hc"* &raquo;.
+        The value of the [[RelevantExtensionKeys]] internal slot is &laquo; *"ca"*, *"hc"*, *"nu"* &raquo;.
       </p>
 
       <emu-note>
-        Unicode Technical Standard 35 describes three locale extension keys that are relevant to date and time formatting, *"ca"* for calendar, *"tz"* for time zone, *"hc"* for hour cycle, and implicitly *"nu"* for the numbering system of the number format used for numbers within the date format. DateTimeFormat, however, requires that the time zone is specified through the *"timeZone"* property in the options objects.
+        Unicode Technical Standard 35 describes four locale extension keys that are relevant to date and time formatting: *"ca"* for calendar, *"hc"* for hour cycle, *"nu"* for numbering system (of formatted numbers), and *"tz"* for time zone. DateTimeFormat, however, requires that the time zone is specified through the *"timeZone"* property in the options objects.
```
leobalter, line 558: off-by-one

Issue #402, opened 2020-01-21
Default hour-cycle computation can resolve to non-preferred hour format


anba:
<https://github.com/tc39/test262/blob/master/test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js> is currently failing in SpiderMonkey, revealing a potential spec issue. 

Reduced test case:
```js
function hourCycle(hour12) {
  return new Intl.DateTimeFormat("en", {hour: "numeric", hour12}).resolvedOptions().hourCycle;
}

print("default:", hourCycle(undefined));
print("hour12=true:", hourCycle(true));
print("hour12=false:", hourCycle(false));
```

Returns in SpiderMonkey:
```
default: h12
hour12=true: h12
hour12=false: h23
```

But in V8 (and per current spec):
```
default: h12
hour12=true: h12
hour12=false: h24
```

[InitializeDateTimeFormat](https://tc39.es/ecma402/#sec-initializedatetimeformat), step 30.d for `hcDefault=h12`, computes `hc=hc12` for `hour12=true` resp. `hc=hc24` for `hour12=false`.

But when we're consulting CLDR's [supplementalData.xml](https://github.com/unicode-org/cldr/blob/6f276d3251727298124b931d8333aee5ea7a1026/common/supplemental/supplementalData.xml#L4777-L4778) for "US" [1], the preferred hour formats are either "h" or "H", which means either "hc12" or "hc23". So "hc24" is likely an unexpected result for US, because "k" isn't listed in the preferred hour formats for that region. (See [2] for "h", "H", and "k".)

[1] Adding likely subtags to the "en" locale results in "en-Latn-US".
[2] <https://unicode.org/reports/tr35/tr35-dates.html#dfst-hour>

anba:
cc @zbraniecki

mobz:
See also https://bugs.chromium.org/p/chromium/issues/detail?id=1045791

littledan:
This issue seems somewhat important--it is a sort of user-observable "bug". Should we put it on the ES2021 milestone?

mobz:
I consider it a bug. The assumption in the spec is that h11 maps to h23 and h12 maps to h24, as a logical preference is wrong.

We can see from various bugs in date formatting packages that are following the spec that this is surprising.

hourCycle options map to keys in https://unicode.org/reports/tr35/tr35-dates.html#dfst-hour
K = h11 = 0-11
h = h12 = 1-12
H = h23 = 0-23
[...]

PR #436, opened 2020-05-13
Normative: Better defaults for hourCycle

mobz:
Addresses #402 

Fixes a bugs in the spec where;
* hour12: false would resolve hourCycle to h24 in some locales. This is unexpected as those locales prefer h23
* hour12: true could resolve to hourCycle to h11 is some locales. This is unexpected as those locales prefer h12

Current logic maps h11 to h23 and h12 to h24 which is unexpected. eg
* if hcDefault is h12 but hour12 is false set hourCycle to h24
* if hcDefault is h23 but hour12 is true set hourCycle to h11

New logic
* h12 is always preferred for hour12 is true unless hcDefault is h11
* h23 is always preferred for hour12 is false unless hcDefault is h24

This logic satisfies all locales allowed hour formats in the cldr unicode dataset
Note in particular that in the cldr dataset
* all regions prefer h12 and h23 over h11 and h24.
* No region defines h24 has an allowed format
* Only JP is defines h11 as an allowed (but also accepts h12), In practice h11 is always shown with a dayPeriod.

I believe this brings the spec in to line with the current implementation in spidermonkey and the previous implementation in v8 (see https://bugs.chromium.org/p/chromium/issues/detail?id=1045791 )

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -137,16 +137,16 @@
             1. Set _hc_ to _hcDefault_.
           1. If _hour12_ is not *undefined*, then
             1. If _hour12_ is *true*, then
-              1. If _hcDefault_ is *"h11"* or *"h23"*, then
+              1. If _hcDefault_ is *"h11"*, then
                 1. Set _hc_ to *"h11"*.
               1. Else,
                 1. Set _hc_ to *"h12"*.
             1. Else,
               1. Assert: _hour12_ is *false*.
-              1. If _hcDefault_ is *"h11"* or *"h23"*, then
-                1. Set _hc_ to *"h23"*.
-              1. Else,
+              1. If _hcDefault_ is *"h24"*, then
                 1. Set _hc_ to *"h24"*.
+              1. Else,
+                1. Set _hc_ to *"h23"*.
           1. Set _dateTimeFormat_.[[HourCycle]] to _hc_.
           1. If _dateTimeformat_.[[HourCycle]] is *"h11"* or *"h12"*, then
             1. Let _pattern_ be _bestFormat_.[[pattern12]].


anba:
This will select `"h12"` for the locale `"ja"` with `hour12 = true`, but SpiderMonkey uses `"h11"` for that case: 
```js
js> new Intl.DateTimeFormat("ja", {hour12: true, hour: "numeric"}).resolvedOptions().hourCycle
"h11"
```
---

I'd suggest to add two new internal slots to [13.3.3](https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots), to provide defaults for 12- and 24-hour-cycles for each locale. Something along the lines of:
```
[[LocaleData]].[[<locale>]].[[hourCycle12]] must be a String value equal to "h11" or "h12". 
[[LocaleData]].[[<locale>]].[[hourCycle24]] must be a String value equal to "h23" or "h24". 
```

InitializeDateTimeFormat can then be rewritten to:
```
If hour12 is true, then
    Set hc to dataLocaleData.[[hourCycle12]].
Else,
    Assert: hour12 is false.
    Set hc to dataLocaleData.[[hourCycle24]].
```

mobz:
@anba - do you know why spidermonkey selects h11 for ja. The only source of this data is cldr which specifies:

    <hours preferred="H" allowed="H h K" regions="JP"/>

(eg h23, h12 and h12)

Also confirmed that safari and old v8 both return h11 for ja

anba:
Because [common/main/ja.xml](https://github.com/unicode-org/cldr/blob/3f9fcc3e13848da0d598ba323f23d494f457dd96/common/main/ja.xml#L2986) contains:
```xml
<dateFormatItem id="h">aK時</dateFormatItem>
```

Per <https://unicode.org/reports/tr35/tr35-dates.html#Time_Data>, the order in `<hours>` is only relevant for the `C` skeleton symbol.

Ambroos:
Considering how many people this has bitten in the [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1045791) it would be really cool to see this merged so the old, more sensible behaviour can be restored in Chromium. It would allow us (and many others) to remove a bunch of workarounds in our code.

sffc:
@FrankYFTang @ryzokuken Is this PR in good shape to bring up for consensus at TC39?

mobz:
@sffc Changes planned - does not handle some edge cases correctly.

FrankYFTang:
any progress on this?

sffc:
This PR is stalled, so I'm closing it for now.  Continue discussion on #402.  There may be a new follow-up PR with a new author.  If @mobz wants to continue work, please re-open the PR.

AndyOGo:
I would really like to see this being resolved.
Is any help needed?

sffc:
> I would really like to see this being resolved.
> Is any help needed?

There is no one currently assigned to this issue.  You are free to open a new PR for consideration.

jufemaiz:
So … I've hit this … and I GAF about time and presentation. @mobz you ok if I take this up?

mobz:
@jufemaiz sure, I have not looked into it for a while, but I think I ran into an issue, where it looked like to solve everything satisfactorily I would need additional defaults defined and the scope was growing..., but I don't remember the details

Issue #437, opened 2020-05-15
Update Table 4: Numbering systems with simple digit mappings


FrankYFTang:
Table 4: Numbering systems with simple digit mappings
https://tc39.es/ecma402/#table-numbering-system-digits
lists "Numbering systems with simple digit mappings" that are easy to implement and mandated to be supported by ECMA402. [Newer version of CLDR ](https://raw.githubusercontent.com/unicode-org/cldr/master/common/supplemental/numberingSystems.xml) now include MORE type="numeric". I think we should include them into this table

```
        <numberingSystem id="adlm" type="numeric" digits="&#x1E950;&#x1E951;&#x1E952;&#x1E953;&#x1E954;&#x1E955;&#x1E956;&#x1E957;&#x1E958;&#x1E959;"/>
        <numberingSystem id="ahom" type="numeric" digits="&#x11730;&#x11731;&#x11732;&#x11733;&#x11734;&#x11735;&#x11736;&#x11737;&#x11738;&#x11739;"/>
       <numberingSystem id="bhks" type="numeric" digits="&#x11C50;&#x11C51;&#x11C52;&#x11C53;&#x11C54;&#x11C55;&#x11C56;&#x11C57;&#x11C58;&#x11C59;"/>
        <numberingSystem id="brah" type="numeric" digits="&#x11066;&#x11067;&#x11068;&#x11069;&#x1106A;&#x1106B;&#x1106C;&#x1106D;&#x1106E;&#x1106F;"/>
        <numberingSystem id="cakm" type="numeric" digits="&#x11136;&#x11137;&#x11138;&#x11139;&#x1113A;&#x1113B;&#x1113C;&#x1113D;&#x1113E;&#x1113F;"/>
        <numberingSystem id="cham" type="numeric" digits="꩐꩑꩒꩓꩔꩕꩖꩗꩘꩙"/>
        <numberingSystem id="diak" type="numeric" digits="&#x11950;&#x11951;&#x11952;&#x11953;&#x11954;&#x11955;&#x11956;&#x11957;&#x11958;&#x11959;"/>
        <numberingSystem id="gong" type="numeric" digits="&#x11DA0;&#x11DA1;&#x11DA2;&#x11DA3;&#x11DA4;&#x11DA5;&#x11DA6;&#x11DA7;&#x11DA8;&#x11DA9;"/>
[...]

PR #438, opened 2020-05-15
Update Table 4 Numbering systems

FrankYFTang:
Add more numbering systems which on CLDR 37

Close https://github.com/tc39/ecma402/issues/437

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -551,16 +551,16 @@
             <td>U+0C66 to U+0C6F</td>
           </tr>
           <tr>
-            <td>tirh</td>
-            <td>U+114D0 to U+114D9</td>
-          </tr>
-          <tr>
             <td>thai</td>
             <td>U+0E50 to U+0E59</td>
           </tr>
           <tr>
             <td>tibt</td>
             <td>U+0F20 to U+0F29</td>
+          </tr>          
+          <tr>
+            <td>tirh</td>
+            <td>U+114D0 to U+114D9</td>
           </tr>
           <tr>
             <td>vaii</td>
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -311,6 +311,14 @@
             </tr>
           </thead>
           <tr>
+            <td>adlm</td>
+            <td>U+1E950 to U+1E959</td>
+          </tr>
+          <tr>
+            <td>ahom</td>
+            <td>U+11730 to U+11739</td>
+          </tr>
+          <tr>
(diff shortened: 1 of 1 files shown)


sffc:
Would be good to get sign-off from the browser vendors.

@jswalden @litherum

FrankYFTang:
@rkirsling

FrankYFTang:
The list I proposed here (except "diak" and "segment") are ALREADY supported by two browser engines already:

* Safari version 13.1
* Chrome / v8 m81

"diak" and "segment" are new in CLDR / ICU and Chrome will support them in m84 with the update to ICU 67.1 .

rkirsling:
Sounds good to me. JSC is aiming to defer to ICU / CLDR directly as much as possible, even if our pace of updating is slower than than V8's (due to reliance on system ICU where available).

leobalter:
Let's bring this up for the meeting this week.

sffc:
This has ECMA-402 consensus.  We would like to see a more regular process of updating this list, but that doesn't block the current PR.  We need to achieve TC39 consensus before it can be merged.

anba:
So we don't expect any compatibility issues with other software (https://github.com/tc39/ecma402/issues/437#issuecomment-632054710)?

spec/numberformat.html
```diff
@@ -390,6 +550,10 @@ <h1>PartitionNotationSubPattern ( _numberFormat_, _x_, _n_, _exponent_ )</h1>
             <td>telu</td>
             <td>U+0C66 to U+0C6F</td>
           </tr>
+          <tr>
+            <td>tirh</td>
```
anba, line 554: Nit: Move `tirh` after `tibt`.

FrankYFTang: done

sffc:
Is this PR considered editorial or normative?  Is it normative because conforming implementations could become non-conforming based on this PR, or is it editorial because it reflects web reality?

FrankYFTang:
> Is this PR considered editorial or normative? Is it normative because conforming implementations could become non-conforming based on this PR, or is it editorial because it reflects web reality?

It is **normative** because https://ecma-international.org/ecma-402/#sec-partitionnumberpattern stated in step 8.e.iii.3 
"If the numberFormat.[[NumberingSystem]] matches one of the values in the "Numbering System" column of Table 3 below, then"

FrankYFTang:
So.. TC39 reach consensus this morning. @sffc @leobalter  @gibson042  could you approve?

PR #457, opened 2020-06-12
Normative: Adding Intl.DateTimeFormat dateStyle/timeStyle specification.

### LINKED ISSUE #108: short/medium/long/full style for Intl.DateTimeFormat
I'm working on trying to design an alignment between how Intl API works and OS Intl preferences work.

An example of such task is how to incorporate the date/time formatting OS user preferences into DateTimeFormat API.

Currently, we expect users of our API to list the elements of the date/time format as options, or, if none are listed, we take:
- for `Intl.DateTimeFormat` and `Date.prototype.toLocaleString`: hour/minute/second/day/month/year as `numeric`
- for `Date.prototype.toLocaleTimeString`: hour/minute/second as `numeric`
- for `Date.prototype.toLocaleDateString`: day/month/year as `numeric`

On the other hand, MacOS exposes four styles for date and time: short, medium, long and full.
Windows exposes short/long.

I see two solutions to bind those two APIs:

1) Create OS Intl prefs API that returns Intl API options.

``` javascript
let opts = intlOSPrefs.DateFormat(intlOSPrefs.formatFull);
  --> returns {weekday: 'long', month: 'long', day: 'numeric', year: 'numeric'}
let dtf = new Intl.DateTimeFormat(navigator.locales, opts);
dtf.format(date);
```

2) Add `style` option and change how we handle "defaults" in Intl.DateTimeFormat

In this approach, we would use the language in spec similar to one we use for timezones: if user did not specify a timezone, we try to retrieve it from the host OS and we land on the current defaults (all as `numeric`) only if we cannot get anything from the OS.

``` javascript
date.toLocaleString() == date.toLocaleString(navigator.locales, {
  style: 'medium'
}) == date.toLocaleString(navigator.locales, {
  day: 'numeric',
  month: 'short',
  year: 'numeric'
}); // In MacOS

date.toLocaleDateString(navigator.locales, {
  style: 'long'
}) == date.toLocaleString(navigator.locales, {
  day: 'numeric',
  month: 'long',
  year: 'numeric'
}); // default in MacOS


date.toLocaleDateString(navigator.locales, {
  style: 'full'
}) == date.toLocaleString(navigator.locales, {
  weekday: 'long',
  day: 'numeric',
  month: 'long',
  year: 'numeric'
}); // default in MacOS
```

That would require adding an additional step in `12.1.2 ToDateTimeOptions` as `6.a` and `7.a` where the implementation could reach to host OS for the preferences and only fallback to the current `6.a` and `7.a` if those cannot be retrieved.

In this approach, `style` option would be mutually exclusive with any of the token options.

Opinions? @caridy , @rxaviers , @jungshik, @srl295 ?

zbraniecki:
Fixes #108.

Implements https://github.com/tc39/proposal-intl-datetime-style

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -119,12 +119,24 @@
           1. Let _value_ be ? GetOption(_options_, _prop_, *"string"*, &laquo; the strings given in the Values column of the row &raquo;, *undefined*).
           1. Set _opt_.[[&lt;_prop_&gt;]] to _value_.
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
-        1. Let _formats_ be _dataLocaleData_.[[formats]].[[&lt;_calendar_&gt;]].
-        1. Let _matcher_ be ? GetOption(_options_, *"formatMatcher"*, *"string"*, &laquo; *"basic"*, *"best fit"* &raquo;, *"best fit"*).
-        1. If _matcher_ is *"basic"*, then
-          1. Let _bestFormat_ be BasicFormatMatcher(_opt_, _formats_).
+        1. Let _dateStyle_ be ? GetOption(_options_, *"dateStyle"*, *"string"*, &laquo; *"full"*, *"long"*, *"medium"*, *"short"* &raquo;, *undefined*).
+        1. Set _dateTimeFormat_.[[DateStyle]] to _dateStyle_.
+        1. Let _timeStyle_ be ? GetOption(_options_, *"timeStyle"*, *"string"*, &laquo; *"full"*, *"long"*, *"medium"*, *"short"* &raquo;, *undefined*).
+        1. Set _dateTimeFormat_.[[TimeStyle]] to _timeStyle_.
+        1. If _dateStyle_ is not *undefined* or _timeStyle_ is not *undefined*, then
+          1. For each row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, do
+            1. Let _prop_ be the name given in the Property column of the row.
+            1. Let _p_ be _opt_.[[&lt;_prop_&gt;]].
+            1. If _p_ is not *undefined*, then
+              1. Throw a *TypeError* exception.
+          1. Let _bestFormat_ be DateTimeStyleFormat(_dateStyle_, _timeStyle_, _dataLocaleData_).
         1. Else,
-          1. Let _bestFormat_ be BestFitFormatMatcher(_opt_, _formats_).
+          1. Let _formats_ be _dataLocaleData_.[[formats]].[[&lt;_calendar_&gt;]].
+          1. Let _matcher_ be ? GetOption(_options_, *"formatMatcher"*, *"string"*, &laquo; *"basic"*, *"best fit"* &raquo;, *"best fit"*).
+          1. If _matcher_ is *"basic"*, then
+            1. Let _bestFormat_ be BasicFormatMatcher(_opt_, _formats_).
+          1. Else,
+            1. Let _bestFormat_ be BestFitFormatMatcher(_opt_, _formats_).
         1. For each row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, in table order, do
           1. Let _prop_ be the name given in the Property column of the row.
           1. Let _p_ be _bestFormat_.[[&lt;_prop_&gt;]].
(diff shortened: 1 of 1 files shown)


spec/datetimeformat.html
```diff
@@ -119,12 +119,24 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
...
+          1. For each row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, do
+            1. Let _prop_ be the name given in the Property column of the row.
+            1. Let _p_ be _opt_.[[&lt;_prop_&gt;]].
+            1. If _p_ is not *undefined*, then
+              1. Throw a *TypeError* exception.
+          1. Let _bestFormat_ be DateTimeStyleFormat(_dateStyle_, _timeStyle_, _dataLocaleData_).
         1. Else,
-          1. Let _bestFormat_ be BestFitFormatMatcher(_opt_, _formats_).
+          1. Let _formats_ be _dataLocaleData_.[[formats]].[[&lt;_calendar_&gt;]].
+          1. Let _matcher_ be ? GetOption(_options_, *"formatMatcher"*, *"string"*, &laquo; *"basic"*, *"best fit"* &raquo;, *"best fit"*).
```
anba, line 135: Why was the step to retrieve `"formatMatcher"` moved into the else-branch, even though it's always validated in the proposal spec? The convention is to validate all options, even when they're (currently) unused.

spec/datetimeformat.html
```diff
@@ -181,6 +193,13 @@ <h1>ToDateTimeOptions ( _options_, _required_, _defaults_ )</h1>
             1. Let _prop_ be the property name.
             1. Let _value_ be ? Get(_options_, _prop_).
             1. If _value_ is not *undefined*, let _needDefaults_ be *false*.
+        1. Let _dateStyle_ be ? Get(_options_, *"dateStyle"*).
+        1. Let _timeStyle_ be ? Get(_options_, *"timeStyle"*).
+        1. If _dateStyle_ is not *undefined* or _timeStyle_ is not *undefined*, let _needDefaults_ be *false*.
+        1. If _required_ is *"date"* and _timeStyle_ is not *undefined*,
```
anba, line 199: Nit: Missing "then".

spec/datetimeformat.html
```diff
@@ -181,6 +193,13 @@ <h1>ToDateTimeOptions ( _options_, _required_, _defaults_ )</h1>
...
+          1. Throw a *TypeError* exception.
+        1. If _required_ is *"time"* and _dateStyle_ is not *undefined*,
```
anba, line 201: Ditto.

spec/datetimeformat.html
```diff
@@ -585,6 +632,9 @@ <h1>Internal slots</h1>
...
         </li>
+        <li>
+          [[LocaleData]][locale] must contain [[DateFormat]], [[TimeFormat]] and [[DateTimeFormat]] fields, the value of these fields are Records, where each of which has [[full]], [[long]], [[medium]] and [[short]] fields. For [[DateFormat]] and [[TimeFormat]], the value of these fields must be a record, which has a subset of the fields shown in Table 1, where each field must have one of the values specified for the field in Table 1. Each of the records must also have a pattern field, whose value is a String value that contains for each of the date and time format component fields of the record a substring starting with *"{"*, followed by the name of the field, followed by *"}"*. If the record has an hour field, it must also have a pattern12 field, whose value is a String value that, in addition to the substrings of the pattern field, contains a substring *"{ampm}"*. For [[DateTimeFormat]], the field value must be a string pattern which contains the strings *"{0}"* and *"{1}"*.
```
anba, line 636: I didn't spot this when reading the rendered proposal text, because it didn't include #349, but the proposal doesn't allow the style choice to be calendar-dependent. Can you update the PR to include the necessary changes to mirror #349?

spec/datetimeformat.html
```diff
@@ -687,6 +737,9 @@ <h1>Intl.DateTimeFormat.prototype.resolvedOptions ()</h1>
             1. Else, let _v_ be *undefined*.
           1. Else,
             1. Let _v_ be the value of _dtf_'s internal slot whose name is the Internal Slot value of the current row.
+          1. If the Internal Slot value of the current row is an Internal Slot value in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then
```
lroling8350, line 740: This seems to contradict the constructor of the DateTimeFormat which has logic to throw a type error if any of these fields are not undefined. From what i've seen chrome 83 decided not to throw a type error during construction and instead follow this logic while firefox 79 decided to throw an exception during construction.

anba: These internal slots are filled from the resolved pattern and not from the input arguments (cf. the "Set dateTimeFormat's internal slot whose name is the Internal Slot column of the row to p." step in InitializeDateTimeFormat). So this step is actually needed to be consistent with the "throw TypeError" steps in InitializeDateTimeFormat.

zbraniecki:
hi all. I won't have a chance to get this fix the issues listed above before tomorrow's TC39 call. I can either ask for help with the PR, or ask if we are ok requesting Stage 4 with the pending fixes to the PR.

sffc:
Bring the editorial issues to the committee's attention.  I think they should be fine as long as there are no normative changes that are pending.

spec/datetimeformat.html
zbraniecki: Are we good here then? `CreateDataPropertyOrThrow` throws `TypeError`.

spec/datetimeformat.html
zbraniecki: @anba - I synced the language and added the `calendars` field in the patch with feedback application. Let me know if you think it's enough to resolve this.

zbraniecki:
This proposal reached Stage 4 as of last week.

I applied the feedback and am awaiting responses from two reviewers before I request merge.

spec/datetimeformat.html
lroling8350: From my end this looks good, @anba cleared up my misunderstanding

zbraniecki:
@anba ping?

zbraniecki:
@leobalter @gibson042 - ping?

leobalter (approved):

leobalter:
LGTM, also for addressing https://github.com/tc39/ecma402/pull/457#discussion_r450232228

@anba please let me know if there is anything else we missed here.

PR #459, opened 2020-06-16
Normative: Allow Collator to get collation from option

### LINKED ISSUE #380: Add "collation" option to Intl.Collator
https://github.com/tc39/ecma402/pull/175
add "calendar" and "numberingSystem" option to Intl.DateTimeFormat and Intl.NumberFormat
Also notice  "numberingSystem"  is already an option in Intl.RelativeTimeFormat.

We should also consider add "collation" to Intl.Collator.

Notes by @littledan  in https://github.com/tc39/ecma402/pull/175 stated
"This patch leaves out "collation" because of a lack
of clear use cases."

Not sure what does that mean.

I think it is useful. For example, in Chinese, we may construct the following collator

var pinyin = new Intl.Collator("zh", {collator: "pinyin"});
var zhuyin = new Intl.Collator("zh", {collator: "zhuyin"});
var stroke = new Intl.Collator("zh", {collator: "stroke"});

FrankYFTang:
Collator already have "co" in  [[RelevantExtensionKeys]] internal slot and therefore the  usage by using "-u-co-$collationkey" is already supported in ECMA402.

This PR will also make our extension / option handling in ECMA402 consistent.

I check other object and I believe this is the only inconsistent handling between u extension and option now.

Close #380

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -14,6 +14,10 @@
         The abstract operation InitializeCollator accepts the arguments _collator_ (which must be an object), _locales_, and _options_. It initializes _collator_ as a *Collator* object. The following steps are taken:
       </p>
 
+      <p>
+        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+      </p>
+      
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. If _options_ is *undefined*, then
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -33,7 +33,7 @@
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
-        1. Let _collation_ be ? GetOption(_options_, `"collation"`, `"string"`, *undefined*, *undefined*).
+        1. Let _collation_ be ? GetOption(_options_, *"collation"*, *"string"*, *undefined*, *undefined*).
         1. If _collation_ is not *undefined*, then
           1. If _collation_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
         1. Set _opt_.[[co]] to _collation_.
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -15,7 +15,7 @@
       </p>
 
       <p>
-        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
+        The following algorithm refers to the `type` nonterminal from <a href="https://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
       </p>
       
       <emu-alg>
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -35,6 +35,10 @@
         1. Set _opt_.[[kn]] to _numeric_.
         1. Let _caseFirst_ be ? GetOption(_options_, *"caseFirst"*, *"string"*, &laquo; *"upper"*, *"lower"*, *"false"* &raquo;, *undefined*).
         1. Set _opt_.[[kf]] to _caseFirst_.
+        1. Let _collation_ be ? GetOption(_options_, `"collation"`, `"string"`, *undefined*, *undefined*).
+        1. If _collation_ is not *undefined*, then
+          1. If _collation_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
+        1. Set _opt_.[[co]] to _collation_.
         1. Let _relevantExtensionKeys_ be %Collator%.[[RelevantExtensionKeys]].
         1. Let _r_ be ResolveLocale(%Collator%.[[AvailableLocales]], _requestedLocales_, _opt_, _relevantExtensionKeys_, _localeData_).
         1. Set _collator_.[[Locale]] to _r_.[[locale]].
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -29,16 +29,16 @@
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _collation_ be ? GetOption(_options_, `"collation"`, `"string"`, *undefined*, *undefined*).
+        1. If _collation_ is not *undefined*, then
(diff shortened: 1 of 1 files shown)


spec/collator.html
```diff
@@ -35,6 +35,10 @@ <h1>InitializeCollator ( _collator_, _locales_, _options_ )</h1>
         1. Set _opt_.[[kn]] to _numeric_.
         1. Let _caseFirst_ be ? GetOption(_options_, *"caseFirst"*, *"string"*, &laquo; *"upper"*, *"lower"*, *"false"* &raquo;, *undefined*).
         1. Set _opt_.[[kf]] to _caseFirst_.
+        1. Let _collation_ be ? GetOption(_options_, `"collation"`, `"string"`, *undefined*, *undefined*).
+        1. If _collation_ is not *undefined*, then
+          1. If _collation_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
```
sffc, line 40: #175 also says:

```
1. Set _calendar_ to the result of mapping _calendar_ to lower case as as specified in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
```

spec/collator.html
```diff
@@ -35,6 +35,10 @@ <h1>InitializeCollator ( _collator_, _locales_, _options_ )</h1>
         1. Set _opt_.[[kn]] to _numeric_.
```
sffc, line 35: Is it necessary to add the following line from #175?

```



      <p>
        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
      </p>
```

spec/collator.html
FrankYFTang: No, we don't need that now. See https://github.com/tc39/ecma402/commit/ca92351087c1a66631f86900c1dc3c2d1ef30f10#diff-1f2e1333fe08618e8af2d7344b031dc1

spec/collator.html
FrankYFTang: done

sffc:
Looks good!  Let's try to get consensus at the next 402 meeting.

spec/collator.html
```diff
@@ -29,6 +33,10 @@ <h1>InitializeCollator ( _collator_, _locales_, _options_ )</h1>
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _collation_ be ? GetOption(_options_, `"collation"`, `"string"`, *undefined*, *undefined*).
+        1. If _collation_ is not *undefined*, then
+          1. If _collation_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
+        1. Set _opt_.[[co]] to _collation_.
```
leobalter, line 39:
```suggestion
        1. Let _collation_ be ? GetOption(_options_, *"collation"*, *"string"*, *undefined*, *undefined*).
        1. If _collation_ is not *undefined*, then
          1. If _collation_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
        1. Set _opt_.[[co]] to _collation_.
```

spec/collator.html
```diff
@@ -14,6 +14,10 @@ <h1>InitializeCollator ( _collator_, _locales_, _options_ )</h1>
         The abstract operation InitializeCollator accepts the arguments _collator_ (which must be an object), _locales_, and _options_. It initializes _collator_ as a *Collator* object. The following steps are taken:
       </p>
 
+      <p>
+        The following algorithm refers to the `type` nonterminal from <a href="http://www.unicode.org/reports/tr35/#Unicode_locale_identifier">UTS 35's Unicode Locale Identifier grammar</a>.
```
anba, line 18: Nit: "https" instead of "http".

sffc:
TC39 gave consensus for this change on 2020-07-20.

FrankYFTang:
Just FYI
v8 implementation in https://chromium-review.googlesource.com/c/v8/v8/+/2309149 
v8 bug https://bugs.chromium.org/p/v8/issues/detail?id=10732

FrankYFTang:
I still have "You’re not authorized to merge this pull request." next to the grey "Merge pull request" button.

FrankYFTang:
FYI- landed into v8 and should be available in Chrome 86

sffc:
This PR has TG1 and TG2 consensus:

- https://github.com/tc39/ecma402/blob/master/meetings/notes-2020-07-09.md#normative-allow-collator-to-get-collation-from-option-459
- https://github.com/tc39/notes/blob/master/meetings/2020-07/july-20.md#ecma402-update

Constellation:
FYI, JSC enabled it on Trunk https://trac.webkit.org/changeset/267102/webkit

FrankYFTang:
> FYI, JSC enabled it on Trunk https://trac.webkit.org/changeset/267102/webkit

thanks. I updated https://github.com/tc39/ecma402/wiki/Proposal-and-PR-Progress-Tracking

leobalter (approved):

ryzokuken:
@FrankYFTang @sffc https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#parameters already lists `"collation"` as an accepted option for `Intl.Collator`. Does this PR need any further MDN documentation?

FrankYFTang:
It seems they list the possible values there. Not sure what is missing. Please at least add "zhuyin" as possible value

    <keyword>
        <key name="co" description="Collation type key" alias="collation">
            <type name="big5han" description="Pinyin ordering for Latin, big5 charset ordering for CJK characters (used in Chinese)"/>
            <type name="compat" description="A previous version of the ordering, for compatibility" since="26"/>
            <type name="dict" description="Dictionary style ordering (such as in Sinhala)" alias="dictionary"/>
            <type name="direct" description="Binary code point order (used in Hindi)" deprecated="true"/>
            <type name="ducet" description="The default Unicode collation element table order" since="2.0.1"/>
            <type name="emoji" description="Recommended ordering for emoji characters" since="27"/>
            <type name="eor" description="European ordering rules" since="24"/>
            <type name="gb2312" description="Pinyin ordering for Latin, gb2312han charset ordering for CJK characters (used in Chinese)" alias="gb2312han"/>
            <type name="phonebk" description="Phonebook style ordering (such as in German)" alias="phonebook"/>
            <type name="phonetic" description="Phonetic ordering (sorting based on pronunciation)"/>
            <type name="pinyin" description="Pinyin ordering for Latin and for CJK characters (used in Chinese)"/>
            <type name="reformed" description="Reformed ordering (such as in Swedish)"/>
            <type name="search" description="Special collation type for string search" since="1.9"/>
            <type name="searchjl" description="Special collation type for Korean initial consonant search" since="2.0.1"/>
            <type name="standard" description="Default ordering for each language"/>
            <type name="stroke" description="Pinyin ordering for Latin, stroke order for CJK characters (used in Chinese)"/>
            <type name="trad" description="Traditional style ordering (such as in Spanish)" alias="traditional"/>
            <type name="unihan" description="Pinyin ordering for Latin, Unihan radical-stroke ordering for CJK characters (used in Chinese)"/>
            <type name="zhuyin" description="Pinyin ordering for Latin, zhuyin order for Bopomofo and CJK characters (used in Chinese)" since="22"/>
        </key>

I think we should add all those missed from above (for example "zhuyin") except "search" and "standard" that is clearly stated not allowed in ECMA402

Issue #424, opened 2020-04-02
Explicitly state in the spec why we don't have parse methods


sffc:
We have received requests for parsing on numerous occasions; for example, #1, #147, and #342.  Besides the discussion in those threads, I've also written a [blog post](https://blog.sffc.xyz/post/190943794505/why-you-should-not-parse-localized-strings) on the subject.

Because such bugs won't stop coming in, and because the personnel on this committee may change and forget why we consistently have shot the idea down, I think it would be beneficial to add a paragraph to the specification document itself laying out why exactly we feel that parsing methods don't belong in ECMA-402.

littledan:
I've gotten this question lots of times too. Good to know about your blog post, so I can refer people to it in the future!

ryzokuken:
@sffc where do you think this note should go?

sffc:
Add a paragraph in Section 4 "Overview", either underneath one of the existing headings, or under a new heading if it doesn't fit on any of the existing headings.

https://tc39.es/ecma402/#overview

PR #463, opened 2020-06-19
editorial: add note about parsing

ryzokuken:
Add a brief note to the specification about the reason for non-existence
of parsing functionality.

Fixes: https://github.com/tc39/ecma402/issues/424

/cc @sffc @littledan

### DIFF
--- a/spec/overview.html
+++ b/spec/overview.html
@@ -43,6 +43,8 @@
     <p>
       The Intl object is used to package all functionality defined in the ECMAScript 2021 Internationalization API Specification to avoid name collisions.
     </p>
+
+    <emu-note>If you take a look at the specification, you would realize that while the API includes a variety of formatters, it does not provide any parsing facilities. This is something intentional that has been discussed extensively and concluded after weighing in all the benefits and drawbacks of including said functionality. You could learn more about the discussion on the <a href="https://github.com/tc39/ecma402/issues/342">issue tracker</a>.</emu-note>
   </emu-clause>
 
   <emu-clause id="sec-implementation-dependencies">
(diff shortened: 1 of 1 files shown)


anba:
Neither ECMA-402 nor ECMA-262 directly address the reader ("you"), so the style for this note feels a bit unusual.

spec/overview.html
```diff
@@ -43,6 +43,8 @@ <h1>API Overview</h1>
     <p>
       The Intl object is used to package all functionality defined in the ECMAScript 2021 Internationalization API Specification to avoid name collisions.
     </p>
+
+    <emu-note>If you take a look at the specification, you would realize that while the API includes a variety of formatters, it does not provide any parsing facilities. This is something intentional that has been discussed extensively and concluded after weighing in all the benefits and drawbacks of including said functionality. You could learn more about the discussion on the <a href="https://github.com/tc39/ecma402/issues/342">issue tracker</a>.</emu-note>
```
ljharb, line 47: > If you take a look at the specification, you would realize

this presumes that anyone who looks at the spec would realize anything in particular; this comes off as exclusionary (a la "just" or "simply"). Perhaps:
```suggestion
    <emu-note>While the API includes a variety of formatters, it does not provide any parsing facilities. This is intentional, has been discussed extensively, and concluded after weighing in all the benefits and drawbacks of including said functionality. See the discussion on the <a href="https://github.com/tc39/ecma402/issues/342">issue tracker</a>.</emu-note>
```
?

leobalter:
Let's see how it works using the PR tooling from GitHub...

ryzokuken:
Sorry @leobalter that you had to fix this up. I'd been planning to do it today :sweat_smile: Thanks!

PR #464, opened 2020-06-20
Change url to https

FrankYFTang:
@anba @gibson042 @leobalter 
Reminded by @anba in https://github.com/tc39/ecma402/pull/459

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -74,7 +74,7 @@
       </p>
 
       <p>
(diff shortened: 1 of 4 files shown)


srl295:
> I like it in theory, except the unicode.org SSL certificate has a bad track record for not being renewed on time, server going down, etc. @srl295, do you think HTTPS will be stable enough going forward to make this change?

First of all, help is needed here. So let's lend a hand instead of hacking around it.
"not renewed on time" is old news. The https certificate is an autorenewed letsencrypt certificate, so it won't expire.

PR #474, opened 2020-07-09
Editorial: use dotted intrinsic form

ljharb:
Since https://github.com/tc39/ecma262/pull/1376, the dotted form of intrinsic notation is preferred.

Additionally, https://github.com/tc39/ecma262/pull/2056 intends to remove these legacy intrinsics, once it is non-normative to do so. HTML has already been updated to no longer use them.

I did not delete the legacy ones from "conventions.html" yet, and I don't think that should happen until https://github.com/tc39/ecma262/pull/1948 is resolved (otherwise, removing items from that table is normative).


anba:
I think you missed to update `%Date_now%`?

ljharb:
@anba oops, thanks. updated!

PR #476, opened 2020-07-13
Editorial: Change "be this value" to "be the this value"

mpcsh:
I noticed we're not consistent about whether we write

> Let foo be *this* value

vs

> Let foo be the *this* value

262 consistently uses the latter form, so this PR just unifies to that language.


leobalter:
Thank you, @mpcsh !

PR #477, opened 2020-07-13
Editorial: Correct FormatNumber to FormatNumeric

FrankYFTang:
FormatNumber was renamed to FormatNumeric but didn't change this line while merging.

@anba  @gibson042  @littledan @rwaldron @leobalter


FrankYFTang:
Notice this error is also in https://ecma-international.org/ecma-402/ . Should we try to provide an erratum to correct the published version? @sffc

FrankYFTang:
@leobalter @gibson042  I do not have the permission to mege.

PR #485, opened 2020-07-20
Editorial: Use ECMA-262 StringIndexOf rather than %StringProto_indexOf%

gibson042:


leobalter (approved):

leobalter:
FYI, I resolved some minor conflicts - locally as well - and triple checked the diff.

PR #490, opened 2020-07-21
Editorial: Simplify assignment to _locale_.[[Numeric]]

gibson042:


anba:
See <https://github.com/tc39/proposal-intl-locale/issues/80> for why we need to test for the empty string, too.

gibson042:
Thanks, updated. Do we need something similar for CaseFirst? Web reality there is... inconsistent.

```sh
$ eshost -sx 'try{ print(JSON.stringify((new Intl.Locale("en-u-kf")).caseFirst)) }catch(x){ print("<unsupported>") }'
#### ChakraCore, engine262, Moddable XS, QuickJS
<unsupported>

#### JavaScriptCore
"yes"

#### SpiderMonkey
""

#### V8
"true"
```

anba:
No, CaseFirst doesn't need this special-casing. It's only needed for Numeric, because when the "numeric" property is read, it's treated as a boolean and stored by stringifying that boolean value (so we get either `"false"` or `"true"`).

Per [UTS 35](http://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers), canonicalised Unicode BCP 47 locale identifiers have all `type` subtags with the value `"true"` removed. For example the canonical form of "en-u-kn-true" is "en-u-kn". 

When we merge the language tag and the options record in [ApplyUnicodeExtensionToTag](https://tc39.es/ecma402/#sec-apply-unicode-extension-to-tag), "kn" values from the language tag, assuming "kn" is present, are either the empty string or a string unequal to `"true"`, because the language tag is always canonicalised. But option values for "numeric" are either `"true"` or `"false"`! So when we return from ApplyUnicodeExtensionToTag, `r.[[kn]]` can be `"true"` or the empty string, depending on where it was derived from, so we need to handle both cases.

Valid option values for CaseFirst are only "upper", "lower", "false". So `"true"` isn't allowed here, which means no special-casing needed. 

For your example, the correct output is the empty string. Returning `"true"` resp. `"yes"` is just a side-effect of showing the unfiltered ICU output.

PR #491, opened 2020-07-22
Editorial: Document constructor [[Call]] behavior

### LINKED ISSUE #384: Should all new constructors throw when called as a function?
Every constructor except PluralRules is specified to be callable as a function and supported as such:
```js
$ eshost -se 'typeof Intl === "undefined" ?
  "Intl: undefined" : 
  "Collator,NumberFormat,DateTimeFormat,PluralRules".split(",").map(name => {
    let src = `Intl.${name}()`, result;
    try {
      result = eval(src).constructor.name;
    } catch ( err ) {
      result = err.name;
    }
    return `${src}: ${JSON.stringify(result)}`;
  }).join("\n")'
#### Chakra, JavaScriptCore, SpiderMonkey, V8
Intl.Collator(): "Collator"
Intl.NumberFormat(): "NumberFormat"
Intl.DateTimeFormat(): "DateTimeFormat"
Intl.PluralRules(): "TypeError"

#### XS
Intl: undefined
```

PluralRules is intentionally different per https://github.com/tc39/proposal-intl-plural-rules/pull/24 , but the door was left open to bring it in line with the rest of spec. Can we reach consensus on whether or not the behavior of Collator, NumberFormat, and DateTimeFormat when called as a function is to be treated as the ECMA 402 convention or as an abandoned legacy, and document accordingly in https://tc39.es/ecma402/#locale-and-parameter-negotiation ?

gibson042:
Fixes #384

### DIFF
--- a/spec/overview.html
+++ b/spec/overview.html
@@ -16,7 +16,6 @@
     </p>
   </emu-clause>
 
-
   <emu-clause id="sec-api-overview">
     <h1>API Overview</h1>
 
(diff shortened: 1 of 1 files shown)


spec/overview.html
```diff
@@ -47,6 +46,14 @@ <h1>API Overview</h1>
     <emu-note>While the API includes a variety of formatters, it does not provide any parsing facilities. This is intentional, has been discussed extensively, and concluded after weighing in all the benefits and drawbacks of including said functionality. See the discussion on the <a href="https://github.com/tc39/ecma402/issues/342">issue tracker</a>.</emu-note>
   </emu-clause>
 
+  <emu-clause id="sec-api-conventions">
+    <h1>API Conventions</h1>
+
+    <p>
+      Every built-in constructor should behave as if defined inside a class, throwing a *TypeError* exception when called as a function (without NewTarget). For backwards compatibility with past editions, this does not apply to %Collator%, %NumberFormat%, or %DateTimeFormat%, each of which construct and return a new object when called as a function.
```
leobalter, line 53: It always bugs me that `NewTarget` goes without special quotes or between underscores in the spec source code. Although, this remains correct.

Issue #480, opened 2020-07-16
Require options to be an object


sffc:
Currently, we coerce the options argument to an object by calling `ToObject`.  @justingrant suggested in https://github.com/tc39/proposal-temporal/pull/693 that we should be stricter and throw instead.

I don't think we should change older Intl objects, due to backwards compatibility, but we can adopt this pattern in new objects.

anba:
If there's agreement to perform this change, we should definitely also try to apply it to the existing Intl constructors. And only if we detect any web-compat issues, back it out for the affected constructors. Similar to how we managed to restrict the "IntlLegacyConstructedSymbol" hack to just Intl.NumberFormat and Intl.DateTimeFormat, but not Intl.Collator.

ptomato:
There's a way to make this change non-normatively as well. We could change the definition of GetOption but make sure to use ToObject first on all objects passed to existing uses of GetOption. That way, the definition of GetOption could move into 262 when Temporal does, but the behaviour in Intl could remain unchanged.

I'll prepare a PR with both the editorial and normative changes, split into two commits, for discussion.

ptomato:
OK, this is in #493, one non-normative commit followed by a normative commit.

sffc:
Another option, which I mentioned in the meeting:

Keep `GetOption` defined as it is now, and move it to 262.  In 262, make new abstract operations as follows:

```html
<emu-clause id="sec-getoption" aoid="NormalizeOptionsObject">
  <h1>NormalizeOptionsOption ( _options_, _legacyMode_ )</h1>
  <p>
    The abstract operation NormalizeOptionsObject massages _options_ into an Object to be subsequently passed to GetOption.
  </p>
  <emu-alg>
    1. If _options_ is *undefined*, then
      1. Return ObjectCreate(*null*).
    1. If Type(_options_) is Object, then
      1. Return _options_.
    1. If _legacyMode_, then
      1. Return ToObject(_options_).
    1. Throw a *TypeError* exception.
  </emu-alg>
</emu-clause>
```

Change the 402 call sites as follows:

```html
<!-- Current 4 lines, present in several places, preceding calls to GetOption -->
1. If _options_ is *undefined*, then	
  1. Let _options_ be ObjectCreate(*null*).	
1. Else,	
  1. Let _options_ be ? ToObject(_options_).

<!-- Replace with: -->
1. Let _options_ be NormalizeOptionsObject(_options_, true).
```

[...]

PR #493, opened 2020-08-03
Editorial: GetOption reform

ptomato:
Discussion: #480

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -113,10 +113,7 @@
           1. Let _tag_ be _tag_.[[Locale]].
         1. Else,
           1. Let _tag_ be ? ToString(_tag_).
-        1. If _options_ is *undefined*, then
-          1. Let _options_ be ! ObjectCreate(*null*).
-        1. Else,
-          1. Let _options_ be ? ToObject(_options_).
+        1. Set _options_ to ? CoerceOptionsToObject(_options_).
         1. Set _tag_ to ? ApplyOptionsToTag(_tag_, _options_).
         1. Let _opt_ be a new Record.
         1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, *"string"*, *undefined*, *undefined*).
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -289,16 +289,41 @@
       </p>
 
       <emu-alg>
-        1. If _options_ is *undefined*, then
-          1. Let _matcher_ be *"best fit"*.
-        1. Else,
-          1. Let _options_ be ? ToObject(_options_).
-          1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
+        1. Set _options_ to ? CoerceOptionsToObject(_options_).
+        1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, *"string"*, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. If _matcher_ is *"best fit"*, then
           1. Let _supportedLocales_ be BestFitSupportedLocales(_availableLocales_, _requestedLocales_).
         1. Else,
           1. Let _supportedLocales_ be LookupSupportedLocales(_availableLocales_, _requestedLocales_).
         1. Return CreateArrayFromList(_supportedLocales_).
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-getoptionsobject" aoid="GetOptionsObject">
+      <h1>GetOptionsObject ( _options_ )</h1>
+      <p>
+        The abstract operation GetOptionsObject returns an Object suitable for use with GetOption, either _options_ itself or a default empty Object.
+        It throws a TypeError if _options_ is not undefined and not an Object.
+      </p>
+      <emu-alg>
+        1. If _options_ is *undefined*, then
+          1. Return ! ObjectCreate(*null*).
+        1. If Type(_options_) is Object, then
+          1. Return _options_.
+        1. Throw a *TypeError* exception.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-coerceoptionstoobject" aoid="CoerceOptionsToObject">
+      <h1>CoerceOptionsToObject ( _options_ )</h1>
+      <p>
+        The abstract operation CoerceOptionsToObject coerces _options_ into an Object suitable for use with GetOption, defaulting to an empty Object.
+        Because it coerces non-null primitive values into objects, its use is discouraged for new functionality in favor of GetOptionsObject.
+      </p>
+      <emu-alg>
+        1. If _options_ is *undefined*, then
+          1. Return ! ObjectCreate(*null*).
+        1. Return ? ToObject(_options_).
       </emu-alg>
     </emu-clause>
 
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -340,7 +340,7 @@
         1. If _value_ is *undefined*, return _fallback_.
         1. Assert: _type_ is *"boolean"* or *"string"*.
         1. If _type_ is *"boolean"*, then
-          1. Let _value_ be ToBoolean(_value_).
+          1. Let _value_ be ! ToBoolean(_value_).
         1. If _type_ is *"string"*, then
           1. Let _value_ be ? ToString(_value_).
         1. If _values_ is not *undefined* and _values_ does not contain an element equal to _value_, throw a *RangeError* exception.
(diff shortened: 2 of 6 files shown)


spec/datetimeformat.html
```diff
@@ -168,8 +168,12 @@ <h1>ToDateTimeOptions ( _options_, _required_, _defaults_ )</h1>
       </p>
 
       <emu-alg>
-        1. If _options_ is *undefined*, let _options_ be *null*; otherwise let _options_ be ? ToObject(_options_).
-        1. Let _options_ be ObjectCreate(_options_).
+        1. If _options_ is *undefined*, then
+          1. Let _options_ be ObjectCreate(*null*).
```
ljharb, line 172: wouldn't this want to be `fallback`?

spec/datetimeformat.html
```diff
@@ -168,8 +168,12 @@ <h1>ToDateTimeOptions ( _options_, _required_, _defaults_ )</h1>
...
+        1. Else,
+          1. If Type(_options_) is not Object,
+            1. Throw a *TypeError* exception.
+          1. Let _options_ be ObjectCreate(_options_).
```
ljharb, line 176: i'm kind of confused why ObjectCreate is/was used here - does the spec mutate the options object?

spec/datetimeformat.html
ptomato: I guess that would be shorter!

spec/datetimeformat.html
ptomato: I was trying to preserve the previous behaviour, so I don't know the answer to this; maybe because the subsequent Gets would otherwise be observable?

spec/datetimeformat.html
ljharb: ah, i didn’t expand the diff, i just assumed since the options AO had one

spec/datetimeformat.html
gibson042: ToDateTimeOptions does indeed perform some writes, and the consequences are unnecessarily weird.
```js
const wrappers = new Set();
const options = new Proxy({}, {
  get(target, prop, receiver){
    console.warn("get:", prop, (
      receiver === options ?
        "@options" :
      Object.getPrototypeOf(receiver) === options ?
        "@wrapper" + [...wrappers.add(receiver)].indexOf(receiver) :
      receiver
    ));
    let val = target[prop];
    Reflect.defineProperty(target, prop, {
      enumerable: true,
      get(){ return val },
      set(v){
        console.warn("set:", prop, v, (
          this === options ?
            "@options" :
          Object.getPrototypeOf(this) === options ?
            "@wrapper" + [...wrappers.add(this)].indexOf(this) :
          this
        ));
        val = v;
        return true;
      }
    });
    return val;
  }
});

new Intl.DateTimeFormat(undefined, options);

console.log([...wrappers]);
```

If we're making normative changes, I'd like to remove observable internal objects.

ljharb: In that case, shouldn’t the object be cloned before it’s written to, instead of making an inheriting object? (something that’s observable via proxies, and a very weird thing imo to have be observable)

leobalter:
I understand the goal for consistency, but I'm slightly concern about the potential breaking changes.

If a normative change is proposed where we limit a feature we've had before, I'd like to see Test262 reflecting these in details because we need to understand the possible outcomes for a potential breaking change. With these tests I can also ask implementers if they are good with the proposed change.

Probably the easiest outcome would be a different abstract for Temporal.

ptomato:
I've updated this with the suggestions in #480. It now does not change any observable behaviour.

If desired, subsequent normative changes could be made:
- In InitializeCollator: change to NormalizeOptionsObject.
- In ToDateTimeOptions: get rid of the weird `ObjectCreate(_options_)` behaviour mentioned above, and use either NormalizeOptionsObjectLegacy or NormalizeOptionsObject. (This is now untouched in this PR.)
- In the Intl.Locale constructor: change to NormalizeOptionsObject.
- In SupportedLocales: change to NormalizeOptionsObject.
- In InitializeNumberFormat: change to NormalizeOptionsObject.
- In InitializePluralRules: change to NormalizeOptionsObject.
- In InitializeRelativeTimeFormat: change to NormalizeOptionsObject.

spec/negotiation.html
```diff
@@ -304,6 +301,36 @@ <h1>SupportedLocales ( _availableLocales_, _requestedLocales_, _options_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-normalizeoptionsobject" aoid="NormalizeOptionsObject">
+      <h1>NormalizeOptionsObject ( _options_ )</h1>
```
sffc, line 305: Only in 262, right?

ptomato: I put it in here initially as per https://github.com/tc39/ecma402/issues/480#issuecomment-679420266

spec/negotiation.html
```diff
@@ -304,6 +301,36 @@ <h1>SupportedLocales ( _availableLocales_, _requestedLocales_, _options_ )</h1>
...
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-normalizeoptionsobjectlegacy" aoid="NormalizeOptionsObjectLegacy">
+      <h1>NormalizeOptionsObjectLegacy ( _options_ )</h1>
+      <p>
+        The abstract operation NormalizeOptionsObjectLegacy massages _options_ into an Object to be subsequently passed to GetOption.
+        It only throws an exception if _options_ is null.
+        If _options_ is a primitive, then it makes a best-effort attempt at coercing _options_ to an Object.
+        Newer spec code should use ECMA-262's NormalizeOptionsObject instead.
```
gibson042, line 325:
```suggestion
        Newer spec code should use NormalizeOptionsObject instead.
```

spec/numberformat.html
```diff
@@ -724,7 +721,6 @@ <h1>SetNumberFormatUnitOptions ( _intlObj_, _options_ )</h1>
       </p>
       <emu-alg>
         1. Assert: Type(_intlObj_) is Object.
-        1. Assert: Type(_options_) is Object.
```
gibson042, line 727: Why is this assertion being removed?

spec/locale.html
```diff
@@ -16,7 +16,6 @@ <h1>ApplyOptionsToTag( _tag_, _options_ )</h1>
 
       <emu-alg>
         1. Assert: Type(_tag_) is String.
-        1. Assert: Type(_options_) is Object.
```
gibson042, line 19: Why is this assertion being removed?

spec/numberformat.html
ptomato: I can't find the comment now, but I believe @anba had objected to it after I added it in https://github.com/tc39/ecma402/pull/465, on the grounds of being unnecessary.

spec/locale.html
ptomato: Ditto.

spec/numberformat.html
gibson042: Aren't assertions unnecessary by definition? Their purpose is to clarify invariants, and IMO having a type assertion for one input but not the other seems counterproductive to that end.

gibson042 (commented):
The changes look good, but I think some naming and descriptions could be a little better.

spec/negotiation.html
```diff
@@ -306,6 +303,36 @@ <h1>SupportedLocales ( _availableLocales_, _requestedLocales_, _options_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-normalizeoptionsobject" aoid="NormalizeOptionsObject">
+      <h1>NormalizeOptionsObject ( _options_ )</h1>
+      <p>
+        The abstract operation NormalizeOptionsObject massages _options_ into an Object to be subsequently passed to GetOption.
```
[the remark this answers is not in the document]
leobalter: We should avoid new jargons for the spec text, as extra context for this suggestion change.

leobalter:
I'm in agreement with @gibson042's suggestions. 

@ptomato please let me know if you're interesting in updating this. As those are minor edits, I can do it from here before merging.

Thanks!

ptomato:
Thanks for the reviews. I didn't get to update this before my vacation but I intend to do it sometime this week or next, if that timeline works for you.

leobalter:
No problem, my goal is to have those things done by January to get a good cut for the 2021 Edition.

spec/negotiation.html
```diff
@@ -306,6 +303,36 @@ <h1>SupportedLocales ( _availableLocales_, _requestedLocales_, _options_ )</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-normalizeoptionsobject" aoid="NormalizeOptionsObject">
+      <h1>NormalizeOptionsObject ( _options_ )</h1>
```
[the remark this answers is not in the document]
ptomato: In my opinion "get" as a verb sounds a bit vague, but I don't have any better suggestion, so I've decided to go with this.

ptomato:
Updated.

leobalter (approved):

spec/datetimeformat.html
gibson042: Long-overdue issue: #706

PR #497, opened 2020-08-13
Editorial: Update list of editors and advisor

leobalter:
I've contacted @sffc, @ryzokuken, and @gibson042 privately and got at least one +1 there.

I believe it's ok to have this PR open. The advisor role is uncommon in the spec but I believe it's fine.

### DIFF
--- a/spec/introduction.html
+++ b/spec/introduction.html
@@ -39,7 +39,7 @@
     ECMA-402, 7<sup>th</sup> Edition Project Editors
   </p>
   <p>
-    Leo Balter<br>
-    ECMA-402, 8<sup>th</sup> Edition Project Editor
+    Leo Balter, Richard Gibson<br>
+    ECMA-402, 8<sup>th</sup> Edition Project Editors
   </p>
 </emu-intro>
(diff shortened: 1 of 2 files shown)


sffc:
I discussed this with @ryzokuken and I think "apprentice" would more accurately reflect Ujjwal's role.  An "advisor" would be someone like Daniel Ehrenberg who has been an editor in the past and is staying on to share his advice.

leobalter:
@sffc ok, this apprentice title is even more uncommon, but what should be the best way to solve it? Should we add this role to the document as I've just done with the Advisor role?

sffc:
I have no strong opinion on whether or not to include an apprentice in the spec.

leobalter:
@ryzokuken WDYT? does it work for you? cc @littledan for broader reach out.

ryzokuken:
@leobalter I agree with @sffc that "apprentice" it more appropriate. I don't have a strong opinion on whether it should be included in the spec or not,

ryzokuken:
I talked a bit more to @littledan and it seems that it would be best to leave my name out, that said I do intend to work more closely with the editors and help out as much as I can.

leobalter:
Thanks! I've been thinking about this and I believe an Apprentice entry might be seen as exclusive for other contributors. I agree it might be best to not use it for now until we have a good plan ahead.

leobalter:
I removed the second commit, the first remain untouched.

PR #499, opened 2020-09-03
Editorial: Fix linter errors in spec

ptomato:
Based on a [discussion in Temporal](https://github.com/tc39/proposal-temporal/pull/870), it seems like it might be a good idea to run ecmarkup's linter on the spec, now that in version 4.1.0 has brought several bug fixes in the linter.

This adds the linter to the build process and makes it fail if any warnings are generated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "description": "ECMAScript Internationalization API Specification",
   "scripts": {
     "build-es2020": "git remote remove origin && git remote add origin \"git@github.com:$TRAVIS_REPO_SLUG.git\" && git fetch --quiet origin && git checkout --quiet es2020 && mkdir \"out/2020\" && cp -R img \"out/2020\" && ecmarkup --verbose spec/index.html out/2020/index.html --css out/2020/ecmarkup.css --js out/2020/ecmarkup.js",
-    "build-master": "mkdir out && cp -R img out && ecmarkup --verbose spec/index.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
+    "build-master": "mkdir out && cp -R img out && ecmarkup --lint-spec --verbose spec/index.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
     "prebuild": "npm run clean",
     "build": "npm run build-master",
     "prebuild-for-pdf": "npm run clean",
@@ -13,7 +13,7 @@
     "prebuild-travis": "npm run clean",
     "build-travis": "npm run build-master && npm run build-es2020",
     "clean": "rm -rf out",
-    "test": "exit 0",
+    "test": "ecmarkup --lint-spec spec/index.html /dev/null",
     "prewatch": "npm run clean",
     "watch": "npm run build-master -- --watch"
   },
@@ -22,7 +22,7 @@
   "license": "SEE LICENSE IN https://tc39.es/ecma402/#sec-copyright-and-software-license",
   "homepage": "https://tc39.es/ecma402/",
   "dependencies": {
-    "ecmarkup": "3.19.1"
+    "ecmarkup": "^4.1.0"
   },
   "devDependencies": {
     "@alrra/travis-scripts": "^3.0.0"
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -210,12 +210,12 @@
         1. Assert: Type(_formats_) is List.
         1. For each element _format_ of _formats_ in List order, do
           1. Let _score_ be 0.
-          1. For each _property_ shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, do
+          1. For each property name _property_ shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, do
             1. Let _optionsProp_ be _options_.[[&lt;_property_&gt;]].
             1. Let _formatProp_ be _format_.[[&lt;_property_&gt;]].
             1. If _optionsProp_ is *undefined* and _formatProp_ is not *undefined*, then decrease _score_ by _additionPenalty_.
             1. Else if _optionsProp_ is not *undefined* and _formatProp_ is *undefined*, then decrease _score_ by _removalPenalty_.
-            1. Else if _optionsProp_ ≠ _formatProp_,
+            1. Else if _optionsProp_ ≠ _formatProp_, then
               1. Let _values_ be &laquo; *"2-digit"*, *"numeric"*, *"narrow"*, *"short"*, *"long"* &raquo;.
               1. Let _optionsPropIndex_ be the index of _optionsProp_ within _values_.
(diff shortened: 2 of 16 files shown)


spec/intl.html
```diff
@@ -35,42 +35,42 @@ <h1>Intl[ @@toStringTag ]</h1>
     <h1>Constructor Properties of the Intl Object</h1>
 
     <emu-clause id="sec-intl.locale-intro">
-      <h1>Intl.Locale (...)</h1>
+      <h1>Intl.Locale ( . . . )</h1>
```
ryzokuken, line 38: It in necessary to put the space between the dots too?

spec/locale.html
```diff
@@ -251,6 +250,7 @@ <h1>get Intl.Locale.prototype.baseName</h1>
         1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
         1. Let _locale_ be _loc_.[[Locale]].
         1. Return the substring of _locale_ corresponding to the `unicode_language_id` production.
+      </emu-alg>
```
ryzokuken, line 253: Oops.

ryzokuken (approved):
LGTM, few questions. Thanks!

package.json
```diff
@@ -5,15 +5,15 @@
   "description": "ECMAScript Internationalization API Specification",
   "scripts": {
     "build-es2020": "git remote remove origin && git remote add origin \"git@github.com:$TRAVIS_REPO_SLUG.git\" && git fetch --quiet origin && git checkout --quiet es2020 && mkdir \"out/2020\" && cp -R img \"out/2020\" && ecmarkup --verbose spec/index.html out/2020/index.html --css out/2020/ecmarkup.css --js out/2020/ecmarkup.js",
-    "build-master": "mkdir out && cp -R img out && ecmarkup --verbose spec/index.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
+    "build-master": "mkdir -p out && cp -R img out && ecmarkup --lint-spec --verbose spec/index.html out/index.html --css out/ecmarkup.css --js out/ecmarkup.js",
```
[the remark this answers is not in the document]
ptomato: I had it in a separate commit, but yes it is necessary if you want to run `npm run build-master` twice in a row.

spec/intl.html
ptomato: According to the ecmarkup linter, yes.

spec/datetimeformat.html
```diff
@@ -461,12 +458,12 @@ <h1>UnwrapDateTimeFormat( _dtf_ )</h1>
...
       <emu-normative-optional>
       <emu-alg>
-        2. If _dtf_ does not have an [[InitializedDateTimeFormat]] internal slot and ? InstanceofOperator(_dtf_, %DateTimeFormat%) is *true*, then
+        1. If _dtf_ does not have an [[InitializedDateTimeFormat]] internal slot and ? InstanceofOperator(_dtf_, %DateTimeFormat%) is *true*, then
           1. Let _dtf_ be ? Get(_dtf_, %Intl%.[[FallbackSymbol]]).
       </emu-alg>
       </emu-normative-optional>
       <emu-alg>
-        3. Perform ? RequireInternalSlot(_dtf_, [[InitializedDateTimeFormat]]).
+        1. Perform ? RequireInternalSlot(_dtf_, [[InitializedDateTimeFormat]]).
```
gibson042, line 466: These changes break the rendering of https://tc39.es/ecma402/#sec-unwrapdatetimeformat . We must either keep the non-1 starting values or refactor the algorithms to no longer need them.

spec/datetimeformat.html
```diff
@@ -588,7 +585,7 @@ <h1>Internal slots</h1>
       </ul>
 
       <p>
-        EXAMPLE     An implementation might include the following record as part of its English locale data: {[[hour]]: *"numeric"*, [[minute]]: *"2-digit"*, [[second]]: *"2-digit"*, [[pattern]]: *"{hour}:{minute}:{second}"*, [[pattern12]]: *"{hour}:{minute}:{second} {ampm}"*}.
+        <strong>EXAMPLE</strong> An implementation might include the following record as part of its English locale data: {[[hour]]: *"numeric"*, [[minute]]: *"2-digit"*, [[second]]: *"2-digit"*, [[pattern]]: *"{hour}:{minute}:{second}"*, [[pattern12]]: *"{hour}:{minute}:{second} {ampm}"*}.
```
gibson042, line 588: I think this and similar content would be better refactored with \<emu-note> to match ECMA 262 (e.g., https://tc39.es/ecma262/#sec-runtime-semantics-unicodematchproperty-p ).

spec/intl.html
gibson042: It brings this spec into alignment with ECMA 262 ([example](https://tc39.es/ecma262/#sec-constructor-properties-of-the-global-object)), so I'm in favor.

spec/negotiation.html
```diff
@@ -19,7 +19,9 @@ <h1>Internal slots of Service Constructors</h1>
     </ul>
 
     <p>
-      EXAMPLE     An implementation of DateTimeFormat might include the language tag *"th"* in its [[AvailableLocales]] internal slot, and must (according to <emu-xref href="#sec-intl.datetimeformat-internal-slots"></emu-xref>) include the key *"ca"* in its [[RelevantExtensionKeys]] internal slot. For Thai, the *"buddhist"* calendar is usually the default, but an implementation might also support the calendars *"gregory"*, *"chinese"*, and *"islamicc"* for the locale *"th"*. The [[LocaleData]] internal slot would therefore at least include {[[th]]: {[[ca]]: &laquo; *"buddhist"*, *"gregory"*, *"chinese"*, *"islamicc"* &raquo;}}.
+      <strong>EXAMPLE</strong> An implementation of DateTimeFormat might include the language tag *"th"* in its [[AvailableLocales]] internal slot, and must (according to <emu-xref href="#sec-intl.datetimeformat-internal-slots"></emu-xref>) include the key *"ca"* in its [[RelevantExtensionKeys]] internal slot.
```
gibson042, line 22:
```html
<emu-note>
  <p>For example, an implementation of DateTimeFormat might…</p>
</emu-note>
```

package.json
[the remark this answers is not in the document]
gibson042: I don't have strong feelings about this, but adding `-p` here will suppress the error that would otherwise abort progress when an "out" directory already exists. The [ECMA 262 approach](https://github.com/tc39/ecma262/blob/master/package.json) is probably better, but also too complex for my taste.

gibson042 (changes requested):

package.json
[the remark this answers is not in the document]
ptomato: It was just a fix for convenience, if it has other disadvantages then I'll go ahead and remove it.

gibson042 (approved):

PR #501, opened 2020-09-11
Editorial: Supporting xx-Yyyy-ZZ implies supporting xx-ZZ

### LINKED ISSUE #159: Old-style language tags is too vague
https://tc39.github.io/ecma402/#sec-internal-slots:
> For locales that in current usage would include a script subtag (such as Chinese locales), old-style language tags without script subtags must be included such that, for example, requests for "zh-TW" and "zh-HK" lead to output in traditional Chinese rather than the default simplified Chinese.

This description is pretty vague and doesn't really explain how to identify "old-style language tags".

SpiderMonkey has this fixed list of old-style language tags (http://searchfox.org/mozilla-central/rev/f0e4ae5f8c40ba742214e89aba3f554da0b89a33/js/src/builtin/Intl.js#587-596), but to me it's not really clear if this list is complete or has too many/too few entries.

rkirsling:
Resolves #159.


sffc:
Thank you!

I filed https://github.com/tc39/test262/issues/2762 for the Test262 side.  I think we don't need MDN for this.

rwaldron:
@FrankYFTang @ryzokuken Either of you want to take this?

leobalter (approved):

PR #504, opened 2020-09-23
Change the README.md to include a calendar image which point to the meeting calendar link

FrankYFTang:
GitHub won't let us embed google calendar so I can only create a image of that but make it as a link to the real page.


sffc (approved):

sffc:
There's also a text link to the same page.  But maybe it's not discoverable enough.

FrankYFTang:
> There's also a text link to the same page. But maybe it's not discoverable enough.

yes, I know. That is where I copy the link from. I added the image to make it more "obvious"

gibson042 (approved):

PR #512, opened 2020-11-09
Adding guidelines for proposals in CONTRIBUTING.md

### LINKED ISSUE #442: API surface and scope guidelines
This issue is in a slight relation to the conversation in #435.

I'd like to try to develop an explicit consensus on the scope of ECMA402 target API surface.

Given enough time, and using `ad extremum` reasoning, all possible feature will be requested. Not just 100% of the ICU API surface, but likely more.

How we want as a group to react to that?

Over the last couple years, I definitely saw more cases of arguments along the lines of "ICU already has it, so let's add it" or "CLDR gives it to us, so why not".

Are we ok slowly extending the ECMA402 API surface until it covers 100% of ICU and CLDR? 

Is there a line we don't want to cross? Is there a clear *difference* in target API surface compared to ICU/CLDR that comes from the nature of the Web, JS environment and so on?
For example, we could argue that since JS standard is `forver`, we want to keep the API surface significantly smaller than what a library that can deprecate APIs can do.
JS environment is commonly used in privacy-unfriendly situations (Web), so each API can and will be a target of malicious attempt against the user. Some user interactions can be facilitated via DOM/HTML alleviating the risks (datetime pickers, calendar pickers, range pickers and other UI widgets that require rich interactions). Does it impact our decisions or do we prefer to ensure client-side pickers for all UXes are fully possible via ECMA402 APIs?
Or we could claim that since JavaScript is higher level, certain low-level abstractions are unnecessary.
Or we can make an argument that we intend to keep ECMA402 a significantly simplified compared to ICU due to our target userbase having less expertise than typical ICU users (hence, options bag over skeletons, and reluctance to work with datetime patterns).
We could also argue that the difference is in that we want to minimize the payload to enable implementers in highly constrained environments.
Or we could state that we want to continue ensuring that the specification is not ICU/CLDR bound and thus "ICU has it" is not a good argument at all.

Each one of those, and likely more, of the arguments can be challenged and a counter argument can be made.

When I started contributing to ECMA402, with Caridy and Eric at the helm, I felt this topic was fairly well understood, because we were a very tiny group working for years to add a single function and a single new class (`getCanonicalLocales` and `PluralRules`). It was my understanding that we want to expose 5-10 most common Intl APIs and stop there, leaving as much as possible to user libraries.

I think with the success of ECMA402 comes a shift of perspective, and with the drastic reduction of diversity in implementations comes a simplification of the road from a request filled to Stage 4.

This is, in my opinion, *a good thing* (the success of ECMA402, not the reduction in diversity), but a potentially negative consequence is that it makes it easy to continue on a ICU-compliance treadmill until something makes us stop and revisit.

In the world in which everything we standardize stays as part of the JS platform forever, I'd love us to make an attempt to anticipate that "something" and try to avoid having to learn about it the painful way a flock of WebAPIs went when a high number of peripheral APIs have been proposed (WebBattery, WebBluetooth, WebGPS, WebSensors etc.) and many had to be later [painfully backtracked](https://groups.google.com/forum/#!msg/mozilla.dev.platform/5U8NHoUY-1k/9ybyzQIYCAAJ).

I think we are reaching the original goal of the ECMA402 as of the time when I joined - we have DateTime, Number, PluralRules, RelativeTimeFormat, Locale, ListFormat, DisplayNames, Segmenter, unit formatting (within NumberFormat rev. 2) and MessageFormat in the works.

I believe there's an argument to be made that the scope is sufficient for a JS stdcore library and we can switch gears to polish the APIs we have and cool down on extending the surface.
The argument can *also* be made that there's no reason to stop. We have the momentum, best practices and good velocity and we can keep pumping out new APIs for an increased number of use cases.

I believe the latter is particularly natural reaction, since once developers start using ECMA402 APIs, they'll always find "just one more thing they need" and it's reasonable for them to request an extension of ECMA402, rather than pull a library just for a single feature.

In the light of that, I'd like to start a conversation about:
* the revision of the intended surface of ECMA402
* how is it different from "everything"
* how does the the JS environment affect what we want in
* what we *don't* want to add
* how do we want to evaluate requests for in/out of scope

I hope such guidelines, while likely hard to design, will pay off over the next 2-5 years as the current ECMA402 APIs get more usage and we receive an increased number of requests for many APIs useful to a smaller groups of users.

sffc:
Fixes #442

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,9 +14,25 @@ If the fix is trivial you may not need to sign the CLA. If your fix is involved,
 
 Feature requests for future versions of ECMA402 should be made in this repository by creating a new issue. Your goal will be to convince others that your proposal is a useful addition to the language and recruit TC39 members to help turn your request into a proposal and shepherd it into the language.
 
+See also "ECMA-402 Guidelines for Feature Requests and Proposals" below.
+
 ### Proposals
 
 We follow the TC39 staging process for proposals.  See [ecma262 CONTRIBUTING.md](https://github.com/tc39/ecma262/blob/master/CONTRIBUTING.md#new-feature-proposals) for more information and steps to get started.
+
+See also "ECMA-402 Guidelines for Feature Requests and Proposals" below.
+
+## ECMA-402 Guidelines for Feature Requests and Proposals
+
+The ECMA-402 standard should be as small and generic as possible to supply sufficient internationalization capabilities for JavaScript to lower the barrier to turn JavaScript application for global audience. The ECMA-402 Task Group tries to balance the never-ending stream of additions against the need to remain small, lean and stable for very long (10+ years) timeframe of maintenance.
+
+When the ECMA-402 subcommittee reviews proposals for *new features*, we hold them against the following list of criteria.  It is the reponsibility of the proposal champion to provide a compelling case that their feature request meets these criteria.
+
+1. **Prior Art:** Our job is to bring features that i18n experts have already solved to JavaScript developers, not to invent new solutions to those problems. If the feature is not yet available in CLDR, ICU, or Unicode, the champion should start there before coming to ECMA-402.
+2. **Difficult to implement in userland:** Features in Intl must bring something to the table that a third-party library wouldn't be able to do with the same level of efficiency and performance. The champion can cite a heavy locale data dependency or a complex algorithm to satisfy this criterion.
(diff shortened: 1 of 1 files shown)


longlho (approved):

zbraniecki (approved):

sffc:
Thanks for the reviews!

Does anyone else want to review before I hit the merge button?  Unless I hear otherwise, I will hit the merge button on Thursday or Friday.

PR #531, opened 2021-01-11
Meta: Add templates for issues and PRs

leobalter:
I'd like to open this for discussion.

This ECMA-402 repo has a different aspect compared to 262 as we accept issues with feature suggestions in our issues tracker. I'm adapting the current templates from 262 here and we should eventually add templates for new proposals, ideas, etc.

The PR template has a downstream notice to avoid issues on the spec chaining, I should add ECMA-262 in that list.

I'd like to get an idea of how this group sees the proposed additions. We can merge it if useful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug-report.md
+++ b/.github/ISSUE_TEMPLATE/1-bug-report.md
@@ -1,0 +1,13 @@
+---
+name: "\U0001F41B Bug report"
+about: Report a bug with the specification.
+
+---
+
+<!--
+Thank you for reporting a possible bug in ECMA-402.
+
+If this is a bug with a specific implementation of ECMA-402, such as
+a JavaScript engine or a browser, you might need to report it on their bug
+tracker as well.
+-->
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/1-bug-report.md
```diff
@@ -0,0 +1,13 @@
...
+about: Report a bug with the specification.
+
+---
+
+<!--
+Thank you for reporting a possible bug in ECMA-402.
+
+If this is a bug with a specific implementation of ECMA-402, such as
+a JavaScript engine or a browser, you might need to report it on their bug
+tracker as well.
```
ryzokuken, line 12: Awesome recommendation! Should we also link the different engines' issue trackers? Doesn't seem to be a lot of effort on our end, but it might save the reporting a little time...

ryzokuken (approved):
Looks great to me, just one tiny comment.

ryzokuken:
Ping @leobalter

ryzokuken:
This is a net positive and we probably shouldn't keep it sitting too long on nits. I'll merge it in and then we can do anything extra in follow-on PRs.

PR #539, opened 2021-02-18
Editorial: Fix spec typos and GH Workflow

leobalter:
This patch fixes warnings raised from the spec build

PR #543, opened 2021-02-19
Meta: update actions SHAs to long hashes

### LINKED ISSUE #542: The spec was not build and sync with the tree after 2021-02-11
I notice none of the commit after
Create notes-2021-02-11.md 8488eb7856298dee86d8e2f9de049841805e87f9
got build successfully and sync into https://tc39.es/ecma402/

For example, the following are not in there
spec: GetOption reform for ECMA 2021 (#538) … @ryzokuken  
Normative: Adding Intl.DateTimeFormat formatRange/formatRangeToParts … @fabalbon
Normative: Use OrdinaryHasInstance in normative optional steps (#500) @shvaikalesh
Do not allow duplicate variants within the tlang component of a trans… … @jswalden

@ryzokuken, @fabalbon  @shvaikalesh , @sffc , @jswalden 

It break start on https://github.com/tc39/ecma402/runs/1923390814

ljharb:
Fixes #542

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: ljharb/actions/node/run@125357c
+      - uses: ljharb/actions/node/run@125357c94cc5b34492f39fc8886f2789175d0800
         name: 'npm install && npm run build-master'
         with:
           node-version: lts/*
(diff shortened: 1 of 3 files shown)


FrankYFTang:
It turn out leo is fixing that in https://github.com/tc39/ecma402/pull/539 @ljharb  - could you review that PR instead?

ljharb:
True, but that one's also fixing some spec stuff :-) either PR is fine.

leobalter:
@ljharb Thanks! I prefer the change in my PR using `@main` instead of the whole SHA. Otherwise I'd land your changes here.

leobalter:
Please let me know if we should avoid the current usage or anything. Otherwise I'm closing this for now.

Issue #448, opened 2020-05-27
InitializeDateTimeFormat and BasicFormatMatcher assume that reading absent format record fields returns undefined


anba:
Per [13.3.3 Internal slots](https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots):
> The value of this field must be a list of records, each of which has a subset of the fields shown in Table 6, [...].

An example given there is:
>  {[[hour]]: "numeric", [[minute]]: "2-digit", [[second]]: "2-digit", [[pattern]]: "{hour}:{minute}:{second}", [[pattern12]]: "{hour}:{minute}:{second} {ampm}"}

But [13.1.1 InitializeDateTimeFormat](https://tc39.es/ecma402/#sec-initializedatetimeformat) and [13.1.3 BasicFormatMatcher](https://tc39.es/ecma402/#sec-basicformatmatcher) both assume that reading absent fields returns `undefined`:
> 35. For each row in Table 6, except the header row, in table order, do
> [...]
>     b. Let p be bestFormat.[[\<prop\>]].
>     c. If p not undefined, then

and in `BasicFormatMatcher`:
> ii. Let formatProp be format.[[\<property\>]].
> iii. If optionsProp is undefined and formatProp is not undefined, then decrease score by additionPenalty.
> iv. Else if optionsProp is not undefined and formatProp is undefined, then decrease score by removalPenalty.

The algorithms either need to explicitly check for absent record fields or alternatively irrelevant format record fields must be defined to have the value `undefined` in section 13.3.3.

Issue #441, opened 2020-05-21
Synchronize abstractions from 262


leobalter:
@FrankYFTang has reported the `OrdinaryObjectCreate` replacing an old `ObjectCreate` abstraction. We need to keep this and perhaps other names in sync here.

ljharb:
the old one is `ObjectCreate`. The other two removed between 2019 and 2020 aren't in use in 402.

leobalter:
It was a typo, thanks for flagging that and for the quick recap on what have changed.

PR #546, opened 2021-02-25
Bug fixes for 2021 edition

anba:
Fixes various issues in the current draft edition.


leobalter:
wow! Thank you @anba! This is an excellent work!

sffc:
Wow!  Do you have back references to the issue numbers that this PR fixes?

anba:
I think #441 and #448 are the only already reported issues fixed in this PR. Both times I've used `Fixes #issue` in the commit message, so they should be properly tracked in GitHub.

leobalter (approved):

PR #548, opened 2021-03-01
Editorial: Use `CreateBuiltinFunction` with Abstract Closures

ExE-Boss:
## Depends on:
- [ ] <https://github.com/tc39/ecma262/pull/2109>


ryzokuken:
This has been stalled for a while. @ExE-Boss if you're interested in this still, please let me know and we can try to resurrect it.

PR #557, opened 2021-03-12
Meta: Change "Difficult" to "Expensive" in Stage 2 entrance criterion

sffc:


FrankYFTang (approved):

leobalter (approved):
I'm not blocking the approval on my nit picking. It seems good to land one way or another.

ryzokuken (approved):
LGTM either way (wrt @leobalter's suggestion).

zbraniecki (approved):
`suggestion(optional)`: since we're revisiting it, would you be open to adjust `the same level ...` part?

As of now this sounds like if a proposal is at least on par with the cost of third party, it is passing the bar. I believe our intent here is to say that it has to be significantly more expensive to do X with third-party to pass the bar on extending the standard.

sffc:
I rewrote the whole sentence to read:

> Features in Intl must be significantly more efficient and less complex than a third-party library implementing the same feature.

Does this address your concern @zbraniecki ?

ryzokuken:
ping @zbraniecki

zbraniecki (approved):
lgtm! Thank you for addressing my comment

PR #568, opened 2021-04-15
Editorial: Remove the mention of the "the first substring"

### LINKED ISSUE #559: "the first substring of" and "the first Unicode locale extension sequence" is confusing and misleading
I think the spec text of "first substring" and "the first ... sequence" is confusing and misleading

There won't be a second... but the text mislead the reader to think about how to deal with the "second and rest"

A. In ApplyUnicodeExtensionToTag ( tag, options, relevantExtensionKeys )
 https://tc39.es/ecma402/#sec-apply-unicode-extension-to-tag

>"a. Let extension be the String value consisting of the first substring of tag that is a Unicode locale extension sequence."

The "first substring" is confusing- it imply the possibility of a second. Notice ApplyUnicodeExtensionToTag is only called after
"Set tag to ? ApplyOptionsToTag(tag, options)." in Intl.Locale ( tag [ , options ] ) and ApplyOptionsToTag throws if there are more than one unicode locale extension because "3. If IsStructurallyValidLanguageTag(tag) is false, throw a RangeError exception."
which IsStructurallyValidLanguageTag return true only if "contains at most one unicode_locale_extensions component,"

B. In https://tc39.es/ecma402/#sec-lookupmatcher
> "1. Let extension be the String value consisting of the first substring of locale that is a Unicode locale extension sequence.

also in the note
> "...also contains an [[extension]] field whose value is the first Unicode locale extension sequence within the request locale language tag."

Same as case A above,  LookupMatcher is called by only ResolveLocale and all the locales are already process by  CanonicalizeLocaleList which will throw if there are more than one unicode_locale_extensions component

Why do we need to have the word "first" here? 
I suggest we apply an Editorial PR to change A to

>"a. Let extension be the String value consisting of the<del>first substring of tag that is a Unicode locale extension sequence.</del><ins>substring of Unicode locale extension sequence within tag</ins>"

and change B to
>"1. Let extension be the String value consisting of the <del>first substring of locale that is a Unicode locale extension sequence</del><ins>substring of Unicode locale extension sequence within locale</ins>.

also the note
>"...also contains an [[extension]] field whose value is the <del>first Unicode locale extension sequence within the request locale language tag.</del><ins>substring of Unicode locale extension sequence within the request locale language tag.</ins>"

@leobalter @sffc @zbraniecki @anba  @gibson042 @littledan

FrankYFTang:
Close #559
Remove the misleading usage of the "first" substring which imply it is possible to have second or rest but under condition which are not possible.

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -63,7 +63,7 @@
           1. Let _requestedLocale_ be _requestedLocales_[0].
         1. Else,
           1. Let _requestedLocale_ be DefaultLocale().
-        1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with all Unicode locale extension sequences (<emu-xref href="#sec-unicode-locale-extension-sequences"></emu-xref>) removed.
+        1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with any Unicode locale extension sequences (<emu-xref href="#sec-unicode-locale-extension-sequences"></emu-xref>) removed.
         1. Let _availableLocales_ be a List with language tags that includes the languages for which the Unicode Character Database contains language sensitive case mappings. Implementations may add additional language tags if they support case mapping for additional locales.
         1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
         1. If _locale_ is *undefined*, let _locale_ be *"und"*.
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -57,7 +57,7 @@
         1. Assert: Type(_tag_) is String.
         1. Assert: _tag_ matches the `unicode_locale_id` production.
         1. If _tag_ contains a substring that is a Unicode locale extension sequence, then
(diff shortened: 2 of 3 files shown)


ryzokuken:
@gibson042 @leobalter can either of you give this a final +1?

anba:
Nit: There should be a "the" before each occurrence of "Unicode locale extension sequence" in the modified lines.

---

The spec also contains steps like:
> Let noExtensionsLocale be the String value that is locale with all Unicode locale extension sequences removed.

which also kind of implies that multiple Unicode locale extension sequences are possible within a language tag, even though a structurally valid language tag can contain at most one Unicode locale extension sequence. Should this also be changed, for example by switching from "all" to "any"?

> Let noExtensionsLocale be the String value that is locale with any Unicode locale extension sequence removed.

FrankYFTang:
@anba  @ryzokuken  PTAL

ryzokuken (approved):
LGTM, @anba @gibson042 @leobalter please approve.

anba (approved):

leobalter (approved):

ryzokuken:
Merging editorial PR after multiple positive reviews and two editor signoffs.

PR #581, opened 2021-06-09
Editorial: mention "cf" in Note 1 of #sec-intl.numberformat-internal-slots

FrankYFTang:
Mention "cf" in https://tc39.es/ecma402/#sec-intl.numberformat-internal-slots
see also https://unicode.org/reports/tr35/#Key_Type_Definitions Make it clear that "cf" is not supported via extension.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -961,7 +961,7 @@
       </p>
 
       <emu-note>
-        Unicode Technical Standard 35 describes two locale extension keys that are relevant to number formatting: *"cu"* for currency and *"nu"* for numbering system. Intl.NumberFormat, however, requires that the currency of a currency format is specified through the currency property in the options objects.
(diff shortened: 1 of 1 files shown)


sffc (approved):
Seems good; @ryzokuken ?

ryzokuken (approved):
@sffc it does, but it made me think: `cf` and `hc` are two pieces of information that have been handled in the API using options, so maybe this list would get longer over time as the BCP 47 standardization process moves ahead. Maybe we need to make a note to add support for these new tags over time and even effectively deprecate the use of the options as we go about it (I'm not opinionated about the latter, but we should definitely talk about it, right?)

sffc:
2021-07-01 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2021-07-01.md#editorial-mention-cf-in-note-1-of-sec-intlnumberformat-internal-slots-581

Follow-up for editors: #584

ryzokuken:
Merging this since the relevant follow-up issues have been opened. Thanks @FrankYFTang!

PR #587, opened 2021-07-02
Editorial: Clarify the "All registered Zone and Link names"

FrankYFTang:
Clarify that "All registered Zone and Link names" is implementation dependent in a sense that a ealier shipped browser may not have the latest IANA timezone DB in practice.
 (in ECMA262 language - implementation-approximated ??)


ryzokuken (approved):

ryzokuken:
@gibson042 @leobalter @sffc what do you folks think?

gibson042 (approved):

ryzokuken:
Thanks for the review, @gibson042. Merging.

PR #592, opened 2021-07-14
Editorial: normalize property name iteration

jugglinmike:
Prior to this commit, the variable named `prop` was not defined in two
of the loops within ToDateTimeOptions. Define it using the convention
for loops in ECMA262, and update the algorithm's similar loops to match.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -202,13 +202,11 @@
         1. Let _options_ be OrdinaryObjectCreate(_options_).
         1. Let _needDefaults_ be *true*.
         1. If _required_ is *"date"* or *"any"*, then
-          1. For each of the property names *"weekday"*, *"year"*, *"month"*, *"day"*, do
-            1. Let _prop_ be the property name.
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):
Looks good overall. I'm wondering here that since there's nothing quite special about being a "property name" and since we're just iterating a constant list, it should instead be:

`For each element _prop_ of &laquo; ... &raquo;` to simply things.

jugglinmike:
The wording in this patch is based on a similar step in [the BasicFormatMatcher algorithm](https://tc39.es/ecma402/#sec-basicformatmatcher). There, the qualifier is necessary to disambiguate between other columns in the referenced table. That's not a concern here, but it could still be useful as a way to further contextualize the purpose of the list elements. That said, I'm happy to change this according to the editors' preference.

ryzokuken:
@jugglinmike you're right. Besides, either way, this is a net positive with respect to the status quo so let's merge this and discuss any further changes separately.

Thanks for the PR again! :innocent:

PR #594, opened 2021-07-15
Editorial: Define a structured signature for PluralRuleSelect

### LINKED ISSUE #593: Be more clear on valid return values for PluralRuleSelect
Reported by @codehag in https://github.com/tc39/proposal-intl-numberformat-v3/issues/57

The abstract operation PluralRuleSelect does not clearly specify the set of valid return values, and it's not clear what the four arguments signify.  This should be cleaned up with an editorial PR against ECMA-402.  The PluralRuleSelectRange abstract operation in NFv3 should also be adjusted.

gibson042:
Fixes #593


spec/datetimeformat.html
```diff
@@ -374,7 +374,7 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, _patternParts_, _x_, _rangeFormatO
           1. Else if _p_ is equal to *"timeZoneName"*, then
             1. Let _f_ be _dateTimeFormat_.[[TimeZoneName]].
             1. Let _v_ be _dateTimeFormat_.[[TimeZone]].
-            1. Let _fv_ be a String value representing _v_ in the form given by _f_; the String value depends upon the implementation and the effective locale. The String value may also depend on the value of the [[InDST]] field of _tm_.  If the implementation does not have a localized representation of _f_, then use the String value of _v_ itself.
+            1. Let _fv_ be a String value representing _v_ in the form given by _f_; the String value depends upon the implementation and the effective locale. The String value may also depend on the value of the [[InDST]] field of _tm_. If the implementation does not have a localized representation of _f_, then use the String value of _v_ itself.
```
gibson042, line 377: This removal of consecutive whitespace resolves a warning from ecmarkup.

spec/pluralrules.html
```diff
@@ -102,12 +102,19 @@ <h1>GetOperands ( _s_ )</h1>
...
+    <emu-clause id="sec-pluralruleselect" type="implementation-defined abstract operation">
+      <h1>
+        PluralRuleSelect (
+          _locale_: a String,
+          _type_: a String,
+          _n_: a finite Number,
+          _operands_: a Plural Rules Operands Record corresponding with _n_,
+        )
+      </h1>
+      <dl class="header">
```
gibson042, line 114: Further adoption of structured operation headers can be accomplished later.

spec/pluralrules.html
```diff
@@ -102,12 +102,19 @@ <h1>GetOperands ( _s_ )</h1>
...
+        <dt>description</dt>
+        <dd>It returns the String from &laquo; *"zero"*, *"one"*, *"two"*, *"few"*, *"many"*, *"other"* &raquo; that best categorizes _n_ according to the rules for _locale_ and _type_.</dd>
```
gibson042, line 116: Note that _operands_ doesn't actually seem relevant to the signature of PluralRuleSelect... I'd like to take this PR even further, removing _locale_, _type_, and _operands_ and replacing them with a single _pluralRules_ argument (an instance with an [[InitializedPluralRules]] slot) for better comprehensibility and alignment with other parts of the spec, in the process also removing GetOperands as an implementation detail (since _its_ only use is to construct _operands_).

sffc (approved):
Thanks! Once this is merged I'll pull it in to NFv3.

spec/pluralrules.html
```diff
@@ -102,12 +102,19 @@ <h1>GetOperands ( _s_ )</h1>
        </emu-table>
     </emu-clause>
 
-    <emu-clause id="sec-pluralruleselect" aoid="PluralRuleSelect">
-      <h1>PluralRuleSelect ( _locale_, _type_, _n_, _operands_ )</h1>
-
-      <p>
-        When the PluralRuleSelect abstract operation is called with four arguments, it performs an implementation-dependent algorithm to map _n_ to the appropriate plural representation of the Plural Rules Operands Record _operands_ by selecting the rules denoted by _type_ for the corresponding _locale_, or the String value *"other"*.
-      </p>
+    <emu-clause id="sec-pluralruleselect" type="implementation-defined abstract operation">
```
ryzokuken, line 105: The `aoid` is gone, I suppose automagic linking of this AO from algorithms won't work anymore.

spec/pluralrules.html
ryzokuken: I assume this is a new convention from ECMA-262? It looks great, thanks!

ryzokuken (approved):
LGTM. Not sure if the follow-on change you mentioned will keep this PR editorial or not... @sffc @zbraniecki what do you think?

spec/pluralrules.html
gibson042: `aoid` is now pulled from the structured header, cf. https://github.com/tc39/ecmarkup/blob/6253fe32dec8a8b7f4271e76f734fbab2ee4c5f4/src/Clause.ts#L122-L143

ryzokuken: oh wow, this is great. I should really keep up with all the exciting new stuff in ecmarkup

gibson042:
> Not sure if the follow-on change you mentioned [removing _locale_, _type_, and _operands_ and replacing them with a single _pluralRules_ argument (an instance with an [[InitializedPluralRules]] slot) for better comprehensibility and alignment with other parts of the spec, in the process also removing GetOperands as an implementation detail] will keep this PR editorial or not... @sffc @zbraniecki what do you think?

I believe it would because [PluralRuleSelect](https://tc39.es/ecma402/#sec-pluralruleselect) is called **only** from [ResolvePlural](https://tc39.es/ecma402/#sec-resolveplural), which pulls _locale_ and _type_ directly from slots on _pluralRules_ and determines _operands_ by invoking [GetOperands](https://tc39.es/ecma402/#sec-getoperands) on a value from the response of [FormatNumericToString](https://tc39.es/ecma402/#sec-formatnumberstring)(_pluralRules_, _n_). So if [PluralRuleSelect](https://tc39.es/ecma402/#sec-pluralruleselect) were provided with _pluralRules_ and _n_ similar to e.g. [FormatDateTime](https://tc39.es/ecma402/#sec-formatdatetime)(_dateTimeFormat_, _x_) and [FormatNumeric](https://tc39.es/ecma402/#sec-formatnumber)(_numberFormat_, _x_), then it could do those steps itself (reading [[Locale]] and [[Type]] and determining how they affect the result of "zero" vs. "one" vs. "two" vs. "few" vs. "many" vs. "other").

Unless there's some reason why plural rule selection MUST format the number to a numeric string and then use the resulting integer/fraction digit counts to make its determination, even though that determination is itself implementation-defined?

anba:
The number is formatted as a string, because [plural rules](https://github.com/unicode-org/cldr/blob/master/common/supplemental/plurals.xml) make a difference between integers and non-integers:

```js
js> new Intl.PluralRules("en", {minimumFractionDigits: 0}).select(1) 
"one"
js> new Intl.PluralRules("en", {minimumFractionDigits: 2}).select(1) 
"other"
```

And by specifying that `FormatNumericToString` and `GetOperands` must be used, we can control how the input number is formatted, which gives less wiggle room for an implementation to do the wrong thing.

gibson042:
I just don't see how that's the case... non-normative references to CLDR and in-spec examples would help _much_ more than requiring a roundabout trip from Number to String to Record—especially when the Record appears to be an attempt to recover integer/fraction/significant digits configuration that already exists on _pluralRules_. It just seems overly prescriptive.

spec/pluralrules.html
```diff
@@ -79,35 +75,42 @@ <h1>GetOperands ( _s_ )</h1>
              <td>Number of digits of [[Number]].</td>
```
gibson042, line 75: @sffc there are problems with the definitions of these fields:
* [[IntegerDigits]] is defined as "Number of digits of [[Number]]", but is populated with the _number value_ of the integer digits (e.g., it is equal to 8 for an input string like "8.540").
* [[FractionDigits]] and [[NumberOfFractionDigits]] have effectively the same definition, even though the former is populated with the number value of the fractional digits while the latter is populated with the _count_ of such digits (e.g., they are respectively equal to 540 and 3 for an input string like "8.540").
* [[FractionDigitsWithoutTrailing]] and [[NumberOfFractionDigitsWithoutTrailing]] also have effectively the same definition, even though the former is populated with the number value of the truncated fractional digits while the latter is populated with the _count_ of such digits (e.g., they are respectively equal to 54 and 2 for an input string like "8.540").

I'd like to fix this... do you know what the actual intent is?

gibson042:
> Maybe what's putting me off is that PluralRuleSelect has access to _n_ in addition to the Record derived from inspecting its string serialization subject to _pluralRules_, but does not have access to _pluralRules_ itself... it seems so strange to provide both _n_ **and** _operands_.

Concrete proposal: drop GetOperands and change the signature of PluralRuleSelect to accept a String ([[FormattedString]] output from FormatNumericToString) rather than a Number and a Record. The string parsing of GetOperands would become an implementation detail.
```patch
         1. Let _res_ be ! FormatNumericToString(_pluralRules_, _n_).
-        1. Let _s_ be _res_.[[FormattedString]].
-        1. Let _operands_ be ! GetOperands(_s_).
-        1. Return ! PluralRuleSelect(_locale_, _type_, _n_, _operands_).
+        1. Return ! PluralRuleSelect(_locale_, _type_, _res_.[[FormattedString]]).
```

spec/pluralrules.html
sffc: I think these are supposed to correspond to: https://unicode.org/reports/tr35/tr35-numbers.html#Operands

- i | integer digits of n.
- v | number of visible fraction digits in n, with trailing zeros.*
- w | number of visible fraction digits in n, without trailing zeros.*
- f | visible fraction digits in n, with trailing zeros.*
- t | visible fraction digits in n, without trailing zeros.*

gibson042: Thanks. I added references and updated the text to match (which is still just editorial work).

spec/pluralrules.html
```diff
@@ -27,35 +27,33 @@ <h1>InitializePluralRules ( _pluralRules_, _locales_, _options_ )</h1>
...
-           1. Let _w_ be the length of _ft_.
-           1. Let _t_ be ! ToNumber(_ft_).
+           1. Let _intPart_ be _n_.
+           1. Let _fracSlice_ be *""*.
          1. Else,
-           1. Let _w_ be 0.
-           1. Let _t_ be 0.
-         1. Return a new Record { [[Number]]: _n_, [[IntegerDigits]]: _i_, [[NumberOfFractionDigits]]: _v_, [[NumberOfFractionDigitsWithoutTrailing]]: _w_, [[FractionDigits]]: _f_, [[FractionDigitsWithoutTrailing]]: _t_ }.
+           1. Let _intPart_ be the substring of _s_ from position 0, inclusive, to position _dp_, exclusive.
+           1. Let _fracSlice_ be the substring of _s_ from position _dp_, exclusive, to the end of _s_.
```
gibson042, line 49:
```suggestion
           1. Let _intPart_ be the substring of _s_ from 0 to _dp_.
           1. Let _fracSlice_ be the substring of _s_ from _dp_ + 1.
```

PR #603, opened 2021-08-21
Editorial: Normalize internal slot access

Fayti1703:
This PR changes several instances of "[[Slot]][[&lt;_variable_&gt;]]" and "[[Slot]][&lt;_variable_&gt;]" to the proper "[[Slot]].[[&lt;_variable_&gt;]]" notation.

Thanks to @hackmud-dtr for noticing these.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -135,9 +135,9 @@
       </p>
 
       <ul>
-        <li>The first element of [[SortLocaleData]][[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]][[&lt;_locale_&gt;]].[[co]] must be *null*.</li>
+        <li>The first element of [[SortLocaleData]].[[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]].[[&lt;_locale_&gt;]].[[co]] must be *null*.</li>
         <li>The values *"standard"* and *"search"* must not be used as elements in any [[SortLocaleData]].[[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]].[[&lt;_locale_&gt;]].[[co]] list.</li>
-        <li>[[SearchLocaleData]][[&lt;_locale_&gt;]] must have a [[sensitivity]] field with a String value equal to *"base"*, *"accent"*, *"case"*, or *"variant"*.</li>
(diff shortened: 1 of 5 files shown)


gibson042 (approved):

spec/collator.html
```diff
@@ -135,9 +135,9 @@ <h1>Internal Slots</h1>
       </p>
 
       <ul>
-        <li>The first element of [[SortLocaleData]][[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]][[&lt;_locale_&gt;]].[[co]] must be *null*.</li>
+        <li>The first element of [[SortLocaleData]].[[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]].[[&lt;_locale_&gt;]].[[co]] must be *null*.</li>
```
ljharb, line 138: this was really confusing to read at first (either form) because it's implying that `[[SortLocaleData]]` is a first-class thing. Perhaps it would be clearer if it said something like this:
```suggestion
        <li>The first element in any object `_obj_` of _obj_.[[SortLocaleData]].[[&lt;_locale_&gt;]].[[co]] and [[SearchLocaleData]].[[&lt;_locale_&gt;]].[[co]] must be *null*.</li>
```
?

Fayti1703: That would need to be %Collator%.[[SortLocaleData]], not _obj_.[[SortLocaleData]]. (Especially because that uses _obj_ in the definition of what _obj_ refers to)

ljharb: sure, if there's a specific object here then that's the obvious choice :-) it's just very unclear to me to have an internal slot used like this.

ryzokuken (approved):
LGTM. I think @ljharb's comments are important but not blocking since they apply with or without this PR (and to place outside of this diff as well). 

Jordan do you agree? Can I merge this as-is?

ljharb:
Yes, since it’s confusing in either form, it’s fine to merge this as-is; but it’d be great to clean it up ASAP :-)

ryzokuken:
@ljharb agreed. Would you like to file an issue for it or should I? I plan to address a bunch of editorial issues soon so that would put it on my personal TODO list.

ljharb:
Filed as #612.

Issue #422, opened 2020-03-31
Change Array.prototype.toLocaleString to use ListFormat


FrankYFTang:
Currently, there is a section in ECMA402 about Array.prototype.toLocaleString
https://ecma-international.org/ecma-402/#sup-array.prototype.tolocalestring

With the newly Stage 3 Intl.ListFormat moving to Stage 4, we should consider to rewrite that section to delegate the behavior to Intl.ListFormat

anba:
Two potential issues which can make the transition to `Intl.ListFormat` a bit harder than expected:
1. How to handle the `options` parameter? Should it apply only to the individual array elements or also to the implicitly constructed `ListFormat`?
    - I'll guess I can answer this myself: I don't think we can reuse the `options` parameter to apply to `Intl.ListFormat` object, because some property names like `"style"` can have different meanings for the `ListFormat` and for the elements, cf. `[1].toLocaleString("en", {style: "percent"})`.
1. The current `Array.prototype.toLocaleString` specification ensures all elements are always separated with the same delimiter. This won't be the case anymore when `ListFormat` is used. Here's an example using English, German, and French which shows that there's no `type`/`style` combination which ensures only a single delimiter is used between elements for all locales.
```js
for (let locale of ["en", "de", "fr"]) {
  for (let type of ["conjunction", "unit"]) {
    for (let style of ["long", "short", "narrow"]) {
      print(`${locale} (${type}-${style}):`, new Intl.ListFormat(locale, {type, style}).format(["1", "2"]))
    }
  }
}
```
---

3. ChakraCore uses [LOCALE_SLIST](https://docs.microsoft.com/en-us/windows/win32/intl/locale-slist), which can give better output for certain locales and element types (or more specifically: numbers). `LOCALE_SLIST` is similar to [`numbers/symbols/list` in CLDR](https://unicode.org/reports/tr35/tr35-numbers.html#Number_Symbols). 
    - This applies to locales where the list separator is the same character as the decimal separator. 
    - Also see: <https://bugzilla.mozilla.org/show_bug.cgi?id=1388777> and <https://bugs.chromium.org/p/chromium/issues/detail?id=753841>
    - (But I guess if we want to support number list formatting, it should happen in a follow-up proposal to `Intl.ListFormat`.)

sffc:
> How to handle the options parameter?

This is a great question that I don't have an immediate answer to.  One possible solution: namespace the child options like so:

```javascript
[1].toLocaleString({
[...]

PR #615, opened 2021-10-07
Normative: Use ListFormat for Array.prototype.toLocaleString

FrankYFTang:
This PR address https://github.com/tc39/ecma402/issues/422

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -238,20 +238,18 @@
       </p>
 
       <emu-alg>
+        1. Let _lf_ be ? Construct(%ListFormat%, &laquo; _locales_, _options_ &raquo;).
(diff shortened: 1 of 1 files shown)


anba (changes requested):
The `options` parameter conflict described in #422 is still unresolved.

ryzokuken (approved):
+1, I agree with @zbraniecki

FrankYFTang:
We discussed this in the 2021-11-04 TG2 and I feel there are still other concerns which need to be addressed. But I don't think I could address them myself. I would leave this to whoever want to take it from here to work on that.

sffc:
See the issue #422 for additional discussion

sffc:
2022-05-19: This issue is too large for a PR; it should have a proposal. See #422. Volunteers would be appreciated.

PR #618, opened 2021-10-25
Editorial: Remove unnecessary specification of supportedLocalesOf `length` properties

gibson042:
Some but not all `supportedLocalesOf` definitions explicitly specify `length`, but none of them should need to. However, we might consider updating https://tc39.es/ecma402/#conventions to explicitly adopt https://tc39.es/ecma262/#sec-ecmascript-standard-built-in-objects
> Every built-in <a href="https://tc39.es/ecma262/#function-object">function object</a>, including <a href="https://tc39.es/ecma262/#constructor">constructors</a>, has a **"length"** property whose value is a non-negative <a href="https://tc39.es/ecma262/#integral-number">integral Number</a>. Unless otherwise specified, this value is equal to the number of required parameters shown in the subclause heading for the function description. Optional parameters and rest parameters are not included in the parameter count.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -113,10 +113,6 @@
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Return ? SupportedLocales(_availableLocales_, _requestedLocales_, _options_).
       </emu-alg>
-
-      <p>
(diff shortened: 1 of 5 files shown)


ryzokuken:
Looks like everything is in order, merging!

Thanks @gibson042 @anba and @ljharb

PR #622, opened 2021-11-04
Normative: add intl-displaynames-v2

FrankYFTang:
Based on https://tc39.es/intl-displaynames-v2/

pending on TC39 2021-Dec decision

### DIFF
--- a/spec/displaynames.html
+++ b/spec/displaynames.html
@@ -222,8 +222,8 @@
         <li>The display name style fields under display name type *"region"* should contain Records with keys corresponding to region codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"script"* should contain Records with keys corresponding to script codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"currency"* should contain Records with keys corresponding to currency codes. The value of these fields must be string values.</li>
-        <li>The display name style fields under display name type *"calendar"* should contain Records, with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
-        <li>The display name style fields under display name type *"dateTimeField"* should contain Records, with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
+        <li>The display name style fields under display name type *"calendar"* should contain Records with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
+        <li>The display name style fields under display name type *"dateTimeField"* should contain Records with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
       </ul>
 
       <emu-note>
--- a/spec/displaynames.html
+++ b/spec/displaynames.html
@@ -222,8 +222,8 @@
         <li>The display name style fields under display name type *"region"* should contain Records with keys corresponding to region codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"script"* should contain Records with keys corresponding to script codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"currency"* should contain Records with keys corresponding to currency codes. The value of these fields must be string values.</li>
-        <li>The display name styles fields under display name type *"calendar"* should contain Records, with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
-        <li>The display name styles fields under display name type *"dateTimeField"* should contain Records, with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
+        <li>The display name style fields under display name type *"calendar"* should contain Records, with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
+        <li>The display name style fields under display name type *"dateTimeField"* should contain Records, with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
       </ul>
 
       <emu-note>
--- a/spec/displaynames.html
+++ b/spec/displaynames.html
@@ -38,11 +38,16 @@
     </emu-clause>
   </emu-clause>
 
-    <emu-clause id="sec-isvaliddatetimefieldcode" aoid="IsValidDateTimeFieldCode">
-      <h1>IsValidDateTimeFieldCode ( _field_ )</h1>
-      <p>
-        The IsValidDateTimeFieldCode abstract operation is called with argument _field_. It verifies that the _field_ argument represents a valid date time field code. The following steps are taken:
-      </p>
+    <emu-clause id="sec-isvaliddatetimefieldcode" type="abstract operation">
+      <h1>
+        IsValidDateTimeFieldCode (
+          _field_: a String,
+        )
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It verifies that the _field_ argument represents a valid date time field code.</dd>
+      </dl>
       <emu-alg>
         1. If _field_ is listed in the Code column of <emu-xref href="#table-validcodefordatetimefield"></emu-xref>, return *true*.
         1. Return *false*.
@@ -107,7 +112,6 @@
           </tr>
         </table>
       </emu-table>
-
     </emu-clause>
 
   <emu-clause id="sec-intl-displaynames-constructor">
--- a/spec/displaynames.html
+++ b/spec/displaynames.html
@@ -92,7 +92,7 @@
           </tr>
           <tr>
             <td>*"dayPeriod"*</td>
-            <td>The field indicating the day period, either am/pm marker or others, e.g. noon, evening.</td>
+            <td>The field indicating the day period, either am, pm, etc. or noon, evening, etc..</td>
           </tr>
           <tr>
             <td>*"hour"*</td>
--- a/spec/displaynames.html
+++ b/spec/displaynames.html
@@ -210,9 +210,9 @@
       </p>
 
       <ul>
-        <li>[[LocaleData]].[[&lt;_locale_&gt;]] must have a [[types]] field for all locale values _locale_. The value of this field must be a Record, which must have fields with the names of all display name types: *"language"*, *"region"*, *"script"*, *"currency"*, `"calendar"`, and `"dateTimeField"`.</li>
-        <li>The value of the field `"language"` must be a Record which must have fields with the names of one of the valid language displays: *"dialect"* and *"standard"*.</li>
-        <li>The language display fields under display name type `"language"` should contain Records which must have fields with the names of one of the valid display name styles: *"narrow"*, *"short"*, and *"long"*.</li>
+        <li>[[LocaleData]].[[&lt;_locale_&gt;]] must have a [[types]] field for all locale values _locale_. The value of this field must be a Record, which must have fields with the names of all display name types: *"language"*, *"region"*, *"script"*, *"currency"*, *"calendar"*, and *"dateTimeField"*.</li>
+        <li>The value of the field *"language"* must be a Record which must have fields with the names of one of the valid language displays: *"dialect"* and *"standard"*.</li>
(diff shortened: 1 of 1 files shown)


spec/displaynames.html
```diff
@@ -126,12 +210,16 @@ <h1>Internal slots</h1>
...
+        <li>[[LocaleData]].[[&lt;_locale_&gt;]] must have a [[types]] field for all locale values _locale_. The value of this field must be a Record, which must have fields with the names of all display name types: *"language"*, *"region"*, *"script"*, *"currency"*, `"calendar"`, and `"dateTimeField"`.</li>
+        <li>The value of the field `"language"` must be a Record which must have fields with the names of one of the valid language displays: `"dialect"` and `"standard"`.</li>
+        <li>The language display fields under display name type `"language"` should contain Records which must have fields with the names of one of the valid display name styles: `"narrow"`, `"short"`, and `"long"`.</li>
+        <li>The value of the fields *"region"*, *"script"*, *"currency"*, `"calendar"`, and `"dateTimeField"` must be Records, which must have fields with the names of all display name styles: *"narrow"*, *"short"*, and *"long"*.</li>
         <li>The display name style fields under display name type *"language"* should contain Records with keys corresponding to language codes matching the `unicode_language_id` production. The value of these fields must be string values.</li>
```
ryzokuken, line 217: I think this line needs to be reworded a bit? Right now it seems conflicting with the line https://github.com/tc39/ecma402/pull/622/files#diff-e523319a519e9b301a04564d9f821c6e35720c21b603d77b5c0d982e1bfb5c53R215.

spec/displaynames.html
```diff
@@ -126,12 +210,16 @@ <h1>Internal slots</h1>
...
         <li>The display name style fields under display name type *"region"* should contain Records with keys corresponding to region codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"script"* should contain Records with keys corresponding to script codes. The value of these fields must be string values.</li>
         <li>The display name style fields under display name type *"currency"* should contain Records with keys corresponding to currency codes. The value of these fields must be string values.</li>
+        <li>The display name styles fields under display name type `"calendar"` should contain Records, with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
```
ryzokuken, line 221:
```suggestion
        <li>The display name style fields under display name type `"calendar"` should contain Records with keys corresponding to a String value with the `type` given in Unicode Technical Standard 35 for the calendar used for formatting. The value of these fields must be string values.</li>
```

ryzokuken: It seems like you say "string" twice here, so maybe that could be fixed?

spec/displaynames.html
```diff
@@ -126,12 +210,16 @@ <h1>Internal slots</h1>
...
+        <li>The display name styles fields under display name type `"dateTimeField"` should contain Records, with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
```
ryzokuken, line 222:
```suggestion
        <li>The display name style fields under display name type `"dateTimeField"` should contain Records with keys corresponding to codes listed in <emu-xref href="#table-validcodefordatetimefield"></emu-xref>. The value of these fields must be string values.</li>
```

ryzokuken (changes requested):
LGTM overall, please check out my feedback inline.

ryzokuken:
One more thing: since this is a new addition, should this use the new structured argument syntax for new AOs? Otherwise it would increase the amount of editorial work.

spec/displaynames.html
```diff
@@ -58,7 +137,7 @@ <h1>Intl.DisplayNames ( _locales_, _options_ )</h1>
         1. Let _r_ be ResolveLocale(%DisplayNames%.[[AvailableLocales]], _requestedLocales_, _opt_, %DisplayNames%.[[RelevantExtensionKeys]]).
         1. Let _style_ be ? GetOption(_options_, *"style"*, *"string"*, &laquo; *"narrow"*, *"short"*, *"long"* &raquo;, *"long"*).
         1. Set _displayNames_.[[Style]] to _style_.
-        1. Let _type_ be ? GetOption(_options_, *"type"*, *"string"*, &laquo; *"language"*, *"region"*, *"script"*, *"currency"* &raquo;, *undefined*).
+        1. Let _type_ be ? GetOption(_options_, *"type"*, *"string"*, &laquo; *"language"*, *"region"*, *"script"*, *"currency"*, `"calendar"`, `"dateTimeField"` &raquo;, *undefined*).
```
ljharb, line 140: Why are most of the strings here using asterisks, but the new ones using backticks?

spec/displaynames.html
```diff
@@ -252,9 +344,10 @@ <h1>Properties of Intl.DisplayNames Instances</h1>
     <ul>
       <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
       <li>[[Style]] is one of the String values *"narrow"*, *"short"*, or *"long"*, identifying the display name style used.</li>
-      <li>[[Type]] is one of the String values *"language"*, *"region"*, *"script"*, or *"currency"*, identifying the type of the display names requested.</li>
+      <li>[[Type]] is one of the String values *"language"*, *"region"*, *"script"*, *"currency"*, `"calendar"`, or `"dateTimeField"`, identifying the type of the display names requested.</li>
```
ljharb, line 347: here also

spec/displaynames.html
```diff
@@ -31,6 +38,78 @@ <h1>CanonicalCodeForDisplayNames ( _type_, _code_ )</h1>
...
+        <table class="real-table">
+          <thead>
+            <tr>
+              <th>Code</th>
+              <th>Description</th>
+            </tr>
+          </thead>
+          <tr>
+            <td>`"era"`</td>
+            <td>The field indicating the era, e.g. AD or BC in the Gregorian (Julian) calendar.</td>
```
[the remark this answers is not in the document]
FrankYFTang: But the "AD or BC" are exactly the same in both calendar. The day or month in these two calendar could be different, but the era are always the same- spell the same and no date will result different era in these two calendar.

spec/displaynames.html
```diff
@@ -68,6 +147,11 @@ <h1>Intl.DisplayNames ( _locales_, _options_ )</h1>
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
         1. Let _types_ be _dataLocaleData_.[[types]].
         1. Assert: _types_ is a Record (see <emu-xref href="#sec-Intl.DisplayNames-internal-slots"></emu-xref>).
+        1. Let _languageDisplay_ be ? GetOption(_options_, `"languageDisplay"`, `"string"`, &laquo; `"dialect"`, `"standard"` &raquo;, `"dialect"`).
+        1. If _type_ is `"language"`, then
+          1. Set _displayNames_.[[LanguageDisplay]] to _languageDisplay_.
+          1. Let _typeFields_ be _typeFields_.[[&lt;_languageDisplay_&gt;]].
+          1. Assert: _typeFields_ is a Record (see <emu-xref href="#sec-Intl.DisplayNames-internal-slots"></emu-xref>).
```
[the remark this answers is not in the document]
FrankYFTang: good catch, I insert the additon two lines too early. Shifted now

spec/displaynames.html
FrankYFTang: sorry, I think you put in the wrong link. I cannot see what you try to say about the conflict

spec/displaynames.html
FrankYFTang: The "styles fields" were referring to those fields inside typeFields "26. Let styleFields be typeFields.[[<style>]].". If you look at the spec text before, I wrote the same way. If we want to change it, I think we should make an editorial changes on ECMA402 on the preexisting one separated from here.

FrankYFTang:
> One more thing: since this is a new addition, should this use the new structured argument syntax for new AOs? Otherwise it would increase the amount of editorial work.

you are talking about IsValidDateTimeFieldCode right? I will change it accordingly later today.

spec/displaynames.html
FrankYFTang: DONE

FrankYFTang:
> One more thing: since this is a new addition, should this use the new structured argument syntax for new AOs? Otherwise it would increase the amount of editorial work.

DONE

FrankYFTang:
PTAL

spec/displaynames.html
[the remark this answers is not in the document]
gibson042:
```suggestion
            <td>The field indicating the era, e.g. AD or BC in the Gregorian or Julian calendar.</td>
```

spec/displaynames.html
```diff
@@ -31,6 +38,82 @@ <h1>CanonicalCodeForDisplayNames ( _type_, _code_ )</h1>
...
+              <th>Description</th>
+            </tr>
+          </thead>
+          <tr>
+            <td>*"era"*</td>
+            <td>The field indicating the era, e.g. AD or BC in the Gregorian (Julian) calendar.</td>
+          </tr>
+          <tr>
+            <td>*"year"*</td>
+            <td>The field indicating the year.</td>
```
gibson042, line 71: Should this be more specific?
```suggestion
            <td>The field indicating the year (within an era).</td>
```

spec/displaynames.html
```diff
@@ -31,6 +38,82 @@ <h1>CanonicalCodeForDisplayNames ( _type_, _code_ )</h1>
...
+            <td>*"quarter"*</td>
+            <td>The field indicating the quarter, e.g. Q2, 2nd quarter, etc.</td>
+          </tr>
+          <tr>
+            <td>*"month"*</td>
+            <td>The field indicating the month, e.g. Sep, September, etc.</td>
+          </tr>
+          <tr>
+            <td>*"weekOfYear"*</td>
+            <td>The field indicating the week number within the current year.</td>
```
gibson042, line 83:
```suggestion
            <td>The field indicating the week number within a year.</td>
```

spec/displaynames.html
```diff
@@ -31,6 +38,82 @@ <h1>CanonicalCodeForDisplayNames ( _type_, _code_ )</h1>
...
+            <td>*"weekday"*</td>
+            <td>The field indicating the day of week, e.g. Tue, Tuesday, etc.</td>
+          </tr>
+          <tr>
+            <td>*"day"*</td>
+            <td>The field indicating the day in month.</td>
+          </tr>
+          <tr>
+            <td>*"dayPeriod"*</td>
+            <td>The field indicating the day period, either am/pm marker or others, e.g. noon, evening.</td>
```
gibson042, line 95:
```suggestion
            <td>The field indicating the day period, either am, pm, etc. or noon, evening, etc..</td>
```

gibson042 (approved):

FrankYFTang:
@gibson042 @ryzokuken Is there any action I need to take for this PR?

ryzokuken (approved):
LGTM overall, just that old comment about rewording one sentence, would be nice if you addressed it.

FrankYFTang:
PTAL

ryzokuken:
Looks great, @gibson042 would you like to review this before we hit merge?

spec/displaynames.html
```diff
@@ -68,8 +151,13 @@ <h1>Intl.DisplayNames ( _locales_, _options_ )</h1>
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
         1. Let _types_ be _dataLocaleData_.[[types]].
         1. Assert: _types_ is a Record (see <emu-xref href="#sec-Intl.DisplayNames-internal-slots"></emu-xref>).
+        1. Let _languageDisplay_ be ? GetOption(_options_, *"languageDisplay"*, *"string"*, &laquo; *"dialect"*, *"standard"* &raquo;, *"dialect"*).
```
gibson042, line 154: Is it intentional to read `languageDisplay` even if `type` is not "language"? The consequences are observable to ECMAScript code, e.g.
```js
new Intl.DisplayNames(undefined, {
  type: "currency",
  get languageDisplay(){ throw new Error("[[Get]] languageDisplay") }
})
```

gibson042 (approved):
No firm objection to merging as-is, but I do have a question about what seems like a mistake.

spec/displaynames.html
FrankYFTang: Please see https://github.com/tc39/intl-displaynames-v2/issues/34
```
4. Intl.DisplayNames ( locales, options )
Normal convention should still be to always validate all options, even when unused.
That means the GetOption(options, "languageDisplay", ...) step should happen before If type is "language", then. (But don't move assigning displayNames.[[LanguageDisplay]] outside of the if-step.)
For example compare to SetNumberFormatUnitOptions().
```

FrankYFTang: @anba

ryzokuken:
I think all outstanding comments have been addressed and this PR has the editorial signoffs needed? Merging.

PR #624, opened 2021-11-16
Editorial: Clarify handling of time zone names

gibson042:


ptomato (commented):
I looked at Richard's similar changes in Temporal and I think these changes are a good idea. (But I'm not a 402 reviewer)

spec/locales-currencies-tz.html
```diff
@@ -155,41 +164,44 @@ <h1>IsWellFormedCurrencyCode ( _currency_ )</h1>
...
+        1. Return *false*.
+      </emu-alg>
     </emu-clause>
 
+    <emu-note>
+      Any value returned from DefaultTimeZone must be recognized as valid.
+    </emu-note>
+
     <emu-clause id="sec-canonicalizetimezonename" aoid="CanonicalizeTimeZoneName">
       <h1>CanonicalizeTimeZoneName</h1>
```
ryzokuken, line 193: Why does this AO not have an argument list?

spec/locales-currencies-tz.html
```diff
@@ -155,41 +164,44 @@ <h1>IsWellFormedCurrencyCode ( _currency_ )</h1>
...
       </p>
 
       <emu-alg>
-        1. Let _ianaTimeZone_ be the Zone or Link name of the IANA Time Zone Database such that _timeZone_, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>, is equal to _ianaTimeZone_, converted to upper case as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
-        1. If _ianaTimeZone_ is a Link name, let _ianaTimeZone_ be the corresponding Zone name as specified in the *"backward"* file of the IANA Time Zone Database.
+        1. Let _ianaTimeZone_ be the String value of the Zone or Link name of the IANA Time Zone Database that is an ASCII-case-insensitive match of _timeZone_ as described in <emu-xref href="#sec-case-sensitivity-and-case-mapping"></emu-xref>.
+        1. If _ianaTimeZone_ is a Link name, let _ianaTimeZone_ be the String value of the corresponding Zone name as specified in the file <code>backward</code> of the IANA Time Zone Database.
```
ryzokuken, line 201: are the `<code>` tags necessary vs `*`?

ryzokuken (approved):
LGTM, few questions inline.

spec/locales-currencies-tz.html
gibson042: `*backward*` suggests that "backward" is an ECMAScript Language Type Value and renders as `<emu-val>backward</emu-val>`, which isn't right. ``` `backward` ``` suggests that it is Source Code, which also isn't right (even though it would render as `<code>backward</code>`, which _looks_ fine). Ultimately this isn't a case covered well by [ecmarkup](https://tc39.es/ecmarkup/#emd-overview) AFAICT, but I was following the lead of https://tc39.es/ecma262/#sec-shared-memory-guidelines ("_… such as <code>LOCK</code>-prefixed instructions on x86_"). However, `~backward~` would suggest that it is an instance of a specification type ("_specification artefacts that do not necessarily correspond to any specific entity within an ECMAScript implementation_" per [ECMA-262](https://tc39.es/ecma262/#sec-ecmascript-specification-types)) and would render as `<emu-const>backward</emu-const>`, which I think can work. What do you think?

ryzokuken: I agree that this is the best representation in that case, but `~` might work also. One question would be: do we have any other instances of either? Should we align?

gibson042: Yes, there's a reference to Unicode Character Database file SpecialCasings.txt at [String.prototype.toLocaleLowerCase](https://tc39.es/ecma402/#sup-string.prototype.tolocalelowercase) and files UnicodeData.txt and SpecialCasings.txt at [ECMA-262 String.prototype.toLowerCase](https://tc39.es/ecma262/#sec-string.prototype.tolowercase) and file CaseFolding.txt at [ECMA-262 Canonicalize](https://tc39.es/ecma262/#sec-runtime-semantics-canonicalize-ch), which I was planning to address after this. In fact, let me open an ECMA-262 PR now.

ryzokuken (approved):

spec/locales-currencies-tz.html
gibson042: Update: The ECMA-262 editors expressed a preference for `<code>…</code>` as used in this PR.

ryzokuken:
Thanks @gibson042! Let's merge this.

PR #625, opened 2021-11-18
Editorial: Improve case sensitivity and conversion

gibson042:


spec/locale-sensitive-functions.html
```diff
@@ -58,32 +58,52 @@ <h1>String.prototype.toLocaleLowerCase ( [ _locales_ ] )</h1>
...
+          1. Let _availableLocales_ be a List with language tags that includes the languages for which the Unicode Character Database contains language sensitive case mappings. Implementations may add additional language tags if they support case mapping for additional locales.
+          1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
+          1. If _locale_ is *undefined*, set _locale_ to *"und"*.
+          1. Let _codePoints_ be ! StringToCodePoints(_S_).
+          1. If _targetCase_ is ~lower~, then
+            1. Let _newCodePoints_ be a List whose elements are the result of a lower case transformation of _codePoints_ according to an implementation-derived algorithm using _locale_ or the Unicode Default Case Conversion algorithm.
+          1. Else,
+            1. Assert: _targetCase_ is ~upper~.
+            1. Let _newCodePoints_ be a List whose elements are the result of an upper case transformation of _codePoints_ according to an implementation-derived algorithm using _locale_ or the Unicode Default Case Conversion algorithm.
+          1. Return ! CodePointsToString(_newCodePoints_).
```
gibson042, line 96: This algorithm is technically omitting a possible error, but I'm not fully decided about how to remedy that. It occurs because ECMAScript String values are bounded to a maximum of 2<sup>53</sup> - 1 code units and some case transformations can increase length, meaning there are inputs for which the output is well-defined but not representable (e.g., `"ß".repeat(2**53-1).toLocaleUpperCase("en")`). I think the right approach is updating CodePointsToString (which is [defined in ECMA-262](https://tc39.es/ecma262/multipage/ecmascript-language-source-code.html#sec-codepointstostring)) to throw an error when its input corresponds to out-of-bounds output and then updating this invocation to be explicitly fallible with `?`, although I'm willing to wait on that because AFAICT every current implementation is incapable of dealing with strings near the limit.

gibson042: https://github.com/tc39/ecma262/issues/2623

leobalter (approved):

PR #626, opened 2021-11-19
Editorial: correct references to negative zero

jugglinmike:
"Negative zero" cannot be expressed with an ECMA262 mathematical value,
but it may be expressed with an ECMA262 Number value. Update new
references to "negative zero" to use an appropriate type. Update
corresponding references to "zero" to avoid comparing mathematical
values with Number values.

/cc @bakkot

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -814,7 +814,7 @@
             1. Let _pattern_ be _patterns_.[[negativePattern]].
         1. Else,
           1. Assert: _signDisplay_ is *"exceptZero"*.
-          1. If _x_ is *NaN* or _x_ is finite and ℝ(_x_) is 0, then
+          1. If _x_ is *NaN*, or if _x_ is finite and ℝ(_x_) is 0, then
             1. Let _pattern_ be _patterns_.[[zeroPattern]].
           1. Else if ℝ(_x_) &gt; 0, then
             1. Let _pattern_ be _patterns_.[[positivePattern]].
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -157,7 +157,7 @@
       </p>
 
       <emu-alg>
-        1. If _x_ &lt; 0<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, let _isNegative_ be *true*; else let _isNegative_ be *false*.
+        1. If ℝ(_x_) &lt; 0 or _x_ is *-0*<sub>𝔽</sub>, let _isNegative_ be *true*; else let _isNegative_ be *false*.
         1. If _isNegative_, then
           1. Let _x_ be -_x_.
         1. If _intlObject_.[[RoundingType]] is ~significantDigits~, then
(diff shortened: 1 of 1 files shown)


spec/numberformat.html
```diff
@@ -157,7 +157,7 @@ <h1>FormatNumericToString ( _intlObject_, _x_ )</h1>
       </p>
 
       <emu-alg>
-        1. If _x_ &lt; 0 or _x_ is *-0*<sub>𝔽</sub>, let _isNegative_ be *true*; else let _isNegative_ be *false*.
+        1. If _x_ &lt; 0<sub>𝔽</sub> or _x_ is *-0*<sub>𝔽</sub>, let _isNegative_ be *true*; else let _isNegative_ be *false*.
```
bakkot, line 160: This one is kind of tricky - this operation is generic over both Number and BigInt values, so it's not obviously reasonable to compare against the floating-point 0.

I'd suggest instead:

```suggestion
        1. If ℝ(_x_) &lt; 0 or _x_ is *-0*<sub>𝔽</sub>, let _isNegative_ be *true*; else let _isNegative_ be *false*.
```

spec/numberformat.html
```diff
@@ -814,9 +814,9 @@ <h1>GetNumberFormatPattern ( _numberFormat_, _x_ )</h1>
             1. Let _pattern_ be _patterns_.[[negativePattern]].
         1. Else,
           1. Assert: _signDisplay_ is *"exceptZero"*.
-          1. If _x_ is 0 or _x_ is -0 or _x_ is *NaN*, then
+          1. If _x_ is 0<sub>𝔽</sub> or _x_ is -0<sub>𝔽</sub> or _x_ is *NaN*, then
```
bakkot, line 817: This operation is also generic over both Number and BigInt, so

```suggestion
          1. If _x_ is *NaN* or ℝ(_x_) is 0, then
```

(ℝ requires its argument to be finite, so we have to rule out the infinities and NaN before calling it. I'm pretty sure the infinities are screened out at the callsite, so I didn't guard against them here, but if I'm wrong this should be `or _x_ is finite and ℝ(_x_) is 0`.)

spec/numberformat.html
```diff
@@ -814,9 +814,9 @@ <h1>GetNumberFormatPattern ( _numberFormat_, _x_ )</h1>
...
             1. Let _pattern_ be _patterns_.[[zeroPattern]].
-          1. Else if _x_ &gt; 0, then
+          1. Else if _x_ &gt; 0<sub>𝔽</sub>, then
```
bakkot, line 819:
```suggestion
          1. Else if ℝ(_x_) &gt; 0, then
```

spec/numberformat.html
jugglinmike: Thanks for calling attention to the infinities in particular. I don't think those are screened at the call site: [PartitionNumberPattern](https://tc39.es/ecma402/#sec-partitionnumberpattern) has a branch for them, but it still ends up passing them to this abstract operation.

Is there any ambiguity in the prose, "If A or B and C", though?

jugglinmike:
Thanks for the review, @bakkot!

spec/numberformat.html
bakkot: I suppose it's somewhat ambiguous, though I think in this context there's no other coherent reading.

But it's probably best to be more explicit. You can distinguish the cases as in `If A, or if B and C`, maybe.

jugglinmike: Sounds good to me!

sffc:
I don't get it; in the current spec, `_x_` is a Number, not a mathematical value.  NFv3 changes it to be an Intl mathematical value.

bakkot:
`_x_` is a Number (or BigInt), but `0` is a mathematical value, and `-0` isn't anything. So saying `_x_ is 0`, as the spec text does without this PR, doesn't really mean anything - it's comparing an ES language value with a mathematical value. This PR changes it so that the comparisons (at least those it touches) are all between like types.

leobalter (approved):

PR #630, opened 2021-12-16
Editorial: Add and use definitions for ASCII-specific operations

gibson042:
Ref #624

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -11,6 +11,9 @@
     <p>
       The String values used to identify locales, currencies, scripts, and time zones are interpreted in an ASCII-case-insensitive manner, treating the code units 0x0041 through 0x005A (corresponding to Unicode characters LATIN CAPITAL LETTER A through LATIN CAPITAL LETTER Z) as equivalent to the corresponding code units 0x0061 through 0x007A (corresponding to Unicode characters LATIN SMALL LETTER A through LATIN SMALL LETTER Z), both inclusive. No other case folding equivalences are applied.
     </p>
+    <emu-note>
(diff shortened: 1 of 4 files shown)


spec/locales-currencies-tz.html
```diff
@@ -9,21 +9,20 @@ <h1>Identification of Locales, Currencies, Time Zones, and Measurement Units</h1
...
     </p>
+    <emu-note>
+      For example, *"ß"* (U+00DF) must not match or be mapped to *"SS"* (U+0053, U+0053). *"ı"* (U+0131) must not match or be mapped to *"I"* (U+0049).
+    </emu-note>
     <p>
-      Mapping to upper case takes a String value and produces a new String value in which each lower case ASCII letter code unit (0x0061 through 0x007A, inclusive) is replaced with the corresponding upper case ASCII letter code unit (0x0041 through 0x005A, inclusive) and all other code units are preserved.
+      The <dfn>ASCII-uppercase</dfn> of a String value _S_ is the String value derived from _S_ by replacing each occurrence of an ASCII lowercase letter code unit (0x0061 through 0x007A, inclusive) with the corresponding ASCII uppercase letter code unit (0x0041 through 0x005A, inclusive) while preserving all other code units.
```
ljharb, line 18: are these things that could live in 262 and be used by toUpperCase?

Alternatively, why not use the phrasing in 262:
> the result of toUppercase(sText), according to the Unicode Default Case Conversion algorithm.

gibson042: I expect these to migrate into 262 as part of Temporal (which needs ASCII-case-insensitive time zone name comparisons), but they won't use Unicode to{Upper,Lower}case or be used in the algorithms of ECMAScript `to{Upper,Lower}Case` because their transformations are explicitly restricted to the ASCII range (e.g., unlike toUppercase and `toUpperCase`, ASCII-uppercase must not transform "ö").

ljharb: Thanks for clarifying.

leobalter (approved):

Issue #648, opened 2022-01-19
PartitionNumberPattern fails to properly handle mathematical values vs. Number/BigInt


gibson042:
cf. https://github.com/tc39/ecma402/pull/645#r784295285

> FormatNumericToString expects a Number or BigInt value, while _x_ is a mathematical value that AFAICT might suffer output-affecting lossiness if converted.
> 
> And similar problems in later steps (in which whether or not _x_ is a mathematical value is dependent upon whether or not the original input was finite!)
> 
> …
> 
> I think the best fix would be updating FormatNumericToString/GetNumberFormatPattern/PartitionNotationSubPattern/etc. to accept new parameters allowing them to identify special non-mathematical Number values such as negative zero and infinities and NaN.

gibson042:
https://github.com/tc39/ecma402/pull/651#pullrequestreview-869935114

> I don't think this holds up unless the scope is expanded to properly fix #648, because _many_ related operations are incoherent in the same way:
> * [FormatNumeric](https://tc39.es/ecma402/#sec-formatnumber) is documented to accept <var>x</var> (which must be a Number or BigInt value), but FormatDateTimePattern incorrectly invokes it with mathematical values.
> * FormatNumeric then sends it to PartitionNumberPattern, which also expects <var>x</var> (a Number or a BigInt).
> * PartitionNumberPattern handles finite <var>x</var> by sending it to ComputeExponent (which implicitly and incorrectly treats it as a mathematical value and returns a mathematical value) and FormatNumericToString, replacing <var>x</var> with the [[RoundedNumber]] slot of the record returned by FormatNumericToString.
>   * FormatNumericToString accepts <var>x</var> (which must be a Number or BigInt value), but incorrectly performs some operations upon it that are only valid with mathematical values. It replaces <var>x</var> with a mathematical value from the [[RoundedNumber]] slot of a record returned by either ToRawPrecision or ToRawFixed, and then returns either that mathematical value or its negation in [[RoundedNumber]].
> * Back in PartitionNumberPattern, <var>x</var> could be either the supposedly Number/BigInt input **or** a mathematical value from FormatNumericToString.
> * PartitionNumberPattern sends <var>x</var> to GetNumberFormatPattern, which performs some operations upon it that are only valid for mathematical values and others that can only succeed if it is a Number.
[...]

PR #651, opened 2022-01-24
Editorial: clarify parameter type

jugglinmike:
The term "numeric value" does not describe a type. Replace it with a
type inferred from the abstract operation's only call site.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -263,7 +263,7 @@
     <emu-clause id="sec-partitionnotationsubpattern" aoid="PartitionNotationSubPattern">
       <h1>PartitionNotationSubPattern ( _numberFormat_, _x_, _n_, _exponent_ )</h1>
       <p>
-        The PartitionNotationSubPattern abstract operation is called with arguments _numberFormat_ (which must be an object initialized as a NumberFormat), _x_ (which is a numeric value after rounding is applied), _n_ (which is an intermediate formatted string), and _exponent_ (an integer), and creates the corresponding parts for the number and notation according to the effective locale and the formatting options of _numberFormat_. The following steps are taken:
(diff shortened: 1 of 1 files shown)


gibson042 (changes requested):
I don't think this holds up unless the scope is expanded to properly fix #648, because **many** related operations are incoherent in the same way:
* [FormatNumeric](https://tc39.es/ecma402/#sec-formatnumber) is documented to accept <var>x</var> (which must be a Number or BigInt value), but FormatDateTimePattern incorrectly invokes it with mathematical values.
* FormatNumeric then sends it to PartitionNumberPattern, which also expects <var>x</var> (a Number or a BigInt).
* PartitionNumberPattern handles finite <var>x</var> by sending it to ComputeExponent (which implicitly and incorrectly treats it as a mathematical value and returns a mathematical value) and FormatNumericToString, replacing <var>x</var> with the [[RoundedNumber]] slot of the record returned by FormatNumericToString.
  * FormatNumericToString accepts <var>x</var> (which must be a Number or BigInt value), but incorrectly performs some operations upon it that are only valid with mathematical values. It replaces <var>x</var> with a mathematical value from the [[RoundedNumber]] slot of a record returned by either ToRawPrecision or ToRawFixed, and then returns either that mathematical value or its negation in [[RoundedNumber]].
* Back in PartitionNumberPattern, <var>x</var> could be either the supposedly Number/BigInt input **or** a mathematical value from FormatNumericToString.
* PartitionNumberPattern sends <var>x</var> to GetNumberFormatPattern, which performs some operations upon it that are only valid for mathematical values and others that can only succeed if it is a Number.
* PartitionNumberPattern then sends the type-confused <var>x</var> to PartitionNotationSubPattern, the current object of this PR, which expects <var>x</var> (which is a numeric value after rounding is applied) and is sufficiently vague in its treatment that no further specification seems required.

ryzokuken:
@jugglinmike any updates on this?

jugglinmike:
Nope

sffc:
I'm going to close this; @jugglinmike  please make an issue with follow-up work.

PR #653, opened 2022-01-29
Editorial: Rename [[SegmenterGranularity]] to [[Granularity]]

IdanHo:
This eliminates the repetitiveness of:
`Set segmenter.[[SegmenterGranularity]] to granularity.`

### DIFF
--- a/spec/segmenter.html
+++ b/spec/segmenter.html
@@ -17,7 +17,7 @@
 
       <emu-alg>
         1. If NewTarget is *undefined*, throw a *TypeError* exception.
-        1. Let _internalSlotsList_ be &laquo; [[InitializedSegmenter]], [[Locale]], [[SegmenterGranularity]] &raquo;.
+        1. Let _internalSlotsList_ be &laquo; [[InitializedSegmenter]], [[Locale]], [[Granularity]] &raquo;.
         1. Let _segmenter_ be ? OrdinaryCreateFromConstructor(NewTarget, *"%Segmenter.prototype%"*, _internalSlotsList_).
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Set _options_ to ? GetOptionsObject(_options_).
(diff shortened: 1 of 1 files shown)


IdanHo:
If this is an acceptable change, I'd like to do the same for `[[SegmentsSegmenter]]` and `[[SegmentsString]]`

ljharb:
It’s only repetitive when the variable has this name tho - it’s generally best to have internal slots have as distinct names as possible, even at the cost of verbosity, to avoid collisions with other types of objects that may have the same slots.

IdanHo:
AFAICT the only use of `granularity` is in segmenter, so segmenter.[[SegmenterGranularity]] does not differentiate it from any other kind of granularity, it simply repeats the already inferred fact, that it is the granularity of the segmenter.

linusg:
> It’s only repetitive when the variable has this name tho - it’s generally best to have internal slots have as distinct names as possible, even at the cost of verbosity, to avoid collisions with other types of objects that may have the same slots.

This doesn't really match reality of current and upcoming specs. Examples from 262 that grossly violate this rule: `[[Capability]]`, `[[Type]]`, `[[Handler]]`, `[[Resolve]]`, `[[Reject]]`, `[[ContentType]]`, `[[ByteOffset]]` - the list goes on. `[[FooData]]` seems to be the exception.
Likewise in temporal, generic names like `[[ISOYear]]`, `[[Nanoseconds]]`, or `[[Calendar]]` all over the place; identification of the exact object type via `[[InitializedTemporalFoo]]`.

ljharb:
It’s not a rule - it’s a convention, discovered to be needed from those very examples.

ryzokuken:
@gibson042 what are your thoughts on this?

gibson042:
I agree with @ljharb—it's _good_ for internal slot names to have prefixes relating them to the relevant specification section, both for avoiding future collisions and for aiding spec reader comprehension.

ryzokuken:
Looks like there's no consensus here to proceed so I'll close this PR for now. @IdanHo please feel free to reopen it something changes.

sffc:
Small additional bit of discussion in the TG2 meeting: https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-03-17.md#editorial-rename-segmentergranularity-to-granularity

PR #610, opened 2021-09-09
Editorial: Identify the sanctioned units as a subset of CLDR

gibson042:
…and correctly use UTS 35 terms.

Fixes #564

PR #602, opened 2021-08-17
Editorial: Cleanup initialize date time format

gibson042:
* use standard conventions
* avoid alias redefintion
* simplify {date,time}Style vs. explicit component checking

PR #660, opened 2022-03-17
Editorial: Add "!" before infallible calls

gibson042:
This cleans up some calls that are not already addressed by my other open PRs (#602, #610, #625, #630).

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -78,7 +78,7 @@
       <emu-alg>
         1. Assert: Type(_tag_) is String.
         1. Assert: Type(_options_) is Object.
-        1. If IsStructurallyValidLanguageTag(_tag_) is *false*, throw a *RangeError* exception.
+        1. If ! IsStructurallyValidLanguageTag(_tag_) is *false*, throw a *RangeError* exception.
         1. Let _language_ be ? GetOption(_options_, *"language"*, *"string"*, *undefined*, *undefined*).
         1. If _language_ is not *undefined*, then
           1. If _language_ does not match the `unicode_language_subtag` production, throw a *RangeError* exception.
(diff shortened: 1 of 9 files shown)


spec/locale.html
```diff
@@ -78,7 +78,7 @@ <h1>ApplyOptionsToTag ( _tag_, _options_ )</h1>
       <emu-alg>
         1. Assert: Type(_tag_) is String.
         1. Assert: Type(_options_) is Object.
-        1. If IsStructurallyValidLanguageTag(_tag_) is *false*, throw a *RangeError* exception.
+        1. If ! IsStructurallyValidLanguageTag(_tag_) is *false*, throw a *RangeError* exception.
```
ljharb, line 81: given 262's completion reform, where most AOs don't need to return completions, would it make more sense to update 402 to match rather than adding a bunch of prefixes?

gibson042: Yes, quite possibly. Do you have a sense of what that would like like here?

ljharb: I’d expect a separate PR that updates every AO to use the new signature, and ecmarkup will enforce that only those that return completion records are used with the ? or ! prefix - and predicates all don’t.

It’s probably a good amount of work, but it’s also probably a good idea to do before 2022, so the two 2022’s can match.

ljharb: that makes sense, but i'm not actually sure the 402 spec is valid anymore since the implicit completion rules for AOs in the 262 spec it relies on no longer exist.

in other words, i'm not sure the 402 2022 spec *can* be cut until the completion reform is done :-/

bakkot: Given that the use of completion records in 262 was technically "invalid" for years, I don't think completion reform needs to block cutting the 2022 edition unless the current state is actually unclear to human readers.

ljharb: How was it invalid?

bakkot: Completion records were described as only containing ECMAScript language values, but held many other types of values.

(Also the "Any reference to a Completion Record value that is in a context that does not explicitly require a complete Completion Record value is equivalent to an explicit reference to the [[Value]] field of the Completion Record value unless the Completion Record is an abrupt completion." wasn't really precise enough to be meaningful.)

Ms2ger:
Can you review the `ArrayCreate` calls as well (unless you're doing those elsewhere)?

gibson042:
@Ms2ger Done, along with some other broad patterns (although I'd like to stop here for the other PRs to settle before continuing).

leobalter (approved):

PR #687, opened 2022-06-02
Editorial: Extend GetOption to support a *"number"* type

gibson042:
Ref https://github.com/tc39/proposal-temporal/pull/2243

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -327,18 +327,23 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-getoption" aoid="GetOption">
-      <h1>GetOption ( _options_, _property_, _type_, _values_, _fallback_ )</h1>
-
-      <p>
-        The abstract operation GetOption extracts the value of the property named _property_ from the provided _options_ object, converts it to the required _type_, checks whether it is one of a List of allowed _values_, and fills in a _fallback_ value if necessary. If _values_ is *undefined*, there is no fixed set of values and any is permitted.
-      </p>
-
+    <emu-clause id="sec-getoption" type="abstract operation">
+      <h1>
+        GetOption (
+          _options_: an Object,
+          _property_: a property key,
+          _type_: *"boolean"* or *"number"* or *"string"*,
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):
LGTM overall. I suppose it's editorial since it changes an AO that isn't exposed directly but it does change the behavior around `type: <something wrong>`, although I guess it's fair to assume that this case just wouldn't happen.

Also, not sure how `~required~` works?

gibson042:
@ryzokuken There is one use of GetOption in ECMA-402 that _could_ have used `~required~`, except that it throws a TypeError rather than a RangeError when the required property is undefined: the Intl.DisplayNames constructor reading a "type" option (https://tc39.es/ecma402/#sec-Intl.DisplayNames ).

spec/negotiation.html
```diff
@@ -327,21 +327,32 @@ <h1>CoerceOptionsToObject ( _options_ )</h1>
...
+          1. If _default_ is ~required~, throw a *RangeError* exception.
+          1. Return _default_.
         1. If _type_ is *"boolean"*, then
           1. Set _value_ to ! ToBoolean(_value_).
-        1. If _type_ is *"string"*, then
+        1. If _type_ is *"number"*, then
+          1. Set _value_ to ? ToNumber(_value_).
+          1. If _value_ is *NaN*, throw a *RangeError* exception.
+        1. Else,
+          1. Assert: _type_ is *"string"*.
```
ljharb, line 355: Why are these strings and not spec enums?

gibson042: I was just about to open a discussion on that topic. ECMA-402 currently uses `*"boolean"*`/`*"number"*`/`*"string"*` strings, Temporal currently uses direct `Boolean`/`String`/`Number` [Type](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-data-types-and-values) references, and the more conventional approach would be a spec enum like `~boolean~`/`~number~`/`~string~`. The string approach is obviously dispreferred, but how do we feel about direct Type vs. spec enum?

PR #689, opened 2022-06-02
Editorial: Replace use of privileged language values in GetOption

gibson042:
cf. https://github.com/tc39/ecma402/pull/687#discussion_r888113770

Use of privileged language values in abstract operation arguments is bad form. The generally preferred alternative is to use spec enum values, but [Temporal GetOption](https://tc39.es/proposal-temporal/#sec-getoption) currently uses an alternative approach that directly references [Types](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-data-types-and-values).

I'm not sure which we we want to go here, so I took the Temporal Type-reference approach in the first commit and the conventional ECMA-262 spec-enum-approach in the second for easy adoption of either. I'm interested to hear others' opinions.

Note also that Temporal will move GetOption into ECMA-262, so this is really a question of pre-alignment.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -37,24 +37,24 @@
       <emu-alg>
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Set _options_ to ? CoerceOptionsToObject(_options_).
-        1. Let _usage_ be ? GetOption(_options_, *"usage"*, *"string"*, &laquo; *"sort"*, *"search"* &raquo;, *"sort"*).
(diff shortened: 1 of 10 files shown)


michaelficarra:
The spec enums are preferred by the 262 editors. We should use the latter commit, especially if this is making its way into ECMA-262.

gibson042:
@michaelficarra is that @tc39/ecma262-editors approval of this PR in its present state?

michaelficarra:
yup, that's a good use of enums

leobalter:
LGTM

I don't want to stamp a PR approval as I'm not fully reviewing the contents of ECMA-402 to verify for any missing parts. The diff seems good, thou.

PR #690, opened 2022-06-03
Meta: Clarify the layering relationship between ECMA-262 and ECMA-402

gibson042:
As discussed on Matrix: https://matrix.to/#/!pGTjbcoTUqwfzHcRaW:matrix.org/$69qLclthEoIpaEgNKrp3TTIQKWRJ7qxgeQo8ablgfS4?via=matrix.org

> I've been thinking about Temporal, and would like to be more specific about the relationship between ECMA-262 and ECMA-402 in general. What are your feelings on proposing something formal like the following at the upcoming plenary?
> > 1. Any behavior specified in ECMA-402 (including circumstances that require throwing an exception) must be valid in an ECMA-262 implementation that does not support ECMA-402, allowing any ECMA-262 implementation to adopt it incrementally.
> > 2. ECMA-402 must not prohibit any behavior that is specified in ECMA-262 as available to implementations other than introduction of new built-in values and use of reserved function parameters, allowing any ECMA-262 implementation to adopt it without introducing backwards incompatibility.


spec/conformance.html
```diff
@@ -1,7 +1,7 @@
 <emu-clause id="conformance">
   <h1>Conformance</h1>
   <p>
-    A conforming implementation of the ECMAScript 2023 Internationalization API Specification must conform to the ECMAScript 2023 Language Specification (ECMA-262 14<sup>th</sup> Edition, or successor), and must provide and support all the objects, properties, functions, and program semantics described in this specification.
+    A conforming implementation of the ECMAScript 2023 Internationalization API Specification must conform to the ECMAScript 2023 Language Specification (ECMA-262 14<sup>th</sup> Edition, or successor), and must provide and support all the objects, properties, functions, and program semantics described in this specification. Nothing permitted by this specification is prohibited by ECMA-262 to an implementation that does not conform with this specification.
```
ljharb, line 4: this wording is a bit confusing. can you help me understand what it's trying to say?

spec/overview.html
```diff
@@ -20,7 +20,7 @@ <h1>Internationalization, Localization, and Globalization</h1>
...
-      The ECMAScript 2023 Internationalization API Specification is designed to complement the ECMAScript 2023 Language Specification by providing key language-sensitive functionality. The API can be added to an implementation of the ECMAScript 2023 Language Specification (ECMA-262 14<sup>th</sup> Edition, or successor).
+      The ECMAScript 2023 Internationalization API Specification is designed to complement the ECMAScript 2023 Language Specification by providing key language-sensitive functionality. The API can be added to an implementation of the ECMAScript 2023 Language Specification (ECMA-262 14<sup>th</sup> Edition, or successor) in whole or in part. Excepting interaction with language values introduced by this specification that are observable to ECMAScript code (such as the value of a [[FallbackSymbol]] internal slot and the set of values transitively reachable from %Intl% by <emu-xref href="#sec-property-accessors">property access</emu-xref>), nothing required by this specification is incompatible with behaviour permitted by ECMA-262 to conforming implementations and nothing prohibited by this specification is permitted by ECMA-262 to conforming implementations.
```
ljharb, line 23: why "or in part"? why would we want something to implement part of Intl?

spec/conformance.html
gibson042: That ECMA-402 does not grant powers otherwise unavailable to an ECMA-262 implementation. An example violation is https://github.com/tc39/proposal-temporal/issues/2192 , which observes that an ECMA-262 implementation will be _required_ to throw on `Temporal.Calendar.from("gregory")` unless it also implements ECMA-402.

spec/overview.html
gibson042: It's not necessarily something we _want_, but it should be allowed for an implementation to adopt ECMA-402 incrementally (e.g., `Intl.DisplayNames` before `Intl.PluralRules`). And that is in fact the case even if not mentioned here, because ECMA-262 does not prohibit any implementation from adding an `Intl` property to the global object independent of ECMA-402.

gibson042: I'll remove "in whole or in part" if you feel strongly about it, but I don't see text in either specification that would cause an implementation of ECMA-262 providing an `Intl.DisplayNames` built-in without claiming ECMA-402 support from being considered nonconforming with the former or in conflict with the latter. The clause is intended as recognition of the reality that interfaces specified in ECMA-402 can be incrementally added to a conforming implementation of ECMA-262.

sffc:
TC39-TG2 consensus: https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-06-16.md#meta-clarify-the-layering-relationship-between-ecma-262-and-ecma-402

spec/overview.html
ljharb: They certainly can, but doing so would not be in compliance with 402 until the implementation was complete.

gibson042: Are we saying the same thing? This is not a statement about ECMA-402 conformance at all, but rather about its layering relationship with ECMA-262—specifically, that each part of it is independently compatible with ECMA-262.

ljharb: Hmm, maybe we’re not. My concern is that nobody should be able to claim conformance to 402 with a partial implementation.

If this line doesn’t create that risk, then feel free to resolve it.

littledan:
It looks like this PR is talking about the evolution of the ECMA-402 specification, but it's mixed into a section which is about conformance of implementations. I'd suggest that it'd be more clear if broken out into a separate section with more clear scope.

ljharb:
changes LGTM

gibson042:
That's half right... it is not about the _evolution_ of ECMA-402, but rather about the _nature_ of it with respect to ECMA-262 (specifically that it does not grant any capabilities not already left available by 262, nor does it prohibit behavior left available by 262 for objects and functions defined there).

I've tweaked the language accordingly and invite you to take another look.

littledan (approved):
Looks a lot clearer, thanks

sffc:
TG1 consensus was achieved on 2022-07-21

PR #716, opened 2022-09-30
Normative: Stage4 PR for proposal-intl-enumeration

FrankYFTang:
Plan to suggest moving to Stage 4 in Nov 2022
https://github.com/tc39/proposal-intl-enumeration

It is already implemented in Chrome m99, Safari 15.4 and Firefox 93.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf#browser_compatibility

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -117,5 +117,32 @@
         1. Return CreateArrayFromList(_ll_).
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-intl.supportedvaluesof">
+      <h1>Intl.supportedValuesOf ( _key_ )</h1>
+
+      <p>
+      When the `supportedValuesOf` method is called with argument _key_ , the following steps are taken:
+      </p>
+
+      <emu-alg>
+        1. Let _key_ be ? ToString(_key_).
+        1. If _key_ is *"calendar"*, then
+          1. Let _list_ be AvailableCanonicalCalendars( ).
+        1. Else if _key_ is *"collation"*, then
+          1. Let _list_ be AvailableCanonicalCollations( ).
+        1. Else if _key_ is *"currency"*, then
+          1. Let _list_ be AvailableCanonicalCurrencies( ).
+        1. Else if _key_ is *"numberingSystem"*, then
+          1. Let _list_ be AvailableCanonicalNumberingSystems( ).
+        1. Else if _key_ is *"timeZone"*, then
+          1. Let _list_ be AvailableCanonicalTimeZones( ).
+        1. Else if _key_ is *"unit"*, then
+          1. Let _list_ be AvailableCanonicalUnits( ).
+        1. Else,
+          1. Throw a *RangeError* exception.
+        1. Return CreateArrayFromList( _list_ ).
+      </emu-alg>
+    </emu-clause>
   </emu-clause>
 </emu-clause>
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -361,7 +361,7 @@
     <h1>Numbering System Identifiers</h1>
 
     <p>
-      The ECMAScript 2023 Internationalization API Specification identifies numbering systems using a <em>numbering system identifier</em> as defined by <a href="https://unicode.org/reports/tr35/tr35-numbers.html#Numbering_Systems">Unicode Technical Standard #35, Part 3, Section 1</a>. Their canonical form is a string containing all lowercase letters.
+      This specification identifies numbering systems using a <dfn variants="numbering system identifiers">numbering system identifier</dfn> corresponding with the name referenced by <a href="https://unicode.org/reports/tr35/tr35-numbers.html#Numbering_Systems">Unicode Technical Standard #35 Part 3 Numbers, Section 1 Numbering Systems</a>. Their canonical form is a string containing only Unicode Basic Latin lowercase letters (U+0061 LATIN SMALL LETTER A through U+007A LATIN SMALL LETTER Z).
     </p>
 
     <emu-clause id="sec-availablecanonicalnumberingsystems" type="implementation-defined abstract operation">
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -1,8 +1,8 @@
 <emu-clause id="locales-currencies-tz">
-  <h1>Identification of Locales, Currencies, Time Zones, and Measurement Units</h1>
+  <h1>Identification of Locales, Currencies, Time Zones, Measurement Units, Numbering Systems, Collations, and Calendars</h1>
 
   <p>
-    This clause describes the String values used in the ECMAScript 2023 Internationalization API Specification to identify locales, currencies, time zones, and measurement units.
+    This clause describes the String values used in the ECMAScript 2023 Internationalization API Specification to identify locales, currencies, time zones, measurement units, numbering systems, collations, and calendars.
   </p>
 
   <emu-clause id="sec-case-sensitivity-and-case-mapping">
@@ -159,6 +159,15 @@
     </emu-clause>
   </emu-clause>
 
+  <emu-clause id="sec-availablecanonicalcurrencies" type="implementation-defined abstract operation">
+    <h1>AvailableCanonicalCurrencies (
+    ): a List of Strings</h1>
+    <dl class="header">
+    <dt>description</dt>
+      <dd>The returned List is ordered as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_, and contains unique, well-formed, and upper case canonicalized 3-letter ISO 4217 currency codes, identifying the currencies for which the implementation provides the functionality of Intl.DisplayNames and Intl.NumberFormat objects.</dd>
+    </dl>
+  </emu-clause>
+
   <emu-clause id="sec-time-zone-names">
     <h1>Time Zone Names</h1>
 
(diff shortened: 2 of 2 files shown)


spec/intl.html
```diff
@@ -117,5 +117,32 @@ <h1>Intl.getCanonicalLocales ( _locales_ )</h1>
...
+        1. Else if _key_ is *"currency"*, then
+          1. Let _list_ be AvailableCanonicalCurrencies( ).
+        1. Else if _key_ is *"numberingSystem"*, then
+          1. Let _list_ be AvailableCanonicalNumberingSystems( ).
+        1. Else if _key_ is *"timeZone"*, then
+          1. Let _list_ be AvailableCanonicalTimeZones( ).
+        1. Else if _key_ is *"unit"*, then
+          1. Let _list_ be AvailableCanonicalUnits( ).
+        1. Else,
+          1. Throw a *RangeError* exception.
```
ljharb, line 143: a RangeError, from an unknown key? i'd expect a TypeError here.

FrankYFTang: Why is that? the type of the key is String, we also throw RangeError in https://tc39.es/ecma402/#sec-getoption right?

FrankYFTang: other places in ECMA402 we also throw RangeError
https://tc39.es/ecma402/#sec-canonicalcodefordisplaynames
https://tc39.es/ecma402/#sec-canonicalizelocalelist
"a. If collation does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception."
in https://tc39.es/ecma402/#sec-initializecollator
"a. If calendar does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.
a. If numberingSystem does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.
b. If the result of ! [IsValidTimeZoneName](https://tc39.es/ecma402/#sec-isvalidtimezonename)(timeZone) is false, then
"
in https://tc39.es/ecma402/#sec-initializedatetimeformat

"a. If calendar does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.
a. If collation does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.
a. If numberingSystem does not match the Unicode Locale Identifier type nonterminal, throw a RangeError exception.
"
in https://tc39.es/ecma402/#sec-Intl.Locale
also
https://tc39.es/ecma402/#sec-apply-options-to-tag
https://tc39.es/ecma402/#sec-initializenumberformat

"

ljharb: That seems very strange - RangeErrors are for when something is out of a range - enums aren't a "range", conceptually or otherwise.

gibson042: [RangeError](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-native-error-types-used-in-this-standard-rangeerror) (emphasis mine)
> Indicates a value that is not in the **set** or range of allowable values.

It's also used this way in ECMA-262:
* [NumberToBigInt](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-numbertobigint)
* [String.fromCodePoint](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.fromcodepoint)
* [String.prototype.normalize](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.normalize) (an especially good analog)
* [InitializeTypedArrayFromArrayBuffer](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-initializetypedarrayfromarraybuffer)

ljharb: fromCodePoint is an actual ascii range; the AOs are ranges as well; but you’re right that normalize is a good analog. It still seems super weird to me, since an enum simply isn’t a range of any kind.

sffc:
Ping to the editors to review this PR.

spec/locales-currencies-tz.html
```diff
@@ -315,5 +347,64 @@ <h1>IsSanctionedSingleUnitIdentifier ( _unitIdentifier_ )</h1>
...
+        <dd>The returned List is ordered as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_, and contains unique canonical numbering systems identifiers identifying the numbering systems for which the implementation provides the functionality of Intl.DateTimeFormat, Intl.NumberFormat, and Intl.RelativeTimeFormat objects. The list must include the Numbering System value of every row of <emu-xref href="#table-numbering-system-digits"></emu-xref>, except the header row.</dd>
+      </dl>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-collation-types">
+    <h1>Collation Types</h1>
+
+    <p>
+      The ECMAScript 2023 Internationalization API Specification identifies collations using a <em>collation type</em> as defined by <a href="https://unicode.org/reports/tr35/tr35-collation.html#Collation_Types">Unicode Technical Standard #35, Part 5, Section 3.1</a>. Their canonical form is a string containing all lowercase letters with zero or more hyphens.
```
gibson042, line 381:
```suggestion
      This specification identifies collations using a <dfn variants="collation types">collation type</dfn> as defined by <a href="https://unicode.org/reports/tr35/tr35-collation.html#Collation_Types">Unicode Technical Standard #35 Part 5 Collation, Section 3.1 Collation Types</a>. Their canonical form is a string containing only Unicode Basic Latin lowercase letters (U+0061 LATIN SMALL LETTER A through U+007A LATIN SMALL LETTER Z) with zero or more medial hyphens (U+002D HYPHEN-MINUS).
```

spec/locales-currencies-tz.html
```diff
@@ -315,5 +347,64 @@ <h1>IsSanctionedSingleUnitIdentifier ( _unitIdentifier_ )</h1>
...
+        <dd>The returned List is ordered as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_, and contains unique canonical collation types identifying the collations for which the implementation provides the functionality of Intl.Collator objects.</dd>
+      </dl>
+    </emu-clause>
+  </emu-clause>
+
+  <emu-clause id="sec-calendar-types">
+    <h1>Calendar Types</h1>
+
+    <p>
+      The ECMAScript 2023 Internationalization API Specification identifies calendars using a <em>calendar type</em> as defined by <a href="https://unicode.org/reports/tr35/tr35-dates.html#Calendar_Elements">Unicode Technical Standard #35, Part 4, Section 2</a>. Their canonical form is a string containing all lower case letters with zero or more hyphens.
```
gibson042, line 398:
```suggestion
      The ECMAScript 2023 Internationalization API Specification identifies calendars using a <dfn variants="calendar types">calendar type</dfn> as defined by <a href="https://unicode.org/reports/tr35/tr35-dates.html#Calendar_Elements">Unicode Technical Standard #35 Part 4 Dates, Section 2 Calendar Elements</a>. Their canonical form is a string containing only Unicode Basic Latin lowercase letters (U+0061 LATIN SMALL LETTER A through U+007A LATIN SMALL LETTER Z) with zero or more medial hyphens (U+002D HYPHEN-MINUS).
```

gibson042 (approved):
Looks good after resolving some editorial nits, mostly relating to #735 or recent ECMA-262 changes.

FrankYFTang:
PTAL and merge if ready?

gibson042 (approved):
A couple more phrase replacements.

FrankYFTang:
PTAL and merge

spec/locales-currencies-tz.html
```diff
@@ -213,6 +222,29 @@ <h1>DefaultTimeZone ( )</h1>
...
+
+      <emu-alg>
+        1. Let _names_ be a List of all Zone and Link names in the IANA Time Zone Database that are supported by the implementation.
+        1. Let _result_ be a new empty List.
+        1. For each element _name_ of _names_, do
+          1. Assert: ! IsValidTimeZoneName( _name_ ) is *true*.
+          1. Let _canonical_ be ! CanonicalizeTimeZoneName( _name_ ).
+          1. If _result_ does not contain _canonical_, then
+            1. Append _canonical_ to the end of _result_.
+        1. Sort _result_ in order as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_.
```
gibson042, line 244: There's a lint issue here: https://github.com/tc39/ecma402/actions/runs/4137409543/jobs/7152541910?pr=716
> ```
> error: could not find a preceding declaration for "comparefn" (use-before-def) at spec/locales-currencies-tz.html:249:135:
>   247 |           1. If _result_ does not contain _canonical_, then
>   248 |             1. Append _canonical_ to the end of _result_.
> > 249 |         1. Sort _result_ in order as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_.
>       |                                                                                                                                       ^
>   250 |         1. Return _result_.
>   251 |       </emu-alg>
>   252 |     </emu-clause>
> ```

```suggestion
        1. [declared="comparefn"] Sort _result_ in order as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_.
```

gibson042 (approved):
This still looks good to me, modulo the linting fix. @ryzokuken, could you look it over as well?

spec/locales-currencies-tz.html
```diff
@@ -213,6 +222,29 @@ <h1>DefaultTimeZone ( )</h1>
         The DefaultTimeZone abstract operation returns a String value representing the valid (<emu-xref href="#sec-isvalidtimezonename"></emu-xref>) and canonicalized (<emu-xref href="#sec-canonicalizetimezonename"></emu-xref>) time zone name for the host environment's current time zone.
       </p>
     </emu-clause>
+
+    <emu-clause id="sec-availablecanonicaltimezones" type="implementation-defined abstract operation">
+      <h1>
+        AvailableCanonicalTimeZones (
```
justingrant, line 228:
```suggestion
        AvailableCanonicalTimeZoneIdentifiers (
```

With Temporal approaching Stage 4, "TimeZone" will soon be ambiguous: does it mean string identifier or a `Temporal.TimeZone` object?

Therefore @gibson042 and I are planning to rename all identifier-related AOs with an "Identifier" suffix. Given that this is a new AO, should we rename it here too to save churn later?

FrankYFTang: how about do those after merge this PR. Temporal is not going to be ES2023 but this will be

justingrant: OK, that's fine with me. Not a big deal either way.

PR #717, opened 2022-09-30
Normative: Stage 4 PR for Intl Locale Info proposal

FrankYFTang:
Implemented in
Chrome 99, Safari 15.4, Opera 85
see
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/textInfo


FrankYFTang:
I will close this PR since we need to take a big change to this API

Issue #634, opened 2021-12-28
Non-ASCII characters are inconsistently represented


gibson042:
cf. https://github.com/tc39/ecma402/pull/621#discussion_r769059711 :
> ECMA-262 prefers HTML character entities and ECMA-402 should follow suit. But we'll probably just update everything at once.
> 
> ```shell
> $ egrep --no-filename -ro \
>   '[^][\n !"#$%&'\''()*+,./0-9:;<=>?@A-Z\\^_`a-z{|}~-]|&[[:alnum:]]*;' spec/ | \
>   sort | uniq -c
>      10 ×
>      11 ≠
>       3 ≤
>       3 ≥
>       9 §
>      23 –
>      16 𝔽
>       5 &ge;
>      84 &gt;
>       3 &infin;
>      79 &laquo;
>       2 &ldquo;
>       1 &le;
>      87 &lt;
>       1 &mdash;
>       1 &ne;
>      79 &raquo;
>       2 &rdquo;
>       1 &reg;
> ```

ben-allen:
In the time between when this issue was initially posted and now ECMA-262 has replaced HTML character entities  (see: https://github.com/tc39/ecma262/pull/2901) with the Unicode characters they represent.  I've put up https://github.com/tc39/ecma402/pull/907, which makes character representation in ECMA-402 consistent by using Unicode characters rather than HTML entities &mdash; the representation that 262 uses now, which is the opposite of the representation they standardized on back in 2021.

ben-allen:
Closed by https://github.com/tc39/ecma402/pull/907

PR #719, opened 2022-10-01
Meta: bump ecmarkup to 16.0.0

bakkot:
This bumps ecmarkup, which now includes a lint for variables being declared before being used. I've marked two sites which appear to be introducing a variable as doing so, and corrected a typo in a third place.

There's [another place](https://github.com/tc39/ecma402/blob/469814ca60a2d5e1283e7caabf038393f55a44d8/spec/negotiation.html#L178) which refers to a variable which doesn't appear to be defined anywhere, but I couldn't figure out what that was supposed to be doing, so I've left it alone. That should cause this to fail CI. Please feel free to push up a commit fixing that issue before landing this.

Edit: oh, looks like lint rules aren't enforced in CI, so it's not failing the build. You might want to do that - it's just a matter of adding `--strict` to the build command.

(Ignore the branch name; I was originally intending for this to include a github action to publish the biblio but I'll do that as a followup.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "SEE LICENSE IN https://tc39.es/ecma402/#sec-copyright-and-software-license",
       "dependencies": {
         "@tc39/ecma262-biblio": "2.1.2458",
-        "ecmarkup": "^12.1.1"
+        "ecmarkup": "^16.0.0"
       },
       "devDependencies": {}
     },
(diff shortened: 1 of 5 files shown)


ryzokuken (approved):

gibson042:
Can this also resolve #634 via https://github.com/tc39/ecmarkup/pull/481 ?

bakkot:
Running `npm x emu-format -- --write 'spec/*'` after this PR will translate almost all HTML entities into their corresponding codepoints, yes. It will also change a number of other formatting details to line up with the conventions used in ecma262.

ryzokuken:
@gibson042 can we go ahead and merge this?

spec/numberformat.html
```diff
@@ -641,7 +641,7 @@ <h1>PartitionNotationSubPattern ( _numberFormat_, _x_, _n_, _exponent_ )</h1>
             1. Else if _p_ is equal to *"number"*, then
               1. If the _numberFormat_.[[NumberingSystem]] matches one of the values in the *"Numbering System"* column of <emu-xref href="#table-numbering-system-digits"></emu-xref> below, then
                 1. Let _digits_ be a List whose 10 String valued elements are the UTF-16 string representations of the 10 _digits_ specified in the *"Digits"* column of the matching row in <emu-xref href="#table-numbering-system-digits"></emu-xref>.
-                1. Replace each _digit_ in _n_ with the value of _digits_[_digit_].
+                1. [declared="digit"] Replace each _digit_ in _n_ with the value of _digits_[_digit_].
```
gibson042, line 644: This is no worse than the current text, but it is highlighting a problem that I think we can avoid by following the precedents of [ASCII operations](https://tc39.es/ecma402/#sec-case-sensitivity-and-case-mapping) and [GetSubstitution](https://tc39.es/ecma262/multipage/text-processing.html#sec-getsubstitution).
```suggestion
                1. [declared="digit"] Set _n_ to the String value derived from _n_ by replacing each occurrence of an ASCII digit code unit (0x0030 through 0x0039, inclusive) _digit_ with the corresponding sequence of code units from _digits_[_digit_ - 0x0030].
```
or even more explicitly:
```suggestion
                1. Let _transliterated_ be the empty String.
                1. Let _len_ be the length of _n_.
                1. Let _position_ be 0.
                1. Repeat, while _position_ &lt; _len_,
                  1. Let _c_ be the code unit at index _position_ within _n_.
                  1. If 0x0030 ≤ _c_ ≤ 0x0039, then
                    1. NOTE: _c_ is an ASCII digit.
                    1. Let _i_ be _c_ - 0x0030.
                    1. Set _c_ to _digits_[_i_].
                  1. Set _transliterated_ to the string-concatenation of _transliterated_ and _c_.
                  1. Set _position_ to _position_ + 1.
                1. Set _n_ to _transliterated_.
```

gibson042 (approved):
I have some suggestions to avoid `[declared="digit"]`, but would not object to handling that in a followup.

PR #724, opened 2022-11-02
Normative: Canonicalise "GMT" to "UTC"

anba:
tzdata 2022f using vanguard format made "GMT" a zone and "Etc/GMT" a link to "GMT". Rearguard format still has this the other way around, i.e. "Etc/GMT" is a zone and "GMT" a link to "Etc/GMT". Because we don't control if vanguard or rearguard format is used by an implementation, the `CanonicalizeTimeZoneName` operation should canonicalise both "Etc/GMT" and "GMT" to "UTC".

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -201,7 +201,7 @@
       <emu-alg>
         1. Let _ianaTimeZone_ be the String value of the Zone or Link name of the IANA Time Zone Database that is an ASCII-case-insensitive match for _timeZone_.
         1. If _ianaTimeZone_ is a Link name, let _ianaTimeZone_ be the String value of the corresponding Zone name as specified in the file <code>backward</code> of the IANA Time Zone Database.
-        1. If _ianaTimeZone_ is *"Etc/UTC"* or *"Etc/GMT"*, return *"UTC"*.
+        1. If _ianaTimeZone_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, return *"UTC"*.
         1. Return _ianaTimeZone_.
       </emu-alg>
     </emu-clause>


anba:
tzdata 2022f release notes mentioning this change:
- https://mm.icann.org/pipermail/tz-announce/2022-October/000075.html

tzdata definitions for "Etc/GMT" and "GMT":
- https://github.com/eggert/tz/blob/b4253bbe8f781686595cadf1e2c9ec0cd91bd5a9/etcetera#L23-L35
- https://github.com/eggert/tz/blob/b4253bbe8f781686595cadf1e2c9ec0cd91bd5a9/backward#L63-L74

ryzokuken:
Are you sure? From what I understand, GMT and UTC are semantically different and this mental model is propagated within Temporal (GMT being an actual human timezone and UTC being a constant zero value, quite like `"gregory"` vs `"iso8601"` when it comes to calendars).

anba:
The current definition of [CanonicalizeTimeZoneName](https://tc39.es/ecma402/#sec-canonicalizetimezonename) is already canonicalising "Etc/GMT" to "UTC". This change is only ensuring that the canonicalisation is the same irrespective of whether the vanguard or the rearguard format is used.

ryzokuken (approved):
Fair enough, I should've checked the spec diff first :see_no_evil: 

Either way, pinging @ptomato @justingrant @sffc @gibson042 to discuss this if needed.

ptomato:
No discussion needed from my side. It's true that in some usages "GMT" is a time zone while "UTC" is the quantity, but the TZDB doesn't treat it like that: https://github.com/eggert/tz/blob/main/etcetera

gibson042 (approved):
> It's true that in some usages "GMT" is a time zone while "UTC" is the quantity, but the TZDB doesn't treat it like that: https://github.com/eggert/tz/blob/main/etcetera

@ptomato can you elaborate? As far as I can tell, tzdata has and continues to differentiate GMT and UTC. But the conflation of them in ECMA-402 already exists, and I support making sure that it is at least consistent.

ryzokuken:
> As far as I can tell, tzdata has and continues to differentiate GMT and UTC. But the conflation of them in ECMA-402 already exists, and I support making sure that it is at least consistent.

This exactly. I'm all for continuing to be consistent on the side of 402, but there's this semantic inconsistency between Temporal (which treats both as distinct timezones) and 402. At the very least, we should discuss if this is desirable or if we need to address this in either or both.

ptomato:
> can you elaborate? As far as I can tell, tzdata has and continues to differentiate GMT and UTC.

I meant that TZDB contains a "UTC" time zone whereas some people use slightly different terminology and say that "GMT" is a time zone and "UTC" is not. It's true that they are separate time zones in the TZDB and the conflation of the two comes from ECMA-402's CanonicalizeTimeZone.

> Temporal (which treats both as distinct timezones)

I clarified this in the ECMA-402 meeting, but for the record: in ECMA-402 implementations, [new Temporal.TimeZone](https://tc39.es/proposal-temporal/#sec-temporal.timezone) uses CanonicalizeTimeZone, so there is no discrepancy with Temporal.

ryzokuken:
2022-11-03: TG2 consensus.

sffc:
TG2 discussion and consensus: https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-11-03.md#normative-canonicalise-gmt-to-utc-724

ryzokuken:
TG1 consensus: https://github.com/tc39/notes/blob/main/meetings/2022-11/nov-29.md#conclusiondecision

PR #729, opened 2022-11-15
Normative: Correct Gregorian calendar BC(E) era to signed years less than 1 AD

### LINKED ISSUE #723: Wrong era for ISO year 0 in Gregorian calendar
In the [[Era]] row of https://tc39.es/ecma402/#table-datetimeformat-tolocaltime-record:

> Let `year` be `YearFromTime(tz)` specified in es2023's [Year Number](https://tc39.es/ecma262/#sec-year-number). If year is less than 0, return 'BC', else, return 'AD'.

This should be "less than 1" because ISO year 0 is the year 1 BC (or 1 BCE if you prefer the "common era" nomenclature). There is no year 0 BC or 0 AD.

I ran a quick check with eshost, and none of the engines that I have installed implement the spec as written:
```
$ eshost -e 'new Date(Date.parse("+000000-01-01")).toLocaleDateString(undefined, {era:"short", timeZone:"UTC"})'
#### ChakraCore
1 1, 1 BC

#### engine262

Error: Date.prototype.toLocaleDateString is not implemented

#### GraalJS
1/1/1 BC

#### JavaScriptCore
1 1, 1 BC

#### Moddable XS
Fri Dec 31 -0001

#### QuickJS
12/31/-0001

#### SpiderMonkey
1/1/1 B

#### V8
1/1/1 BC
```
As you can see, output varies quite a lot, but for all the engines that do include an era in the output, it is always BC for ISO year 0. So this change would reflect web reality.

(Spotted by @gibson042.)

gibson042:
Fixes #723

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -1197,7 +1197,7 @@
           </tr>
           <tr>
             <td>[[Era]]</td>
-            <td>Let _year_ be YearFromTime(𝔽(floor(_tz_ / 10<sup>6</sup>))). If _year_ &lt; *-0*<sub>𝔽</sub>, return *"BC"*, else return *"AD"*.</td>
+            <td>Let _year_ be YearFromTime(𝔽(floor(_tz_ / 10<sup>6</sup>))). If _year_ &lt; *1*<sub>𝔽</sub>, return *"BC"*, else return *"AD"*.</td>
           </tr>
           <tr>
             <td>[[Year]]</td>


gibson042:
test262 PR: https://github.com/tc39/test262/pull/3722

ptomato (approved):
It's my understanding that this is correct, and represents existing web reality.

ryzokuken (approved):

sffc:
TG2 consensus: https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-12-08.md#normative-correct-gregorian-calendar-bce-era-to-signed-years-less-than-1-ad-729

gibson042:
This change achieved consensus at the TC39 plenary of 2023-01-31.

PR #733, opened 2022-11-28
Normative: support fractionalSecondDigits to be 0

### LINKED ISSUE #590: DateTimeFormat `fractionalSecondDigits`: conflict between MDN and spec

romulocintra:
Closes #590 

Update 402 to :

Accept `0` to mean the same as `undefined`

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -101,7 +101,8 @@
         1. For each row of <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, in table order, do
           1. Let _prop_ be the name given in the Property column of the row.
           1. If _prop_ is *"fractionalSecondDigits"*, then
-            1. Let _value_ be ? GetNumberOption(_options_, *"fractionalSecondDigits"*, 1, 3, *undefined*).
+            1. Let _value_ be ? GetNumberOption(_options_, *"fractionalSecondDigits"*, 0, 3, *undefined*).
+            1. If _value_ is 0, set _value_ to *undefined*.
           1. Else,
             1. Let _values_ be a List whose elements are the strings given in the Values column of the row.
             1. Let _value_ be ? GetOption(_options_, _prop_, ~string~, _values_, *undefined*).


FrankYFTang:
For dealing with 0, the current shape of ECMA402 disallows 0. And from what I read in https://github.com/tc39/ecma402/issues/590 there are two only reasons mentioned:
MDN states it accept 0 (which is a MDN bug which should be fixed in MDN)
“Temporal allows zero for this option, so regardless of whether zero is allowed or not, we should probably align across Temporal and 402. “ . This is a valid one, but the only changes needed for Temporal about this is in InitializeDateTimeFormat. But Temporal already have a section to modify InitializeDateTimeFormat in https://tc39.es/proposal-temporal/#sec-temporal-initializedatetimeformat I see no reason such change cannot be just amended to Temporal spec only, if that is the ONLY reason we want to accept 0. And I see no value to add 0 now into ECMA402 prior to that.

During the 2022 Oct and Nov meeting, I attempted to get the motivation of this change, beside what needed in Temporal (which could be addressed inside Temporal) but both time the conclusion is
"to discuss offline " and
"
SFC: We should move this discussion offline.

FYT: I'd like the PR author to write more about the intention of the PR.

RCA: I’ll provide the description/intention of PR
"

So I am still waiting for you to write down the intention and motivaiton why is this needed. 

I see no benefit for the users for such change. I simply feel this PR is not needed at all.

FrankYFTang:
Why we cannot just simply change the MDN to reflect the web reality instead?

FrankYFTang:
Be aware, currently in the Temporal spec which passed Stage 3, the fractionalSecondDigits for toLocaleString is set for 1, 2, 3 only 
See  step 36-a-1 of https://tc39.es/proposal-temporal/#sec-temporal-initializedatetimeformat 

And according to https://tc39.es/process-document/
"Post-Acceptance Changes Expected" for Stage 3 is
"Limited: only those deemed critical based on implementation experience" 

and to my undertanding, no one how is currently implmenet Temporal request it change the toLocaleString to accept fractionalSecondDigits  .  The only one raise this issue is @justingrant  who has plenty of chances to change the Temporal spec before reaching Stage 3 for different set of acceptance values of fractionalSecondDigits in toLocaleString but he didn't do so. 

and from my understanding @justingrant is a champion of Temporal spec but NOT someone who is working on "implementation experience"  

I totally understand his desire to make the the option bag share accross toString and toLocaleStrring in Temporal. He hand the opportunity to do so before Stage 3 and he choose not to resolve that so Stage 3 reached consensus  with only 1,2,3 for toLocaleString. Therefore, for Temporal to take 0, 4-9 post Stage 3, it need to come from an issue which "Limited: only those deemed critical based on implementation experience" 

I will block such request in TC39 for sure. at minimum for the 4-9 part from my "critical based on implementation experience" if Temporal champion attempt to make such changes. But what I do not want to see in here is ECMA402 make any move to give the Temporal champion a reason to convince TC39 why that is needed. ECMA402 has no reason to change beside the introduction of Temporal and therefore we should not make such change in ECMA402 independently from Temporal, espeically Temporal already have a dedicate chapter to amend ECMA402. Those necessary changes should be in that chapter. 

If there are other benefit to accept 0, and not accepting 0 cause important issue, please state it and provide a real use case. Otherwise, I really feel this PR is counter produtive, frangment the web and change behavior without any warrent.

FrankYFTang:
Notice, right now, all browser throw RangeError when fractionalSecondDigit is 0 so web compatability could not be a reason to make such change.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2022-12-08.md#normative-support-fractionalseconddigits-to-be-0-733

Conclusion: Put this issue on the backlog and revisit _after_ Temporal is merged.

PR #743, opened 2023-01-11
[Editorial] Add back aoid="PartitionNumberPattern"

sffc:
Seems that the `aoid` may have gotten dropped.


sffc:
OK, and the CI seems to agree with that :)

PR #753, opened 2023-01-30
Normative: Integrate Intl NumberFormat v3 into ECMA-402

sffc:
https://github.com/tc39/proposal-intl-numberformat-v3

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -85,7 +85,7 @@
       <emu-alg>
         1. Let _candidate_ be _locale_.
         1. Repeat,
-          1. If _availableLocales_ contains an element equal to _candidate_, return _candidate_.
+          1. If _availableLocales_ contains _candidate_, return _candidate_.
           1. Let _pos_ be the character index of the last occurrence of *"-"* (U+002D) within _candidate_. If that character does not occur, return *undefined*.
           1. If _pos_ &ge; 2 and the character *"-"* occurs at index _pos_ - 2 of candidate, decrease _pos_ by 2.
           1. Let _candidate_ be the substring of _candidate_ from position 0, inclusive, to position _pos_, exclusive.
@@ -353,7 +353,31 @@
         1. Else,
           1. Assert: _type_ is ~string~.
           1. Set _value_ to ? ToString(_value_).
-        1. If _values_ is not ~empty~ and _values_ does not contain an element equal to _value_, throw a *RangeError* exception.
+        1. If _values_ is not ~empty~ and _values_ does not contain _value_, throw a *RangeError* exception.
+        1. Return _value_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-getbooleanorstringnumberformatoption" type="abstract operation">
+      <h1>
+        GetBooleanOrStringNumberFormatOption (
+          _options_: an Object,
+          _property_: a property key,
+          _stringValues_: a List of Strings,
+          _fallback_: an ECMAScript language value,
+        ): either a normal completion containing either a Boolean, String, or _fallback_, or a throw completion
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It extracts the value of the property named _property_ from the provided _options_ object. It returns _fallback_ if that value is *undefined*, *true* if that value is *true*, *false* if that value coerces to *false*, and otherwise coerces it to a String and returns the result if it is allowed by _stringValues_.</dd>
+      </dl>
+      <emu-alg>
+        1. Let _value_ be ? Get(_options_, _property_).
+        1. If _value_ is *undefined*, return _fallback_.
+        1. If _value_ is *true*, return *true*.
+        1. If ToBoolean(_value_) is *false*, return *false*.
+        1. Let _value_ be ? ToString(_value_).
+        1. If _stringValues_ does not contain _value_, throw a *RangeError* exception.
         1. Return _value_.
       </emu-alg>
     </emu-clause>
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -160,21 +160,20 @@
           1. Else,
             1. Set _intlObj_.[[MinimumFractionDigits]] to _mnfdDefault_.
             1. Set _intlObj_.[[MaximumFractionDigits]] to _mxfdDefault_.
-        1. If _needSd_ is *true* or _needFd_ is *true*, then
-          1. If _roundingPriority_ is *"morePrecision"*, then
-            1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
-          1. Else if _roundingPriority_ is *"lessPrecision"*, then
-            1. Set _intlObj_.[[RoundingType]] to ~lessPrecision~.
-          1. Else if _hasSd_ is *true*, then
-            1. Set _intlObj_.[[RoundingType]] to ~significantDigits~.
-          1. Else,
-            1. Set _intlObj_.[[RoundingType]] to ~fractionDigits~.
-        1. Else,
-          1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
+        1. If _needSd_ is *false* and _needFd_ is *false*, then
           1. Set _intlObj_.[[MinimumFractionDigits]] to 0.
           1. Set _intlObj_.[[MaximumFractionDigits]] to 0.
           1. Set _intlObj_.[[MinimumSignificantDigits]] to 1.
           1. Set _intlObj_.[[MaximumSignificantDigits]] to 2.
+          1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
+        1. Else if _roundingPriority_ is *"morePrecision"*, then
+          1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
+        1. Else if _roundingPriority_ is *"lessPrecision"*, then
+          1. Set _intlObj_.[[RoundingType]] to ~lessPrecision~.
+        1. Else if _hasSd_ is *true*, then
+          1. Set _intlObj_.[[RoundingType]] to ~significantDigits~.
+        1. Else,
+          1. Set _intlObj_.[[RoundingType]] to ~fractionDigits~.
         1. If _roundingIncrement_ is not 1, then
           1. If _intlObj_.[[RoundingType]] is not ~fractionDigits~, throw a *TypeError* exception.
           1. If _intlObj_.[[MaximumFractionDigits]] is not equal to _intlObj_.[[MinimumFractionDigits]], throw a *RangeError* exception.
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -96,14 +96,20 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-setnfdigitoptions" aoid="SetNumberFormatDigitOptions">
-      <h1>SetNumberFormatDigitOptions ( _intlObj_, _options_, _mnfdDefault_, _mxfdDefault_, _notation_ )</h1>
-
-      <p>
-        The abstract operation SetNumberFormatDigitOptions applies digit
-        options used for number formatting onto the intl object.
-        It is used by both Intl.NumberFormat and Intl.PluralRules.
-      </p>
+    <emu-clause id="sec-setnfdigitoptions" type="abstract operation">
+      <h1>
+        SetNumberFormatDigitOptions (
+          _intlObj_: an Object,
+          _options_: an Object,
+          _mnfdDefault_: a Number,
+          _mxfdDefault_: a Number,
+          _notation_: a String,
+        )
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>It populates the internal slots of _intlObj_ that affect locale-independent number rounding (see <emu-xref href="#sec-formatnumberstring"></emu-xref>).</dd>
+      </dl>
       <emu-alg>
         1. Let _mnid_ be ? GetNumberOption(_options_, *"minimumIntegerDigits,"*, 1, 21, 1).
         1. Let _mnfd_ be ? Get(_options_, *"minimumFractionDigits"*).
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -574,87 +574,87 @@
           <tr>
             <td>*"ceil"*</td>
             <td>Toward positive infinity</td>
-            <td>➕ [-1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [2]</td>
+            <td>⬆️ [-1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [2]</td>
           </tr>
           <tr>
             <td>*"floor"*</td>
             <td>Toward negative infinity</td>
-            <td>➖ [-2]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➖ [1]</td>
+            <td>⬇️ [-2]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [1]</td>
           </tr>
           <tr>
             <td>*"expand"*</td>
             <td>Away from zero</td>
-            <td>➖ [-2]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [2]</td>
+            <td>⬇️ [-2]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [2]</td>
           </tr>
           <tr>
             <td>*"trunc"*</td>
             <td>Toward zero</td>
-            <td>➕ [-1]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➖ [1]</td>
+            <td>⬆️ [-1]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [1]</td>
           </tr>
           <tr>
             <td>*"halfCeil"*</td>
             <td>Ties toward positive infinity</td>
-            <td>➕ [-1]</td>
-            <td>➖ [0]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [2]</td>
+            <td>⬆️ [-1]</td>
+            <td>⬇️ [0]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [2]</td>
           </tr>
           <tr>
             <td>*"halfFloor"*</td>
             <td>Ties toward negative infinity</td>
-            <td>➖ [-2]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➕ [1]</td>
-            <td>➖ [1]</td>
+            <td>⬇️ [-2]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬆️ [1]</td>
+            <td>⬇️ [1]</td>
           </tr>
           <tr>
             <td>*"halfExpand"*</td>
             <td>Ties away from zero</td>
-            <td>➖ [-2]</td>
-            <td>➖ [0]</td>
-            <td>➕ [1]</td>
-            <td>➕ [1]</td>
-            <td>➕ [2]</td>
+            <td>⬇️ [-2]</td>
+            <td>⬇️ [0]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [2]</td>
           </tr>
           <tr>
             <td>*"halfTrunc"*</td>
             <td>Ties toward zero</td>
-            <td>➕ [-1]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➕ [1]</td>
-            <td>➖ [1]</td>
+            <td>⬆️ [-1]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬆️ [1]</td>
+            <td>⬇️ [1]</td>
           </tr>
           <tr>
             <td>*"halfEven"*</td>
             <td>Ties toward an even rounding increment multiple</td>
-            <td>➖ [-2]</td>
-            <td>➖ [0]</td>
-            <td>➖ [0]</td>
-            <td>➕ [1]</td>
-            <td>➕ [2]</td>
+            <td>⬇️ [-2]</td>
+            <td>⬇️ [0]</td>
+            <td>⬇️ [0]</td>
+            <td>⬆️ [1]</td>
+            <td>⬆️ [2]</td>
           </tr>
         </tbody>
       </table>
-      <emu-note>The examples are illustrative of the unique behaviour of each option. ➕ means "resolves toward positive infinity"; ➖ means "resolves toward negative infinity".</emu-note>
+      <emu-note>The examples are illustrative of the unique behaviour of each option. ⬆️ means "resolves toward positive infinity"; ⬇️ means "resolves toward negative infinity".</emu-note>
     </emu-table>
 
     <p>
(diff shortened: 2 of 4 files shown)


sffc:
@gibson042 or @ryzokuken can you suggest how to fix the following ecmarkup error:

```
          1. Let _n1_ and _e1_ each be an integer and _r1_ a mathematical value, with <emu-eqn>_r1_ = ToRawPrecisionFn(_n1_, _e1_, _p_)</emu-eqn>, such that <emu-eqn>_r1_ ≤ _x_</emu-eqn> and _r1_ is maximized.
...
          1. Let _r_ be ApplyUnsignedRoundingMode(_x_, _r1_, _r2_, _unsignedRoundingMode_).
```

causes errors such as

```
[2023-01-30T17:13:34.352Z] Error: spec/numberformat.html:1263:56: could not find a preceding declaration for "r1"
[2023-01-30T17:13:34.352Z] Error: spec/numberformat.html:1263:62: could not find a preceding declaration for "r2"
```

gibson042:
@sffc https://tc39.es/ecmarkup/#variable-declarations
```diff
-          1. Let _n1_ and _e1_ each be an integer and _r1_ a mathematical value, with <emu-eqn>_r1_ = ToRawPrecisionFn(_n1_, _e1_, _p_)</emu-eqn>, such that <emu-eqn>_r1_ ≤ _x_</emu-eqn> and _r1_ is maximized.
-          1. Let _n2_ and _e2_ each be an integer and _r2_ a mathematical value, with <emu-eqn>_r2_ = ToRawPrecisionFn(_n2_, _e2_, _p_)</emu-eqn>, such that <emu-eqn>_r2_ ≥ _x_</emu-eqn> and _r2_ is minimized.
+          1. [declared="n1,e1,r1"] Let _n1_ and _e1_ each be an integer and _r1_ a mathematical value, with <emu-eqn>_r1_ = ToRawPrecisionFn(_n1_, _e1_, _p_)</emu-eqn>, such that <emu-eqn>_r1_ ≤ _x_</emu-eqn> and _r1_ is maximized.
+          1. [declared="n2,e2,r2"] Let _n2_ and _e2_ each be an integer and _r2_ a mathematical value, with <emu-eqn>_r2_ = ToRawPrecisionFn(_n2_, _e2_, _p_)</emu-eqn>, such that <emu-eqn>_r2_ ≥ _x_</emu-eqn> and _r2_ is minimized.
```

sffc:
Reviewers: please double-check that the diff in the PR corresponds to the diff in the proposal. I will also double-check this but it's good to have extra verification of that.

spec/pluralrules.html
```diff
@@ -347,7 +385,38 @@ <h1>ResolvePlural ( _pluralRules_, _n_ )</h1>
...
+        1. Let _p_ be ! PluralRuleSelect(_locale_, _type_, _n_, _operands_).
+        1. Return the Record { [[PluralCategory]]: _p_, [[FormattedString]]: _s_ }.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-pluralruleselectrange" aoid="PluralRuleSelectRange">
+      <h1>PluralRuleSelectRange ( _locale_, _type_, _xp_, _yp_ )</h1>
+
+      <p>
+        When the PluralRuleSelectRange abstract operation is called with four arguments, it performs an implementation-dependent algorithm to map the plural form String values _xp_ and _yp_, representing the plural forms of the start and end of a range, to a resolved String value for the plural form of the range as a whole denoted by _type_ for the corresponding _locale_, or the String value *"other"*.
```
gibson042, line 397:
```suggestion
        When the PluralRuleSelectRange abstract operation is called with four arguments, it performs an implementation-dependent algorithm to map the <emu-xref href="#sec-pluralruleselect">plural category</emu-xref> String values _xp_ and _yp_, respectively characterizing the start and end of a range, to a resolved String value for the plural form of the range as a whole denoted by _type_ (which must be *"cardinal"* or *"ordinal"*) for the corresponding _locale_, or the String value *"other"*.
```

But I think it would make more sense to name the { [[PluralCategory]], [[FormattedString]] } Record type and pass instances of _that_ to PluralRuleSelectRange (rather than just the [[PluralCategory]] field).

spec/pluralrules.html
```diff
@@ -347,7 +385,38 @@ <h1>ResolvePlural ( _pluralRules_, _n_ )</h1>
...
+      </p>
+    </emu-clause>
+
+    <emu-clause id="sec-resolvepluralrange" aoid="ResolvePluralRange">
+      <h1>ResolvePluralRange ( _pluralRules_, _x_, _y_ )</h1>
+      <p>
+        When the ResolvePluralRange abstract operation is called with arguments _pluralRules_ (which must be an object initialized as a PluralRules), _x_ (which must be a Number value), and _y_ (which must be a Number value), it returns a String value representing the plural form of the range starting from _x_ and ending at _y_ according to the effective locale and the options of _pluralRules_. The following steps are taken:
```
gibson042, line 404: Why does ResolvePluralRange return a plural category String value rather than e.g. a { [[PluralCategory]], [[FormattedStartString]], [[FormattedEndString]] } Record to better align with ResolvePlural?

spec/pluralrules.html
```diff
@@ -140,7 +140,24 @@ <h1>Intl.PluralRules.prototype.select ( _value_ )</h1>
         1. Let _pr_ be the *this* value.
         1. Perform ? RequireInternalSlot(_pr_, [[InitializedPluralRules]]).
         1. Let _n_ be ? ToNumber(_value_).
-        1. Return ! ResolvePlural(_pr_, _n_).
+        1. Return ! ResolvePlural(_pr_, _n_).[[PluralCategory]].
```
gibson042, line 143: [PartitionRelativeTimePattern](https://tc39.es/ecma402/#sec-PartitionRelativeTimePattern) needs a similar update.

gibson042 (changes requested):
I still need to review the bulk of this PR in numberformat.html, but I have some comments on the changes that touch pluralrules.html.

spec/numberformat.html
```diff
@@ -1251,5 +1506,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+        1. Let _parts_ be ? PartitionNumberRangePattern(_numberFormat_, _x_, _y_).
+        1. Let _result_ be the empty String.
+        1. For each element _part_ in _parts_, do
+          1. Set _result_ to the string-concatenation of _result_ and _part_.[[Value]].
+        1. Return _result_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-formatnumericrangetoparts" aoid="FormatNumericRangeToParts">
+      <h1>FormatNumericRangeToParts ( _numberFormat_, _x_, _y_ )</h1>
```
ryzokuken, line 1797: Can we switch to the structured headers syntax? It doesn't necessarily need to be done before this is merged, but let's do it sooner rather than later.

This applies to every new AO, I suppose.

spec/numberformat.html
```diff
@@ -81,37 +81,46 @@ <h1>InitializeNumberFormat ( _numberFormat_, _locales_, _options_ )</h1>
...
-          _mxfdDefault_: a Number,
-          _notation_: a String,
-        )
-      </h1>
-      <dl class="header">
-        <dt>description</dt>
-        <dd>It populates the internal slots of _intlObj_ that affect locale-independent number rounding (see <emu-xref href="#sec-formatnumberstring"></emu-xref>).</dd>
-      </dl>
+    <emu-clause id="sec-setnfdigitoptions" aoid="SetNumberFormatDigitOptions">
+      <h1>SetNumberFormatDigitOptions ( _intlObj_, _options_, _mnfdDefault_, _mxfdDefault_, _notation_ )</h1>
```
ryzokuken, line 100: Is there an important reason to move away from the structured headers syntax here? Since the new description doesn't include the "type" information unlike the previous style, we're just losing that information...

spec/numberformat.html
```diff
@@ -81,37 +81,46 @@ <h1>InitializeNumberFormat ( _numberFormat_, _locales_, _options_ )</h1>
...
         1. Let _mxsd_ be ? Get(_options_, *"maximumSignificantDigits"*).
         1. Set _intlObj_.[[MinimumIntegerDigits]] to _mnid_.
+        1. Let _roundingPriority_ be ? GetOption(_options_, *"roundingPriority"*, ~string~, &laquo; *"auto"*, *"morePrecision"*, *"lessPrecision"* &raquo;, *"auto"*).
+        1. Let _roundingIncrement_ be ? GetNumberOption(_options_, *"roundingIncrement"*, 1, 5000, 1).
+        1. If _roundingIncrement_ is not in « 1, 2, 5, 10, 20, 25, 50, 100, 200, 250, 500, 1000, 2000, 2500, 5000 », throw a *RangeError* exception.
+        1. Let _roundingMode_ be ? GetOption(_options_, *"roundingMode"*, ~string~, &laquo; *"ceil"*, *"floor"*, *"expand"*, *"trunc"*, *"halfCeil"*, *"halfFloor"*, *"halfExpand"*, *"halfTrunc"*, *"halfEven"* &raquo;, *"halfExpand"*).
+        1. Let _trailingZeroDisplay_ be ? GetOption(_options_, *"trailingZeroDisplay"*, ~string~, &laquo; *"auto"*, *"stripIfInteger"* &raquo;, *"auto"*).
+        1. NOTE: All fields required by SetNumberFormatDigitOptions have now been read from _options_. The remainder of this AO interprets the options and may throw exceptions.
```
ryzokuken, line 119: Don't feel strongly either way, but not entirely convinced that this NOTE is necessary. That said, if it helps implementers enough to justify keeping it in, why not

spec/numberformat.html
```diff
@@ -120,17 +129,21 @@ <h1>
           1. Let _hasFd_ be *true*.
         1. Else,
           1. Let _hasFd_ be *false*.
-        1. Let _needSd_ be _hasSd_.
-        1. If _hasSd_ is *true*, or _hasFd_ is *false* and _notation_ is *"compact"*, then
-          1. Let _needFd_ be *false*.
-        1. Else,
-          1. Let _needFd_ be *true*.
+        1. Let _needSd_ be *true*.
```
ryzokuken, line 132: Full disclosure: this entire `needs*d`/`has*d` part of the algorithm always confuses me so for the time being I'd trust @gibson042 to review it.

spec/numberformat.html
```diff
@@ -458,18 +535,125 @@ <h1>Properties of Intl.NumberFormat Instances</h1>
...
+      <li>[[MinimumSignificantDigits]] and [[MaximumSignificantDigits]] are positive integer Number values indicating the minimum and maximum fraction digits to be shown. If present, the formatter uses however many fraction digits are required to display the specified number of significant digits. These properties are only used when [[RoundingType]] is ~significantDigits~, ~morePrecision~, or ~lessPrecision~.</li>
+      <li>[[UseGrouping]] is a Boolean or String value indicating the conditions under which a grouping separator should be used. The positions of grouping separators, and whether to display grouping separators for a given number, is implementation-defined. A value *"always"* hints the implementation to display grouping separators if possible; *"min2"*, if there are at least 2 digits in a group; *"auto"*, if the locale prefers to use grouping separators for the number. A value *false* disables grouping separators.</li>
```
ryzokuken, line 540: Did you leave out the stringified booleans on purpose?

spec/numberformat.html
```diff
@@ -1251,5 +1506,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-formatapproximately" aoid="FormatApproximately">
+      <h1>FormatApproximately ( _numberFormat_, _result_ )</h1>
+      <p>
+        The FormatApproximately abstract operation modifies _result_, which must be a List of Record values as described in PartitionNumberPattern, by adding a new Record for the approximately sign, which may depend on _numberFormat_ (which must be an object initialized as NumberFormat). The following steps are taken:
+      </p>
+      <emu-alg>
+        1. Let _i_ be an index into _result_, determined by an implementation-defined algorithm based on _numberFormat_ and _result_.
```
ryzokuken, line 1757: nit: could we add a bit more info to guide implementations here at all?

spec/numberformat.html
```diff
@@ -1251,5 +1506,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+    </emu-clause>
+
+    <emu-clause id="sec-applyunsignedroundingmode" aoid="ApplyUnsignedRoundingMode">
+      <h1>ApplyUnsignedRoundingMode ( _x_, _r1_, _r2_, _unsignedRoundingMode_ )</h1>
+      <p>
+        The abstract operation ApplyUnsignedRoundingMode considers _x_ (a mathematical value), bracketed below by _r1_ (a mathematical value) and above by _r2_ (a mathematical value), and returns either _r1_ or _r2_ according to _unsignedRoundingMode_ (a specification type from the *Unsigned Rounding Mode* column of <emu-xref href="#table-intl-unsigned-rounding-modes"></emu-xref> or *undefined*). The following steps are taken:
+      </p>
+      <emu-alg>
+        1. If _x_ is equal to _r1_, return _r1_.
+        1. Assert: _r1_ &lt; _x_ &lt; _r2_.
```
ryzokuken, line 1706: TIL about `emu-eqn`! It's very nifty, could it be used for these comparisons as well?

ryzokuken (approved):
Overall the spec looks great, I left a few comments inline but most of them are minor and some could even be addressed later on.

spec/negotiation.html
```diff
@@ -353,7 +353,31 @@ <h1>
...
           1. Set _value_ to ? ToString(_value_).
-        1. If _values_ is not ~empty~ and _values_ does not contain an element equal to _value_, throw a *RangeError* exception.
+        1. If _values_ is not ~empty~ and _values_ does not contain _value_, throw a *RangeError* exception.
+        1. Return _value_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-getbooleanorstringnumberformatoption" type="abstract operation">
+      <h1>
+        GetBooleanOrStringNumberFormatOption (
```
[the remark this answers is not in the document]
gibson042: It was actually renamed _from_ GetStringOrBooleanOption at my request because the logic should not be generalized beyond NumberFormat; see https://github.com/tc39/proposal-intl-numberformat-v3/pull/125 and https://github.com/tc39/proposal-intl-numberformat-v3/pull/132 .

spec/numberformat.html
```diff
@@ -347,6 +406,12 @@ <h1>Intl.NumberFormat.prototype.resolvedOptions ( )</h1>
           1. Let _v_ be the value of _nf_'s internal slot whose name is the Internal Slot value of the current row.
           1. If _v_ is not *undefined*, then
             1. Perform ! CreateDataPropertyOrThrow(_options_, _p_, _v_).
+        1. If _nf_.[[RoundingType]] is ~morePrecision~, then
+          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"morePrecision"*).
+        1. Else if _nf_.[[RoundingType]] is ~lessPrecision~, then
+          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"lessPrecision"*).
+        1. Else,
+          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"auto"*).
```
gibson042, line 414: Would it make more sense to just introduce a [[RoundingPriority]] slot and avoid the potential of this recovery logic falling out of sync in the future (and also the minor weirdness of "roundingPriority" being required to appear as the last property of the object returned from `resolvedOptions()`)?

spec/numberformat.html
gibson042: > Did you leave out the stringified booleans on purpose?

They are just a quirk of value determination and are never stored in the slot.

spec/numberformat.html
```diff
@@ -458,18 +535,125 @@ <h1>Properties of Intl.NumberFormat Instances</h1>
...
+            <td>Ties toward zero</td>
+            <td>➕ [-1]</td>
+            <td>➖ [0]</td>
+            <td>➖ [0]</td>
+            <td>➕ [1]</td>
+            <td>➖ [1]</td>
+          </tr>
+          <tr>
+            <td>*"halfEven"*</td>
+            <td>Ties toward even cardinality</td>
```
gibson042, line 645: Even cardinality of what?
```suggestion
            <td>Ties toward an even rounding increment multiple</td>
```

spec/numberformat.html
```diff
@@ -458,18 +535,125 @@ <h1>Properties of Intl.NumberFormat Instances</h1>
...
+            <td>➖ [-2]</td>
+            <td>➖ [0]</td>
+            <td>➖ [0]</td>
+            <td>➕ [1]</td>
+            <td>➕ [2]</td>
+          </tr>
+        </tbody>
+      </table>
+      <emu-note>The examples are illustrative of the unique behaviour of each option. ➕ means "resolves toward positive infinity"; ➖ means "resolves toward negative infinity".</emu-note>
```
gibson042, line 654: I appreciate the information, but I wonder if  and  collide too much with actual numeric signs. Did you consider arrows instead?
* ↥ and ↧
* ⤒ and ⤓
* ⬆ and ⬇️
* ⭱ and ⭳
* ⮉ and ⮋
* ⇧ and ⇩
* etc.

gibson042 (commented):
I agree with @ryzokuken , and have a few nits/suggestions but everything overall looks good.

spec/negotiation.html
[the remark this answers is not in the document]
sffc: I'll resolve this because it was discussed in the previous issue @gibson042 linked above.

spec/numberformat.html
sffc: May have gotten lost as part of the merge. Will fix.

spec/numberformat.html
sffc: I added this line as part of resolving an editorial issue back on the spec repository where I was reading options in a bunch of different places. I think the NOTE could be useful for future changes in order to ensure that we are consistent in the point at which we read the options.

spec/numberformat.html
```diff
@@ -143,12 +156,24 @@ <h1>
...
+            1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
+          1. Else if _roundingPriority_ is *"lessPrecision"*, then
+            1. Set _intlObj_.[[RoundingType]] to ~lessPrecision~.
+          1. Else if _hasSd_ is *true*, then
+            1. Set _intlObj_.[[RoundingType]] to ~significantDigits~.
+          1. Else,
+            1. Set _intlObj_.[[RoundingType]] to ~fractionDigits~.
         1. Else,
-          1. Set _intlObj_.[[RoundingType]] to ~fractionDigits~.
+          1. Set _intlObj_.[[RoundingType]] to ~morePrecision~.
```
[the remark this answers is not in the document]
sffc: Done

spec/numberformat.html
sffc: Would rather not make this big of a change at this point. Also, this logic reflects that RoundingPriority is now a computed property: if compact notation is used, it gets computed as "auto", which is needed for round-tripping.

spec/numberformat.html
sffc: Arrow emoji SGTM

spec/numberformat.html
sffc: Added a sentence

spec/numberformat.html
```diff
@@ -1251,5 +1506,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+        1. Let _approximatelySign_ be an ILND String value used to signify that a number is approximate.
+        1. If _approximatelySign_ is not empty, insert a new Record { [[Type]]: *"approximatelySign"*, [[Value]]: _approximatelySign_ } at index _i_ in _result_.
+        1. Return _result_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-collapsenumberrange" aoid="CollapseNumberRange">
+      <h1>CollapseNumberRange ( _result_ )</h1>
+      <p>
+        The CollapseNumberRange abstract operation modifies _result_, which must be a List of Record values as described in PartitionNumberRangePattern, by removing redundant information, and returns the resulting List. The algorithm is implementation dependent, but it must not introduce ambiguity: the resulting list after modification should be unique for all pairs _x_ and _y_.
```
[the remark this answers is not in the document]
sffc: Done

spec/numberformat.html
sffc: Acknowledged. I wrote most of this spec before structured headers were common in ECMA-402. In an effort to get this merged, I would rather make only small changes that are unlikely to require extensive further review. However, this sounds like a great editorial task to be done after merging this PR. :)

spec/pluralrules.html
sffc: Great catch; fixed

spec/pluralrules.html
sffc: I think because I don't need the extra info from `ResolvePluralRange`; I changed `ResolvePlural` to return a Record only because I needed the formatted string to perform some other operation.

spec/numberformat.html
```diff
@@ -1251,5 +1509,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+      </p>
+      <p>
+        For example, an implementation may choose to remove the currency symbol following the range separator, to produce "$3–5" instead of "$3–$5".
+      </p>
+      <p>
+        The implementation may choose to modify symbols for grammatical correctness; for example, "0.5 miles–1 mile" may become "0.5–1 miles".
+      </p>
+      <p>
+        Returning _result_ unmodified is guaranteed to be a correct implementation of CollapseNumberRange.
+      </p>
```
gibson042, line 1780: I think this still fails to truly satisfy [@ryzokuken's request](https://github.com/tc39/ecma402/pull/753#discussion_r1157361429), in particular because _start_ and _end_ have no referents here.
```suggestion
      <p>
        The CollapseNumberRange abstract operation modifies _result_ (which must be a List of { [[Type]], [[Value]], [[Source]] } Records as constructed within PartitionNumberRangePattern) by removing redundant information and resolving internal inconsistency, and returns the resulting List. The algorithm is implementation dependent, but must not introduce ambiguity that would cause the result of <emu-xref href="#sec-intl.numberformat.prototype.formatrange" title></emu-xref> with arguments List « _start1_, _end1_ » to equal the result with arguments List « _start2_, _end2_ » if the results for those same arguments Lists would not be equal with a trivial implementation of CollapseNumberRange that always returns _result_ unmodified.
      </p>
      <p>
        For example, an implementation may remove the Record representing a currency symbol after a range separator to convert a _results_ List representing *"$3–$5"* into one representing *"$3–5"*.
      </p>
      <p>
        An implementation may also modify Record [[Value]] fields for grammatical correctness; for example, converting a _results_ List representing *"0.5 miles–1 mile"* into one representing *"0.5–1 miles"*.
      </p>
      <p>
        Returning _result_ unmodified is guaranteed to be a correct implementation of CollapseNumberRange.
      </p>
```
I have ideas for further improvement as well, but those depend upon generalizing the common "_concatenate a List of Records by their [[Value]] field_" behavior into a new operation and can wait for a later editorial PR.

spec/numberformat.html
gibson042: @ryzokuken Thoughts on the above? I can live with it, albeit unhappily.

gibson042 (changes requested):
A few final nits, with the one applying to PartitionNumberRangePattern being a blocking concern (but easy to resolve).

spec/numberformat.html
```diff
@@ -1251,5 +1509,313 @@ <h1>ComputeExponentForMagnitude ( _numberFormat_, _magnitude_ )</h1>
...
+      <h1>PartitionNumberRangePattern ( _numberFormat_, _x_, _y_ )</h1>
+      <p>
+        The abstract operation PartitionNumberRangePattern creates the parts for a localized number range according to _x_ (which must be an Intl mathematical value), _y_ (which must be an Intl mathematical value), and the formatting options of _numberFormat_ (which must be an object initialized as NumberFormat). The following steps are taken:
+      </p>
+      <emu-alg>
+        1. If _x_ is ~not-a-number~ or _y_ is ~not-a-number~, throw a *RangeError* exception.
+        1. Let _result_ be a new empty List.
+        1. Let _xResult_ be ? PartitionNumberPattern(_numberFormat_, _x_).
+        1. Let _yResult_ be ? PartitionNumberPattern(_numberFormat_, _y_).
+        1. If _xResult_ is equal to _yResult_, then
```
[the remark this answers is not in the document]
sffc: ^ Yeah, if comparing the lists is ill-defined, what you wrote here is basically equivalent

sffc:
Thanks for the suggestions; they all looked good to me

ryzokuken:
> @ryzokuken Thoughts on the above? I can live with it, albeit unhappily.

@gibson042 this could be addressed by a follow-up editorial change, right? I want to kick off the opt-out ASAP and then we could clean up the working draft.

gibson042:
> this could be addressed by a follow-up editorial change, right? I want to kick off the opt-out ASAP and then we could clean up the working draft.

It would only be editorial if we commit to the awkward property ordering: …, `roundingMode`, `roundingIncrement`, `trailingZeroDisplay`, `roundingPriority`.

gibson042 (approved):

PR #754, opened 2023-02-15
editorial: CreatePartsFromList type change

ryzokuken:
Change the expected type of the "list" argument for the CreatePartsFromList abstract operation from a List of String values to a simple List. This makes no outside observable normative change but allows other built-ins to use this operation in more interesting ways, primarily motivated by `Intl.DurationFormat` which utilizes this operation in order to format the list of different duration units.

Refs: https://github.com/tc39/proposal-intl-duration-format/issues/103

### DIFF
--- a/spec/listformat.html
+++ b/spec/listformat.html
@@ -281,7 +281,7 @@
       <h1>CreatePartsFromList ( _listFormat_, _list_ )</h1>
 
       <p>
-        The CreatePartsFromList abstract operation is called with arguments _listFormat_ (which must be an object initialized as a ListFormat) and _list_ (which must be a List of String values), and creates the corresponding list of parts according to the effective locale and the formatting options of _listFormat_.
+        The CreatePartsFromList abstract operation is called with arguments _listFormat_ (which must be an object initialized as a ListFormat) and _list_ (which must be a List), and creates the corresponding list of parts according to the effective locale and the formatting options of _listFormat_.
         Each part is a Record with two fields: [[Type]], which must be a string with values "element" or "literal", and [[Value]] which must be a string or a number.
         The following steps are taken:
       </p>


sffc (commented):
Seems reasonable; the only place where we _might_ care about it being a string is in this line:

>           1. Let _n_ be an implementation-defined index into _listFormat_.[[Templates]] based on _listFormat_.[[Locale]], _head_, and _parts_.

The "implementation-defined" algorithm could previously assume that the values were strings, but it can no longer make that assumption. However, this is already hand-wavy enough that it's probably fine.

ryzokuken:
Yeah, you're spot on. I thought about this deeply and based on how it is specced, the best we could do it further hand-wave to talk more specifically about this case but not much more I'm afraid. There's nothing quite programmatic that would ensure that things work as expected. Perhaps a note for implementers would help? @FrankYFTang @dminor any feedback?

FrankYFTang:
1. Changing from a List of String to a List of anything is a very big jump. Do you want to go somewhere in the middle? Do the type, if not only String, need to be in the same type? 
2. Currently in https://tc39.es/ecma402/#sec-createpartsfromlist
"... creates the corresponding list of parts according to the effective locale and the formatting options of listFormat. Each part is a [Record](https://tc39.es/ecma262/#sec-list-and-record-specification-type) with two fields: [[Type]], which must be a string with values "element" or "literal", and [[Value]] which must be a string or a number."

Notice the output is specified to be a list of parts. And each part is a Record with a Type and an Value. And the Value must be either a string or a number. 
Now, if we read the algorithm, the reason the type of the Value is from either 
a. the pattern  (If part is "literal", then) or
b. placeables.[[<part>]] 
and placeables came from list

And the reason in the return, the part which " [[Value]] which must be a string or a number." is because the list is limited to list of String, if we chnage it to any type, then that statement (" [[Value]] which must be a string or a number") cannot hold true.

gibson042 (changes requested):
Hmm, I don't think I like this. It would be preferable for [PartitionDurationFormatPattern](https://tc39.es/proposal-intl-duration-format/#sec-partitiondurationformatpattern) to just be more clear about conversions.

ryzokuken:
> Hmm, I don't think I like this. It would be preferable for [PartitionDurationFormatPattern](https://tc39.es/proposal-intl-duration-format/#sec-partitiondurationformatpattern) to just be more clear about conversions.

gibson042:
To be honest, I'm really not understanding the mechanics of [PartitionDurationFormatPattern](https://tc39.es/proposal-intl-duration-format/#sec-partitiondurationformatpattern), especially where it comes to manipulating the `_last_` element of `_result_`. I suspect that refactoring that with the goal of improving comprehensibility would clarify this topic as well.

But from what I _do_ understand, this looks like a map–process–unmap-style functional programming problem serving a purpose that is very much analogous to [PartitionNotationSubPattern](https://tc39.es/ecma402/#sec-partitionnotationsubpattern), in which case I agree with the principle of the change but not with its current manifestation. I would really like to see a general solution that covers that and cleans up the related [MakePartsList](https://tc39.es/ecma402/#sec-makepartslist)/[PartitionPattern](https://tc39.es/ecma402/#sec-partitionpattern)/[DeconstructPattern](https://tc39.es/ecma402/#sec-deconstructpattern) mess. I think I want to end up with a ListFormat-related operation that accepts a List in which each element has a String representation (_perhaps as a Record with a [[StringValue]] field, or perhaps by means of an Abstract Closure_) that is used by ILD steps and returns a new List suitable for the various consumers, possibly pre-flattened... it's not far off from what is here now, but is both more general in scope and more specific in input/output signature. We could even tackle it incrementally, in which case the change needed here would be requiring the CreatePartsFromList `_list_` input to be a List in which each element is a _Record_ with a [[StringValue]] field and updating FormatList{,ToParts} operations to either provide that directly or to use a new wrapping operation such as CreatePartsFromStringList.
```diff
-      <h1>CreatePartsFromList ( _listFormat_, _list_ )</h1>
-
-      <p>
-        The CreatePartsFromList abstract operation is called with arguments _listFormat_ (which must be an object initialized as a ListFormat) and _list_ (which must be a List), and creates the corresponding list of parts according to the effective locale and the formatting options of _listFormat_.
-        Each part is a Record with two fields: [[Type]], which must be a string with values "element" or "literal", and [[Value]] which must be a string or a number.
-        The following steps are taken:
-      </p>
+      <h1>
+        CreatePartsFromList(
+          _listFormat_: an object that has a [[InitializedListFormat]] slot,
+          _list_: a List of Records in which each element has a string-valued [[StringValue]] field,
+        ): a List of ListPart Records
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
+        <dd>
+          It returns a List of ListPart Records representing _list_ as a formatted list according to the effective locale and the formatting options of _listFormat_.
+          Each <dfn variants="ListPart Records">ListPart Record</dfn> is a Record with a [[Type]] field and a [[Value]] field.
+          When the value of the [[Type]] field is *"element"*, the value of the [[Value]] field is an element of _list_.
+          When the value of the [[Type]] field is *"literal"*, the value of the [[Value]] field is a String representing literal text constituting the list formatting.
+        </dd>
+      </dl>
       <emu-alg>
         1. Let _size_ be the number of elements of _list_.
         1. If _size_ is 0, then
           1. Return a new empty List.
         1. If _size_ is 2, then
-          1. Let _n_ be an index into _listFormat_.[[Templates]] based on _listFormat_.[[Locale]], _list_[0], and _list_[1].
+          1. Let _n_ be an index into _listFormat_.[[Templates]] based on _listFormat_.[[Locale]], _list_[0].[[StringValue]], and _list_[1].[[StringValue]].
           1. Let _pattern_ be _listFormat_.[[Templates]][_n_].[[Pair]].
           1. Let _first_ be a new Record { [[Type]]: *"element"*, [[Value]]: _list_[0] }.
           1. Let _second_ be a new Record { [[Type]]: *"element"*, [[Value]]: _list_[1] }.
           1. Let _placeables_ be a new Record { [[0]]: _first_, [[1]]: _second_ }.
           1. Return ! DeconstructPattern(_pattern_, _placeables_).
         1. Let _last_ be a new Record { [[Type]]: *"element"*, [[Value]]: _list_[_size_ - 1] }.
         1. Let _parts_ be &laquo; _last_ &raquo;.
         1. Let _i_ be _size_ - 2.
         1. Repeat, while _i_ &ge; 0,
-          1. Let _head_ be a new Record { [[Type]]: *"element"*, [[Value]]: _list_[_i_] }.
-          1. Let _n_ be an implementation-defined index into _listFormat_.[[Templates]] based on _listFormat_.[[Locale]], _head_, and _parts_.
+          1. Let _element_ be _list_[_i_].
+          1. Let _n_ be an implementation-defined index into _listFormat_.[[Templates]] based on _listFormat_.[[Locale]], _element_.[[StringValue]], and _parts_.
           1. If _i_ is 0, then
             1. Let _pattern_ be _listFormat_.[[Templates]][_n_].[[Start]].
           1. Else if _i_ is less than _size_ - 2, then
             1. Let _pattern_ be _listFormat_.[[Templates]][_n_].[[Middle]].
           1. Else,
             1. Let _pattern_ be _listFormat_.[[Templates]][_n_].[[End]].
+          1. Let _head_ be a new Record { [[Type]]: *"element"*, [[Value]]: _element_ }.
           1. Let _placeables_ be a new Record { [[0]]: _head_, [[1]]: _parts_ }.
           1. Set _parts_ to ! DeconstructPattern(_pattern_, _placeables_).
-          1. Decrement _i_ by 1.
+          1. Set _i_ to _i_ - 1.
         1. Return _parts_.
       </emu-alg>
```

ryzokuken:
Closing for now in favor of https://github.com/tc39/proposal-intl-duration-format/pull/134

PR #758, opened 2023-03-13
Normative: Change the hourCycle default logic

FrankYFTang:
While hour12 is  false, let hourCycle to be alwasyt 'h23' but not 'h24'.

The current logic is not reasonable. We see no region in the CLDR use  h24 hour cycle as default.  While we set a hour12: false on a 12 hour system region (which use 1:00 - 12:59), we should set to h23 instead of h24

We need to make this change
so while hour12 is false on 'en' locale, it will be set to h23 instead of h24
but also while hour12 is true on 'ja' locale, it will be set to h11 but not h12.

Address https://github.com/tc39/ecma402/issues/402

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -205,7 +205,7 @@
           [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle]] must be a String value equal to *"h11"*, *"h12"*, *"h23"*, or *"h24"*.
         </li>
         <li>
-          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle12]] must be a String value equal to *"h11"*, *"h12"*.
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle12]] must be a String value equal to *"h11"* or *"h12"*.
         </li>
         <li>
           [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle24]] must be a String value equal to *"h23"*, or *"h24"*.
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -77,7 +77,7 @@
         1. Let _dataLocale_ be _r_.[[dataLocale]].
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
         1. If _hour12_ is *true*, then
-          1. Let _hc_ to _dataLocaleData_.[[hourCycle12]].
+          1. Let _hc_ be _dataLocaleData_.[[hourCycle12]].
         1. Else if _hour12_ is *false*, then
           1. Let _hc_ be _dataLocaleData_.[[hourCycle24]].
         1. Else,
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -76,15 +76,14 @@
         1. Set _dateTimeFormat_.[[NumberingSystem]] to _r_.[[nu]].
         1. Let _dataLocale_ be _r_.[[dataLocale]].
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
-        1. Let _hcDefault_ be _dataLocaleData_.[[hourCycle]].
         1. If _hour12_ is *true*, then
-          1. If _hcDefault_ is *"h11"* or *"h23"*, let _hc_ be *"h11"*. Otherwise, let _hc_ be *"h12"*.
+          1. Let _hc_ to _dataLocaleData_.[[hourCycle12]].
         1. Else if _hour12_ is *false*, then
-          1. Let _hc_ be *"h23"*.
+          1. Let _hc_ to _dataLocaleData_.[[hourCycle24]].
         1. Else,
           1. Assert: _hour12_ is *undefined*.
           1. Let _hc_ be _r_.[[hc]].
-          1. If _hc_ is *null*, set _hc_ to _hcDefault_.
+          1. If _hc_ is *null*, set _hc_ to _dataLocaleData_.[[hourCycle]].
         1. Set _dateTimeFormat_.[[HourCycle]] to _hc_.
         1. Let _timeZone_ be ? Get(_options_, *"timeZone"*).
         1. If _timeZone_ is *undefined*, then
@@ -204,6 +203,12 @@
         </li>
         <li>
           [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle]] must be a String value equal to *"h11"*, *"h12"*, *"h23"*, or *"h24"*.
+        </li>
+        <li>
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle12]] must be a String value equal to *"h11"*, *"h12"*.
+        </li>
+        <li>
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle24]] must be a String value equal to *"h23"*, or *"h24"*.
         </li>
         <li>
           [[LocaleData]].[[&lt;_locale_&gt;]] must have a [[formats]] field. This [[formats]] field must be a Record with [[&lt;_calendar_&gt;]] fields for all calendar values _calendar_. The value of this field must be a list of records, each of which has a subset of the fields shown in <emu-xref href="#table-datetimeformat-components"></emu-xref>, where each field must have one of the values specified for the field in <emu-xref href="#table-datetimeformat-components"></emu-xref>. Multiple records in a list may use the same subset of the fields as long as they have different values for the fields. The following subsets must be available for each locale:
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -76,15 +76,15 @@
         1. Set _dateTimeFormat_.[[NumberingSystem]] to _r_.[[nu]].
         1. Let _dataLocale_ be _r_.[[dataLocale]].
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
+        1. Let _hcDefault_ be _dataLocaleData_.[[hourCycle]].
         1. If _hour12_ is *true*, then
-          1. Let _hc_ be *"h12"*.
+          1. If _hcDefault_ is *"h11"* or *"h23"*, let _hc_ be *"h11"*. Otherwise, let _hc_ be *"h12"*.
         1. Else if _hour12_ is *false*, then
           1. Let _hc_ be *"h23"*.
         1. Else,
           1. Assert: _hour12_ is *undefined*.
           1. Let _hc_ be _r_.[[hc]].
-          1. If _hc_ is *null*, then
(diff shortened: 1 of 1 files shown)


FrankYFTang:
notice grep "<hours.*allowed" ./cldr/common/supplemental/supplementalData.xml show possibility of h H or K but no possibility of k in the latest CLDR and it is unlikely that will happen in the future from my understanding.

ryzokuken (approved):

spec/datetimeformat.html
```diff
@@ -80,7 +80,7 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
         1. If _hour12_ is *true*, then
           1. If _hcDefault_ is *"h11"* or *"h23"*, let _hc_ be *"h11"*. Otherwise, let _hc_ be *"h12"*.
         1. Else if _hour12_ is *false*, then
-          1. If _hcDefault_ is *"h11"* or *"h23"*, let _hc_ be *"h23"*. Otherwise, let _hc_ be *"h24"*.
+          1. Let _hc_ be *"h23"*.
```
sffc, line 83: Why are you changing it here but not 2 lines above?

sffc, line 83: Would better logic be,

```
1. If _hour12_ is *true*, then
  1. If _hcDefault_ is *"h23"* or *"h24"*, let _hc_ be *"h12"*. Otherwise, let _hc_ be _hcDefault_.
1. Else if _hour12_ is *false*, then
  1. If _hcDefault_ is *"h11"* or *"h12"*, let _hc_ be *"h23"*. Otherwise, let _hc_ be _hcDefault_.
```

sffc (changes requested):

spec/datetimeformat.html
FrankYFTang: because if I change the 2 line above, then ja will not work
see https://github.com/tc39/ecma402/issues/402#issuecomment-623628320
see the following code
```js
(new Intl.DateTimeFormat("ja", {hour: "numeric", hour12: true})).resolvedOptions().hourCycle;
// h11
(new Intl.DateTimeFormat("en", {hour: "numeric", hour12: true})).resolvedOptions().hourCycle
// h12
(new Intl.DateTimeFormat("fr", {hour: "numeric", hour12: true})).resolvedOptions().hourCycle
// h12
```
and this should be kept.

FrankYFTang: this will not work for
```js
(new Intl.DateTimeFormat("ja", {hour: "numeric", hour12: true})).resolvedOptions().hourCycle;
```
_hcDefault_ for ja is h23 but if hour12:true it should be h11, not h23 nor h12
your logic will make it h23.
also there are more problem for the case if _hour12_ is *false*

sffc: Can you cite the CLDR data that shows that "ja" in 12-hour mode is expected to use h11, and likewise any locale in 24-hour mode that is expected to use h24?

FrankYFTang: > Can you cite the CLDR data that shows that "ja" in 12-hour mode is expected to use h11, and likewise any locale in 24-hour mode that is expected to use h24?

That is not what my PR proposed.

What my PR proposed is 
the following mapping

| hour12  | undefined | false  |   true |  
|---|---|---|---| 
| h11  |  h11  | h23 |  h11  | 
| h12 | h12  |  h23* |  h12  |
| h23  | h23  |  h23 |  h11 |
| h24  |  h24  |  h24 |  h12  |

The differences from the status quo is while the hour cycle is h12 and hour12:false  instead of making it h24 (which show 24:30:59) we show (00:30:59) and that impact mostly "en-US" or "en-CA" , "fr-CA" locales

FrankYFTang: about japanese, I am not sure this count as what you asked
https://github.com/unicode-org/cldr/blob/main/common/main/ja.xml#L3059
```xml
						<dateFormatItem id="h">aK時</dateFormatItem>
						<dateFormatItem id="H">H時</dateFormatItem>
```
h-> aK =>  use h11 when asking for not in 24 hour mode
H->H => use h23 when asking for in 24 hour mode

sffc: Thanks for the Japanese citation

I'm not super happy with the spec codifying this table. The only restriction should be that you get either h11 or h12 in 12-hour mode or either h23 or h24 in 24-hour mode. Which one should be ILD.

FrankYFTang: I understand why you are not happen with the spec codifying this table. 

" The only restriction should be that you get either h11 or h12 in 12-hour mode or either h23 or h24 in 24-hour mode"
YES, but that is not what the current spec text specify.  The current spec text specifiy a logic  to resolve that instead purely depend on the  CLDR to resolve that.

anba: Alternatively we could add two new fields to [LocaleData](https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots) to describe the preferred hour-cycles for 12- and 24-hour clocks.

> - [[LocaleData]].[[\<locale\>]].[[hourCycle12]] must be a String value equal to "h11" or "h12". 
> - [[LocaleData]].[[\<locale\>]].[[hourCycle24]] must be a String value equal to "h23" or "h24". 


And then rewrite steps 24-26 as:
> 25. If hour12 is true, then
>     a. Let hc be dataLocaleData.[[hourCycle12]].
> 26. Else if hour12 is false, then
>     a. Let hc be dataLocaleData.[[hourCycle24]].
> 27. Else,
>     a. Assert: hour12 is undefined.
>     b. Let hc be r.[[hc]].
>     c. If hc is null, set hc to dataLocaleData.[[hourCycle]].

FrankYFTang: > Alternatively we could add two new fields to [LocaleData](https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots) to describe the preferred hour-cycles for 12- and 24-hour clocks.
> 
> >

hum.... that will make it more a data driven approach ( which is good). Shane- what do you think?

sffc: +1 on @anba's approach.

FrankYFTang:
PTAL I took anba's suggestion and make it data driven.

sffc (approved):

sffc (approved):

ryzokuken (approved):

sffc:
TG2 agreement: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-04-06.md#normative-change-the-hourcycle-default-logic-758

FrankYFTang:
somehow no one ever present this PR to TG39 in May 2023 or July 2023 meeting. I add it to https://github.com/tc39/agendas/blob/main/2023/09.md

FrankYFTang:
we need test262 tests for this PR

jufemaiz:
I'm happy to help out with this if needed @FrankYFTang. What's the requirement here?

FrankYFTang:
@anba - any suggetion how should we write test262 test to test this PR?

anba:
> @anba - any suggetion how should we write test262 test to test this PR?

<https://github.com/tc39/test262/blob/main/test/intl402/DateTimeFormat/prototype/resolvedOptions/hourCycle-default.js> needs to be updated to reflect the PR changes. The test is already marked as locale-specific, so we could simply change it to:
```js
let locales = ["en", "fr", "it", "ja", "zh", "ko", "ar", "hi"];

for (let locale of locales) {
  let hcDefault = new Intl.DateTimeFormat(locale, {hour: "numeric"}).resolvedOptions().hourCycle;
  assert(
    hcDefault === "h11" || hcDefault === "h12" || hcDefault === "h23" || hcDefault === "h24",
    "hcDefault is one of [h11, h12, h23, h24]"
  );

  let hour12 = new Intl.DateTimeFormat(locale, {hour: "numeric", hour12: true}).resolvedOptions().hourCycle;
  assert(hour12 === "h11" || hour12 === "h12", "hour12 is one of [h11, h12]");

  let hour24 = new Intl.DateTimeFormat(locale, {hour: "numeric", hour12: false}).resolvedOptions().hourCycle;
  assert(hour24 === "h23" || hour24 === "h24", "hour24 is one of [h23, h24]");

  if (hcDefault === "h11" || hcDefault === "h12") {
    assert.sameValue(hour12, hcDefault, "hour12 matches hcDefault");
  } else {
    assert.sameValue(hour24, hcDefault, "hour24 matches hcDefault");
  }

  // 24-hour clock uses the "h23" format in all locales.
  assert.sameValue(hour24, "h23");

  // 12-hour clock uses the "h12" format in all locales except "ja".
  assert.sameValue(hour12, locale === "ja" ? "h11" : "h12");
}
```

ryzokuken:
@FrankYFTang @anba is merging this PR blocked on the tests?

ryzokuken:
2023-09-26: This PR achieved TG1 consensus

FrankYFTang:
This PR reach concensus on 2023-09-26 meeting

FrankYFTang:
I believe the PR need to merge first before merging the test, right?

FrankYFTang:
> @FrankYFTang @anba is merging this PR blocked on the tests?

The test is blocked by the merge of this PR https://github.com/tc39/test262/pull/3885

spec/datetimeformat.html
```diff
@@ -205,6 +204,12 @@ <h1>Internal slots</h1>
         <li>
           [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle]] must be a String value equal to *"h11"*, *"h12"*, *"h23"*, or *"h24"*.
         </li>
+        <li>
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle12]] must be a String value equal to *"h11"*, *"h12"*.
```
anba, line 208:
```suggestion
          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle12]] must be a String value equal to *"h11"* or *"h12"*.
```

spec/datetimeformat.html
```diff
@@ -205,6 +204,12 @@ <h1>Internal slots</h1>
...
+        </li>
+        <li>
+          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle24]] must be a String value equal to *"h23"*, or *"h24"*.
```
anba, line 211:
```suggestion
          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle24]] must be a String value equal to *"h23"* or *"h24"*.
```

anba (approved):
Two small issues in the description text, otherwise LGTM!

jufemaiz:
🎉

jufemaiz:
Can I assume that this will form part of a changeset for the 10th edition?

Ref: https://ecma-international.org/publications-and-standards/standards/ecma-402/

jufemaiz:
> https://ecma-international.org/publications-and-standards/standards/ecma-402/

FTR I believe it first was seen in v11.0 (2024-06)

Issue #81, opened 2016-02-27
Editorial: Start all fields and slots with an uppercase code point


caridy:
In order to align with 262 on this, we will have to refactor part of the spec due to the following:

> As an extension to the Record Specification Type, the notation `"[[<name>]]"` denotes a field whose name is given by the variable name, which must have a String value. For example, if a variable s has the value `"a"`, then `[[<s>]]` denotes the field `[[<a>]]`.

> For ECMAScript objects, this standard may use variable-named internal slots: The notation `"[[<name>]]"` denotes an internal slot whose name is given by the variable name, which must have a String value. For example, if a variable s has the value `"a"`, then `[[<s>]]` denotes the `[[<a>]]` internal slot.

This is very tricky, but worth the effort IMO. I might delay this work until after 3rd edition because `formatToParts` will change the way we transform the patterns for number and datetime format, which can facilitate this work.

/cc @littledan @bterlson

littledan:
If it's a 'procedurally generated' name, then I don't think this rule needs to apply the same way. However, I like the way that formatToParts changes this.

caridy:
yeah, I do have a local branch for this, but lets wait until after formatToParts() lands for 4th edition.

caridy:
hopefully 5th edition.

caridy:
These is related to #56 as well. We are almost there, and lowercase is only used in match resolvers and locale data records as far as I can tell. @anba, @littledan, do you guys think that this is realistic?

littledan:
Not sure what you are asking. The change still seems like a good idea to me.

spectranaut:
@leo @littledan I also wonder if we should keep with issue open. Now that I submitted https://github.com/tc39/ecma402/pull/401, all the remaining non-capitalized record fields are from constructing records from locale data (for example, in [resolveLocale](https://tc39.es/ecma402/#sec-resolvelocale)). Maybe it's best to simply allow non-capitalized field names in 402 because of this case.

littledan:
@spectranaut I'm glad your patch landed. I agree it wouldn't make sense to capitalize a record field like `[[nu]]`. However, I don't understand why things like `[[locale]]`, `[[dataLocale]]` and `[[extension]]` should be lower case. Was this left that way deliberately?

sffc:
Related: #339

ryzokuken:
@gibson042 this still needs to happen, right?

gibson042:
[...]

PR #760, opened 2023-03-15
Editorial: change [[localeMatcher]] to [[LocaleMatcher]] to align w/ 262

ben-allen:
(very) partial fix [#81](https://github.com/tc39/ecma402/issues/81) . this one does not seem to be exposed to [[&lt;dynamic&gt;]] access problems. submitted to get reality check on my judgement here.


spec/datetimeformat.html
```diff
@@ -203,7 +203,7 @@ <h1>Internal slots</h1>
           [[LocaleData]].[[&lt;_locale_&gt;]].[[hc]] must be &laquo; *null*, *"h11"*, *"h12"*, *"h23"*, *"h24"* &raquo;.
         </li>
         <li>
-          [[LocaleData]].[[&lt;_locale_&gt;]].[[hourCycle]] must be a String value equal to *"h11"*, *"h12"*, *"h23"*, or *"h24"*.
```
ryzokuken, line 206: This one seems to be imbalanced. I suppose there is an access to this `[[hourCycle]]` field that needs to be updated too?

ljharb: generally every commit landed to main needs to be correct, so it might need to all go in the same PR

sffc:
@ben-allen This is editorial, correct? Please say so in the PR title

ljharb:
@ben-allen why is this closed?

ben-allen:
@ljharb Apologies -- replaced this one with #766.

PR #764, opened 2023-03-22
Editorial: remove `!` from two calls to AOs that can't throw

justingrant:
This commit removes the `!` from two AO calls that return regular values, not abrupt completions.

From a quick look it seems that there are quite a few other AOs that cannot throw but still are called with `!` in the spec, so it's probably worth a full scrub at some point. For this PR I'm only worried about calls that I noticed while reviewing a Temporal PR.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -91,9 +91,9 @@
           1. Set _timeZone_ to DefaultTimeZone().
         1. Else,
           1. Set _timeZone_ to ? ToString(_timeZone_).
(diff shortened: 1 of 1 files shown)


gibson042 (approved):

gibson042:
> From a quick look it seems that there are quite a few other AOs that cannot throw but still are called with `!` in the spec, so it's probably worth a full scrub at some point.

Yes, this stems from a change in conventions that has not been fully cleaned up. But I'll still take improvements when offered!

PR #765, opened 2023-03-26
Editorial: update time zone identifier text

justingrant:
This PR includes a few minor updates to spec text about time zone identifiers that I noticed while preparing the Time Zone Canonicalization proposal:
* Fixes an inconsistency between text in [#sec-time-zone-names](https://tc39.es/ecma402/#sec-time-zone-names) (which requires IANA Time Zone Database) and text in two AOs that implies that such support is optional. This PR replaces the latter optional-implying text with the text from #sec-time-zone-names.
* Renames `_timeZone_` variable to `_timeZoneIdentifier_` when it's guaranteed to be a string.
* Replaces "name" with the more precise "identifier" in a few places.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -629,7 +629,7 @@
       </p>
 
       <emu-note>
-        In this version of the ECMAScript 2023 Internationalization API, the *"timeZone"* property will be the name of the default time zone if no *"timeZone"* property was provided in the options object provided to the Intl.DateTimeFormat constructor. The first edition left the *"timeZone"* property *undefined* in this case.
+        In this version of the ECMAScript 2023 Internationalization API, the *"timeZone"* property will be the identifier of the host environment's time zone if no *"timeZone"* property was provided in the options object provided to the Intl.DateTimeFormat constructor. The first edition left the *"timeZone"* property *undefined* in this case.
       </emu-note>
 
       <emu-note>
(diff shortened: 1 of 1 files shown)


spec/datetimeformat.html
```diff
@@ -991,7 +991,7 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, _patternParts_, _x_, _rangeFormatO
       </emu-note>
 
       <emu-note>
-        It is recommended that implementations use the time zone information of the IANA Time Zone Database.
+        A conforming implementation must recognize *"UTC"* and all other Zone and Link names (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.
```
gibson042, line 994: I'm not sure this note even belongs here, the above algorithm seems to be independent of the tzdb. My primary suggestion is to remove it, and my alternative secondary suggestion is to keep the reference.
```suggestion
        A conforming implementation must recognize *"UTC"* and all Zone and Link names of the IANA Time Zone Database (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.
```

gibson042 (changes requested):
Mostly good, with a few nits to pick.

spec/datetimeformat.html
justingrant: Agree. Removed in latest commit.

spec/datetimeformat.html
```diff
@@ -1237,13 +1237,13 @@ <h1>
...
             <td>[[InDST]]</td>
-            <td>Calculate *true* or *false* using the best available information about the specified _calendar_ and _timeZone_, including current and historical information about time zone offsets from UTC and daylight saving time rules.</td>
+            <td>Calculate *true* or *false* using the best available information about the specified _calendar_ and _timeZoneIdentifier_, including current and historical information from the IANA Time Zone Database about time zone offsets from UTC and daylight saving time rules.</td>
           </tr>
         </table>
       </emu-table>
 
       <emu-note>
-        It is recommended that implementations use the time zone information of the IANA Time Zone Database.
+        A conforming implementation must recognize *"UTC"* and all other Zone and Link names (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.
```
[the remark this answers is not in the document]
justingrant: Agree. Updated in last commit.

justingrant:
oops, just committed to the wrong branch... gimme a few mins to fix.

justingrant:
OK, fixed now. All comments addressed. Should be ready to merge assuming no other issue found.

gibson042 (approved):

PR #766, opened 2023-04-03
Editorial: Capitalize field names to match 262 style guide

### LINKED ISSUE #81: Editorial: Start all fields and slots with an uppercase code point

ben-allen:
fix #81 

I've renamed most (not quite all) of the fields with names in [[camelCase]] to [[PascalCase]] in order to align with 262. As noted in the issue, the thing that makes this tricky is the use of dynamic accesses. The fixes here either removes dynamic accesses when possible to do so relatively elegantly -- for example, if _format_ can only contain "short" or "long":

`Set _blah_.[[<format>]] to *"blah"*.`

is replaced with:

```
If _format_ is *"short"*, then
  Set _blah_.[[Short]] to *"blah"*.
Else,
  Set _blah_.[[Long]] to *"blah"*.
```

or for ones with relatively more possible options, tables are used, and when absolutely necessary, dynamic accesses are maintained but with the first character of the values of variables used for dynamic access replaced with its ASCII-uppercase equivalent. 

The only all-lowercase field names I've preserved are ones like [[nu]] and [[co]] that we've inherited -- this seems stylistically nonoffensive to me, but it shouldn't be much trouble to change those fields to [[NumberingSystem]] and [[Collation]], etc., though some care will have to be taken to make sure those names aren't already in use.

The style on the whole in this PR is somewhat inconsistent due to me trying to more or less match the style of the surrounding spec in each section, but given the scope of the changes I figured it would be a good idea not to do anything further refactoring until more experienced eyes have been on this.


ben-allen:
see #770

PR #768, opened 2023-04-06
Normative: Reorder NF resolved option "roundingPriority"

gibson042:
The second commit is a normative reordering of property sequence; the first commit is a supporting editorial change that should land in some form even if the normative change is rejected.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -17,7 +17,7 @@
 
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
-        1. Let _numberFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%NumberFormat.prototype%"*, &laquo; [[InitializedNumberFormat]], [[Locale]], [[DataLocale]], [[NumberingSystem]], [[Style]], [[Unit]], [[UnitDisplay]], [[Currency]], [[CurrencyDisplay]], [[CurrencySign]], [[MinimumIntegerDigits]], [[MinimumFractionDigits]], [[MaximumFractionDigits]], [[MinimumSignificantDigits]], [[MaximumSignificantDigits]], [[RoundingType]], [[Notation]], [[CompactDisplay]], [[UseGrouping]], [[SignDisplay]], [[RoundingMode]], [[RoundingIncrement]], [[TrailingZeroDisplay]], [[BoundFormat]] &raquo;).
+        1. Let _numberFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%NumberFormat.prototype%"*, &laquo; [[InitializedNumberFormat]], [[Locale]], [[DataLocale]], [[NumberingSystem]], [[Style]], [[Unit]], [[UnitDisplay]], [[Currency]], [[CurrencyDisplay]], [[CurrencySign]], [[MinimumIntegerDigits]], [[MinimumFractionDigits]], [[MaximumFractionDigits]], [[MinimumSignificantDigits]], [[MaximumSignificantDigits]], [[RoundingType]], [[Notation]], [[CompactDisplay]], [[UseGrouping]], [[SignDisplay]], [[RoundingMode]], [[RoundingIncrement]], [[ComputedRoundingPriority]], [[TrailingZeroDisplay]], [[BoundFormat]] &raquo;).
         1. Perform ? InitializeNumberFormat(_numberFormat_, _locales_, _options_).
         1. If the implementation supports the normative optional constructor mode of <emu-xref href="#legacy-constructor"></emu-xref>, then
           1. Let _this_ be the *this* value.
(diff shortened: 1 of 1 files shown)


spec/numberformat.html
```diff
@@ -409,12 +414,6 @@ <h1>Intl.NumberFormat.prototype.resolvedOptions ( )</h1>
           1. Let _v_ be the value of _nf_'s internal slot whose name is the Internal Slot value of the current row.
           1. If _v_ is not *undefined*, then
             1. Perform ! CreateDataPropertyOrThrow(_options_, _p_, _v_).
-        1. If _nf_.[[RoundingType]] is ~morePrecision~, then
-          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"morePrecision"*).
-        1. Else if _nf_.[[RoundingType]] is ~lessPrecision~, then
-          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"lessPrecision"*).
-        1. Else,
-          1. Perform ! CreateDataPropertyOrThrow(_options_, *"roundingPriority"*, *"auto"*).
```
ljharb, line 417: this seems normative (on the editorial commit) - does adding it to the table below cause it to be implicitly set on `options` instead?

sffc:
TG2 discussion (was on the problem, before seeing the concrete PR): https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-04-06.md#nfv3-resolvedoptions

spec/numberformat.html
sffc: Yes.

There is a normative change, which is that the order in which the append order for the options is changing slightly: previously roundingPriority was at the end, but now it is second from the end.

sffc (approved):
LGTM; let's seek approval from the rest of the group.

spec/numberformat.html
ljharb: In the case, the commit is mislabeled; it'd help my review, at least, to split up the editorial and normative parts.

gibson042: The editorial and normative parts **are** separated, as explained in the PR description.
1. [`1ee3bee39e4b2bc21cb9105b7fba2c0c7cef4627`](https://github.com/tc39/ecma402/pull/768/commits/1ee3bee39e4b2bc21cb9105b7fba2c0c7cef4627) is editorial, moving the infallible computation of rounding priority from `NumberFormat.prototype.resolvedOptions` to the `Intl.NumberFormat` constructor and updating the [`resolvedOptions` algorithm](https://tc39.es/ecma402/#sec-intl.numberformat.prototype.resolvedoptions) to add a "roundingPriority" property from the table in step 5 rather than from special steps following table row iteration (but still adding "roundingPriority" as the observably last property).
2. [`0d936078aaf3d83b114ae092f4ac80e7bbfbcf85`](https://github.com/tc39/ecma402/pull/768/commits/0d936078aaf3d83b114ae092f4ac80e7bbfbcf85) is normative, reordering table rows such that "roundingPriority" will be grouped with the other "rounding…" properties and observably precede "trailingZeroDisplay".

ljharb: This comment is on the editorial commit.

gibson042: So what part of «**Editorial: Move computation of NF resolved option "roundingPriority" to initialization**» is mislabeled?

ljharb: @gibson042 that was my question. I only looked at the editorial commit, and I saw the line I'm commenting on here. @sffc said it's normative. Your comment implies that it's not normative ("but still"), which was what i'd been asking about.

ryzokuken (approved):

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-05-04.md#normative-reorder-nf-resolved-option-roundingpriority

sffc:
TG2 discussion again: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-29.md#normative-reorder-nf-resolved-option-roundingpriority

Still needs tests; otherwise has TG2 approval.

gibson042:
@sffc provided a partial answer at https://github.com/tc39/proposal-intl-numberformat-v3/issues/122#issuecomment-1379206743 . But the whole thing is clearly a mess anyway, and at this point I'd be equally happy to rearrange "roundingIncrement" as well (or to keep "roudingPriority" at the end and reduce this PR to the editorial commit that just fixes the `Intl.NumberFormat.prototype.resolvedOptions` algorithm).

sffc:
This PR puts the three rounding modifiers (roundingMode, roundingIncrement, and roundingPriority) next to each other. Why is there desire to change the order amongst those three? At least they are adjacent now.

FrankYFTang:
Here is the READING order of these options now as July 12, 2023)

```
7. Let roundingPriority be ? [GetOption](https://tc39.es/ecma402/#sec-getoption)(options, "roundingPriority", string, « "auto", "morePrecision", "lessPrecision" », "auto").
8. Let roundingIncrement be ? [GetNumberOption](https://tc39.es/ecma402/#sec-getnumberoption)(options, "roundingIncrement", 1, 5000, 1).
9. If roundingIncrement is not in « 1, 2, 5, 10, 20, 25, 50, 100, 200, 250, 500, 1000, 2000, 2500, 5000 », throw a RangeError exception.
10. Let roundingMode be ? [GetOption](https://tc39.es/ecma402/#sec-getoption)(options, "roundingMode", string, « "ceil", "floor", "expand", "trunc", "halfCeil", "halfFloor", "halfExpand", "halfTrunc", "halfEven" », "halfExpand").
11. Let trailingZeroDisplay be ? [GetOption](https://tc39.es/ecma402/#sec-getoption)(options, "trailingZeroDisplay", string, « "auto", "stripIfInteger" », "auto").
```

It seems there are no dependency between these four option readings in the current shape of ECMA402, right?

After this PR merged, the order in the resolvedOptions between these four will be
```
[[RoundingMode]] | "roundingMode"
[[RoundingIncrement]] | "roundingIncrement"
[[ComputedRoundingPriority]] | "roundingPriority"
[[TrailingZeroDisplay]] | "trailingZeroDisplay"
```

so... the new order in both the option reading AND the resolvedOptions()
a. Are NOT alphabetical
b. Are NOT in the same order

so... if the order is really matter and require a PR to change it, would it be better to
A1. Change the option reading order to alphatical between these four?
A2. Change the resolvedOptions() order also to be the same as the reading order?

or at least
B. Change the resolvedOptions() order also to be the same as the reading order if we decide not to do A1?
or 
C. Change the resolvedOptions() order between these two  to be alphatical if we decide not to do A1+A2 or B?

I am not going to block you from merging this PR at this point since it passed TC39 consensus this week already. But I just have hard time to understand the need to change from what we had to this "half baked" state.

FrankYFTang:
v8 bug tracking https://bugs.chromium.org/p/v8/issues/detail?id=14169
v8 cl https://chromium-review.googlesource.com/c/v8/v8/+/4679292 

@gibson042 
If you decide to merge this PR  as is , then please do so and I will land v8 cl 4679292 after you merge it
But if you feel we should reexam the order of both the option reading and the resolvedOptions based on what I just mentioned above  then I will hold the cl 4679292 .

ryzokuken:
Let's take this to a follow-on issue. I'll merge this.

ryzokuken:
@FrankYFTang sorry I missed this, done.

PR #770, opened 2023-04-08
Editorial: Refactor to replace camelCase slot and field names with PascalCase equivalents

### LINKED ISSUE #81: Editorial: Start all fields and slots with an uppercase code point

ben-allen:
fix #81 

Changed slot/field names, carefully avoiding changes to observable behavior that could result from dynamic accesses to those slots. The only names left unchanged are two-letter names like [[co]] and [[nu]], and slots/fields only accessed dynamically.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -54,7 +54,7 @@
         1. Set _options_ to ? ToDateTimeOptions(_options_, *"any"*, *"date"*).
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, ~string~, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
-        1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Set _opt_.[[LocaleMatcher]] to _matcher_.
         1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, ~string~, ~empty~, *undefined*).
         1. If _calendar_ is not *undefined*, then
           1. If _calendar_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
@@ -70,13 +70,13 @@
         1. Set _opt_.[[hc]] to _hourCycle_.
         1. Let _localeData_ be %DateTimeFormat%.[[LocaleData]].
         1. Let _r_ be ResolveLocale(%DateTimeFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %DateTimeFormat%.[[RelevantExtensionKeys]], _localeData_).
-        1. Set _dateTimeFormat_.[[Locale]] to _r_.[[locale]].
+        1. Set _dateTimeFormat_.[[Locale]] to _r_.[[Locale]].
         1. Let _resolvedCalendar_ be _r_.[[ca]].
         1. Set _dateTimeFormat_.[[Calendar]] to _resolvedCalendar_.
         1. Set _dateTimeFormat_.[[NumberingSystem]] to _r_.[[nu]].
-        1. Let _dataLocale_ be _r_.[[dataLocale]].
+        1. Let _dataLocale_ be _r_.[[DataLocale]].
         1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
-        1. Let _hcDefault_ be _dataLocaleData_.[[hourCycle]].
+        1. Let _hcDefault_ be _dataLocaleData_.[[HourCycle]].
         1. If _hour12_ is *true*, then
           1. If _hcDefault_ is *"h11"* or *"h23"*, let _hc_ be *"h11"*. Otherwise, let _hc_ be *"h12"*.
         1. Else if _hour12_ is *false*, then
@@ -96,7 +96,7 @@
           1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZone_).
         1. Set _dateTimeFormat_.[[TimeZone]] to _timeZone_.
         1. Let _formatOptions_ be a new Record.
-        1. Set _formatOptions_.[[hourCycle]] to _hc_.
+        1. Set _formatOptions_.[[HourCycle]] to _hc_.
         1. Let _hasExplicitFormatComponents_ be *false*.
         1. For each row of <emu-xref href="#table-datetimeformat-components"></emu-xref>, except the header row, in table order, do
           1. Let _prop_ be the name given in the Property column of the row.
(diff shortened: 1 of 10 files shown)


spec/datetimeformat.html
```diff
@@ -203,33 +209,33 @@ <h1>Internal slots</h1>
...
-            <li>If the record has an [[hour]] field, it must also have a [[pattern12]] field, whose value is a String value that, in addition to the substrings of the [[pattern]] field, contains at least one of the substrings *"{ampm}"* or *"{dayPeriod}"*.</li>
-            <li>If the record has a [[year]] field, the [[pattern]] and [[pattern12]] values may contain the substrings *"{yearName}"* and *"{relatedYear}"*.</li>
+            <li>A [[Pattern]] field, whose value is a String value that contains for each of the date and time format component fields of the record a substring starting with *"{"*, followed by the name of the field, followed by *"}"*.</li>
+            <li>If the record has an [[Hour]] field, it must also have a [[Pattern12]] field, whose value is a String value that, in addition to the substrings of the [[Pattern]] field, contains at least one of the substrings *"{ampm}"* or *"{dayPeriod}"*.</li>
+            <li>If the record has a [[Year]] field, the [[Pattern]] and [[Pattern12]] values may contain the substrings *"{yearName}"* and *"{relatedYear}"*.</li>
```
gibson042, line 234: Unfortunately, capitalizing the field names in [[LocaleData]].[[<_locale_>]].[[Formats]].[[<_calendar_>]] disrupts this and related mechanisms because of the coupling between field name and brace-enclosed field value substrings (e.g., an "{hour}:{minute} {ampm}" pattern is specifically referencing fields named [[hour]], [[minute]], and [[ampm]]). It's probably most expedient to leave alone any field named after a unit/date–time element/etc., unless you want to do a deeper dive on the format patterns and FormatDateTimePattern's transformation of them.

spec/datetimeformat.html
```diff
@@ -239,37 +245,37 @@ <h1>Internal slots</h1>
...
                 </ul>
             </li>
-            <li>If the record has an [[hour]] field, it must also have a [[rangePatterns12]] field. Its value is similar to the Record in [[rangePatterns]], but it uses a String similar to [[pattern12]] for each part of the range pattern.</li>
-            <li>If the record has a [[year]] field, the [[rangePatterns]] and [[rangePatterns12]] fields may contain range patterns where the [[Pattern]] values may contain the substrings *"{yearName}"* and *"{relatedYear}"*.</li>
+            <li>If the record has an [[Hour]] field, it must also have a [[RangePatterns12]] field. Its value is similar to the Record in [[RangePatterns]], but it uses a String similar to [[Pattern12]] for each part of the range pattern.</li>
+            <li>If the record has a [[Year]] field, the [[RangePatterns]] and [[RangePatterns12]] fields may contain range patterns where the [[Pattern]] values may contain the substrings *"{yearName}"* and *"{relatedYear}"*.</li>
```
gibson042, line 252: This is an example of the inconsistency; [[Year]] and {yearName} and {relatedYear} are elements of the same namespace even though they don't look like it.

spec/datetimeformat.html
```diff
@@ -784,28 +790,40 @@ <h1>ToDateTimeOptions ( _options_, _required_, _defaults_ )</h1>
...
+      <p>The DateTimeStyleFormat abstract operation accepts arguments _dateStyle_ and _timeStyle_, which are each either *undefined*, *"full"*, *"long"*, *"medium"*, or *"short"*, at least one of which is not *undefined*, and _styles_, which is a record from %DateTimeFormat%.[[LocaleData]].[[&lt;_locale_&gt;]].[[Styles]].[[&lt;_calendar_&gt;]] for some locale _locale_ and calendar _calendar_. It returns the appropriate format record for date time formatting based on the parameters.</p>
       <emu-alg>
         1. If _timeStyle_ is not *undefined*, then
           1. Assert: _timeStyle_ is one of *"full"*, *"long"*, *"medium"*, or *"short"*.
-          1. Let _timeFormat_ be _styles_.[[TimeFormat]].[[&lt;_timeStyle_&gt;]].
+          1. Let _first_ be the ASCII-uppercase of the substring of _timeStyle_ from 0 to 1.
+          1. Let _rest_ be the substring of _timeStyle_ from 1.
+          1. Let _timeStyleField_ be the string-concatenation of _first_ and _rest_.
+          1. Let _timeFormat_ be _styles_.[[TimeFormat]].[[&lt;_timeStyleField_&gt;]].
```
gibson042, line 800: More string manipulation that we don't want in the spec.

spec/displaynames.html
```diff
@@ -33,23 +33,66 @@ <h1>Intl.DisplayNames ( _locales_, _options_ )</h1>
...
+        1. Else if _style_ is *"short*", then
+          1. Let _styleFields_ be _typeFields_.[[Short]].
+        1. Else,
+          1. Let _styleFields_ be _typeFields_.[[Long]].
         1. Assert: _styleFields_ is a Record (see <emu-xref href="#sec-Intl.DisplayNames-internal-slots"></emu-xref>).
         1. Set _displayNames_.[[Fields]] to _styleFields_.
         1. Return _displayNames_.
       </emu-alg>
+      <emu-table id="table-displaynames-types">
+        <emu-caption>Fields Corresponding to Type Names</emu-caption>
```
gibson042, line 62: Ugh, there must be a better way.

spec/relativetimeformat.html
```diff
@@ -299,13 +299,31 @@ <h1>PartitionRelativeTimePattern ( _relativeTimeFormat_, _value_, _unit_ )</h1>
...
         1. If _style_ is equal to *"short"*, then
-          1. Let _entry_ be the string-concatenation of _unit_ and *"-short"*.
+          1. Let _entry_ be the string-concatenation of _unit_ and *"Short"*.
         1. Else if _style_ is equal to *"narrow"*, then
-          1. Let _entry_ be the string-concatenation of _unit_ and *"-narrow"*.
+          1. Let _entry_ be the string-concatenation of _unit_ and *"Narrow"*.
         1. Else,
           1. Let _entry_ be _unit_.
-        1. If _fields_ doesn't have a field [[&lt;_entry_&gt;]], then
+        1. If _fields_ desn't have a field [[&lt;_entry_&gt;]], then
```
gibson042, line 326:
```suggestion
        1. If _fields_ doesn't have a field [[&lt;_entry_&gt;]], then
```

gibson042 (changes requested):
This is better than the previous attempt, but is still running into similar issues. I think it will make the most sense to tackle fields piecemeal, leaving those subject to `[[<_…>]]` dynamic lookup for last.

spec/datetimeformat.html
ben-allen: blast, I should have caught that!  I'm guessing it's correct to assume that a deeper dive on FormatDateTimePattern should be understood as a very low priority.

spec/datetimeformat.html
ben-allen: What's best practice here -- table iteration when possible, or just "this is a place where by necessity 402 style guidelines must diverge from 262"?

ben-allen:
Temperature check: would it be most appropriate to:

1. Revert all changes referenced by brace-enclosed field names in patterns, leaving only unambiguously non-observable field name changes,
2. Deeper dive on format patterns -- which is something I'd like to do out of selfish "this refactor is getting me familiar with every corner of the spec" motivations, or just
3. Accept that refactoring to adhere to 262 style is a low priority for a reason.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-05-04.md#editorial-refactor-to-replace-camelcase-slot-and-field-names-with-pascalcase

ben-allen:
closed; abandoned for now

PR #771, opened 2023-04-10
Editorial: Replaced 'Internal Slot' with 'Field Name' in pluralrules, 'Field' wi…

### LINKED ISSUE #711: Record fields mislabeled “internal slots” in table
The first column of [Table 15 (Plural Rules Operands Record Fields)](https://tc39.es/ecma402/#table-plural-operands) is mislabeled “Internal Slot”.

Currently [Table 8 (Record returned by ToLocalTime)](https://tc39.es/ecma402/#table-datetimeformat-tolocaltime-record) uses “Field”. Both tables should probably say “Field Name” given ECMA-262 has ~33 similar tables which all use that term. Its definition of the Record spec-type also seems to distinguish “field names” from “fields” (field name + value).

ben-allen:
fix #711

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -1183,7 +1183,7 @@
         <table class="real-table">
           <thead>
             <tr>
-              <th>Field</th>
+              <th>Field Name</th>
               <th>Value Calculation for Gregorian Calendar</th>
             </tr>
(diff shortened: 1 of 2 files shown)


ryzokuken (approved):
Thanks very much!

PR #772, opened 2023-04-11
Editorial: Regularize self-references to ECMA-402

### LINKED ISSUE #641: Self-references to ECMA-402 are unnecessarily inconsistent
(prompted by https://github.com/tc39/ecma402/pull/610#discussion_r705794932 )

ECMA-402 makes many references to itself, but does so inconsistently:
```sh
$ grep --no-filename -roiE 'this specification|(current )?(version [^[:punct:]]*)?ECMA[^[:punct:]]*Specification' spec/ | \
  sed 's/\(Internationalization API Specification\).*/\1/' | sort | uniq -c
      1 current version of the ECMAScript Internationalization API Specification
      1 Ecmarkup is an HTML and Markdown dialect that provides a framework and toolset for authoring ECMAScript specifications in plaintext and processing the specification
     16 ECMAScript 2022 Internationalization API Specification
      8 ECMAScript 2022 Language Specification
      1 ECMAScript implementation that implements this Internationalization API Specification
      3 ECMAScript Language Specification
      1 ECMAScript objects in the ECMAScript 2022 Language Specification
     10 this specification
      1 this Specification
      4 This specification
```

It would be better to uniformly adopt a consistent pattern, such as "this specification" (which seems to be preferred by ECMA-262).

ben-allen:
fix #641 

Replaced phrases like "current version of the ECMAScript Internationalization API Specification" with "this specification" where appropriate, with some edits to wording of surrounding sentences to avoid ambiguity between references to _this_ specification and references to ECMA-262. 

Thought: ECMA-419 introduces 262 with this sentence:

> This Standard does not make any changes to the ECMAScript language as defined by ECMAScript Language Specification (ECMA-262)

and then subsequently refers to it as ECMA-262. Given how often references to 262 appear in 402, I'm tempted to follow their lead.

### DIFF
--- a/spec/annexes.html
+++ b/spec/annexes.html
@@ -2,7 +2,7 @@
   <h1>Implementation Dependent Behaviour</h1>
 
   <p>
-    The following aspects of the ECMAScript 2023 Internationalization API Specification are implementation dependent:
+    The following aspects of this specification are implementation dependent:
   </p>
 
   <ul>
--- a/spec/conformance.html
+++ b/spec/conformance.html
@@ -1,10 +1,10 @@
 <emu-clause id="conformance">
   <h1>Conformance</h1>
   <p>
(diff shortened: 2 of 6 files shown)


gibson042 (approved):
Thanks!

PR #774, opened 2023-04-12
Editorial: Use consistent wording for List interactions

gibson042:
Align with ECMA-262:
* https://github.com/tc39/ecma262/commit/2df27d363f9622193d5100d769039dde47a9fa64
* https://github.com/tc39/ecma262/pull/2152

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -17,9 +17,9 @@
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _internalSlotsList_ be &laquo; [[InitializedCollator]], [[Locale]], [[Usage]], [[Sensitivity]], [[IgnorePunctuation]], [[Collation]], [[BoundCompare]] &raquo;.
         1. If %Collator%.[[RelevantExtensionKeys]] contains *"kn"*, then
-          1. Append [[Numeric]] as the last element of _internalSlotsList_.
+          1. Append [[Numeric]] to _internalSlotsList_.
         1. If %Collator%.[[RelevantExtensionKeys]] contains *"kf"*, then
-          1. Append [[CaseFirst]] as the last element of _internalSlotsList_.
+          1. Append [[CaseFirst]] to _internalSlotsList_.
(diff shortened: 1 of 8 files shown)


ben-allen:
I'm wondering if by analogy to this change in 262:

```<p>When an algorithm iterates over the elements of a List without specifying an order, the order used is the order of the elements in the List.</p>```


we should add a sentence about how unless specified otherwise tables are iterated in table order, and then remove all the instances of `, in table order,` from loops

ben-allen (changes requested):
To my eye it looks like there's a couple you've missed, but this is otherwise fantastic

sffc:
@gibson042 Can you take a look at @ben-allen's feedback? This looks almost mergeable.

gibson042:
@ben-allen please re-review!

> I've found a couple of other examples of the wording you're fixing:
> 
> negotiation.html:64: `1. If _canonicalizedTag_ is not an element of _seen_, append _canonicalizedTag_ as the last element of _seen_.`
> 
> negotiation.html:270: `1. If _availableLocale_ is not *undefined*, append _locale_ to the end of _subset_.`

Thanks; fixed.

> I'm wondering if by analogy to this change in 262:
> 
> `<p>When an algorithm iterates over the elements of a List without specifying an order, the order used is the order of the elements in the List.</p>`
> 
> we should add a sentence about how unless specified otherwise tables are iterated in table order, and then remove all the instances of `, in table order,` from loops

I've held off on doing this because I can't find a satisfactory way that is both in scope and discoverable—the latter characteristic being important, and handled in ECMA-262 by internal cross referencing (note how e.g. [FindViaPredicate](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-findviapredicate) step 4 iterates over _indices_, which is defined in either step 2 or 3 to be a [List](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-list-and-record-specification-type) with a link to the very section establishing the "_When an algorithm iterates over the elements of a List without specifying an order, the order used is the order of the elements in the List_" convention). We can pursue this in a followup.

ben-allen (approved):

PR #780, opened 2023-05-05
Normative: Added note about sets of locales for web browser implementations needing to not change as a result of user behaviour

### LINKED ISSUE #588: ships the entire payload requirement
This request came from YSV in the 2021-05-25 TC39 meeting

"
YSV: ....   we would require — and I read through the spec, I don't think that the spec is permissive, in the sense that this wouldn't be a requirement — the requirement for us in order to make sure that the fingerprinting part doesn't become an issue is that any implementer who implements this API also ships the entire payload. So those are the two high-level issues we've got.

FYT: Sorry, can you clarify, when you say the entire payload, what does that mean?

YSV: So you can't for example ship part of the data you're referencing. For example, for the calendars and languages, you have to ship all of it, you can't just ship a subset. 

...
SFC: I wanted to reply and clarify on YSV's second ask about requiring that implementations ship all the data. If I understand that correctly, I interpret that to mean that in order to resolve the fingerprinting concern, the set returned by this function needs to be equal to the browser version number. And I think it's an interesting concern that warrants additional discussion. I'm not convinced that that this API is going to cause any fingerprinting concerns in its own right. Because if we do get to a point where browsers are, you know, downloading additional locales on the fly, which would be definitely a nice thing to shoot for, that's going to raise the same types of fingerprinting concerns that this proposal would expose. For example, if you were to call create a date-time format in — pick your favorite esoteric locale, how about Klingon — if you were to create a DateTimeFormat in Klingon, and that is producing strings for you, and then Klingon gets added to the available numbering systems, both of those are equivalent fingerprinting concerns. So I’m not convinced that the requirement that browsers ship all data is necessarily unique to this proposal and I don't necessarily see that fitting in here.

YSV: Yes, you're absolutely right. This isn't going to be unique to this proposal. This is the first proposal that we're asking for this, but this is, if this goes forward we will be asking this from all proposals, that there will be no partial data sets being shipped. 

FYT: Can I ask something here with that? I'm not quite sure about that. Because as you can see here, in this particular API the possible keys we have are `"calendar"`, `"collation"`, `"currency"`, `"numberingSystem"`, `"timeZone"`, and `"unit"`. So whether we ship 100 locales or 10 locales is not discoverable by this API, because `"locale"` is not one of the keys. Whether we ship with ten locales may impact it, but not 100 locales. First 10, locale. So, I'm not sure. I mean, I understand the fingerprint concern. If one of the possible keys is `"locale"`, fine, the thing you require has merit. And if you're saying, please ship all the calendars, then I agree with you. But in this particular case with regard to not shipping all the locales, ten or 100 shouldn't be observable by this API at all. So I don't know that that will be impacting the fingerprinting concern.

YSV: Sorry. Did I say locale? I mean specifically that we don't ship partial data sets of these things that we're going be exposing in a discoverable way. I don't know if I said locale there, but that was my meaning.

FYT: Thank you for the clarification. So I think what you're saying is that we should be shipping all calendars, all collations, all currencies, all number systems, all time zones, and all units. Is that what you mean?

YSV: Yes.

FYT: I understand what you're asking. I will bring that back to TG2 for that discussion. Thank you.

YSV: Yeah. And this also comes with — any future APIs that do something like this, are required to ship the full data set as part of their implementation. 

FYT: Could you clear up? What do you mean by 'like this'? 

YSV: Because there's a very vague if we have other enumeration APIs we don't ship partial data sets.

...
JHD: This is a reply to YSV's thing. So it seems totally reasonable to find a way to mandate the full data set, instead of partial ones, because that reduces the fingerprinting landscape, but I think that's completely orthogonal to this proposal because I can do all of the same fingerprinting that I could do with this, without this. It just means I have to do a little more work in advance, but it's not that hard. I just have to go to a bunch of browsers and build a long array of possibilities and then try them. So I think that it's fine if this proposal surfaced the need for that requirement, but I would hope that that would be pursued independently and not be chained to this proposal.
...
USA: I wanted to echo what JHD was saying in that my understanding of this proposal is that doesn't actually incur any additional data and it just provides a convenient API to have the data exposed in a certain way, but if anybody wanted, they could have access to this data anyway. We have userland modules which do some of these interesting operations and get this data. So just for convenience reasons.
"

ben-allen:
fix #588

### DIFF
--- a/spec/overview.html
+++ b/spec/overview.html
@@ -73,10 +73,7 @@
       </li>
     </ul>
     <p>
-      In browser implementations the initial set of locales, currencies, calendars, numbering systems, and other enumerable items visible to a particular origin must be the same for all users sharing the same user agent string (engine and platform version). Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
-    </p>
-    <p>
-      As a result of this constraint, the first time a browser implementation that allows on-demand locale installation receives a request from a particular origin that could require installing a new locale, it must not reveal whether or not that locale is already installed.
+      In browser implementations the initial set of locales, currencies, calendars, numbering systems, and other enumerable items visible to a particular origin must be the same for all users sharing the same user agent string (engine and platform version). Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions. As a result of this constraint, the first time a browser implementation that allows on-demand locale installation receives a request from a particular origin that could require installing a new locale, it must not reveal whether or not that locale is already installed.
     </p>
 
     <emu-clause id="sec-compatibility">
--- a/spec/overview.html
+++ b/spec/overview.html
@@ -72,6 +72,12 @@
         <em>Subsets of Unicode:</em> Some operations, such as collation, operate on strings that can include characters from the entire Unicode character set. However, both the Unicode Standard and the ECMAScript standard allow implementations to limit their functionality to subsets of the Unicode character set. In addition, locale conventions typically don't specify the desired behaviour for the entire Unicode character set, but only for those characters that are relevant for the locale. While the Unicode Collation Algorithm combines a default collation order for the entire Unicode character set with the ability to tailor for local conventions, subsets and tailorings still result in differences in behaviour.
       </li>
     </ul>
+    <p>
+      In browser implementations the initial set of locales, currencies, calendars, numbering systems, and other enumerable items visible to a particular origin must be the same for all users sharing the same user agent string (engine and platform version). Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
+    </p>
+    <p>
+      As a result of this constraint, the first time a browser implementation that allows on-demand locale installation receives a request from a particular origin that could require installing a new locale, it must not reveal whether or not that locale is already installed.
+    </p>
 
     <emu-clause id="sec-compatibility">
       <h1>Compatibility across implementations</h1>


ljharb:
Why would this restriction only apply to web browsers?

ben-allen:
This PR is in response to feedback from the 2021-05-25 TC39 meeting and is meant to address concerns about potential fingerprinting issues that only pertain to browser implementations.

ljharb:
Is there any reason *not* to apply the same restrictions to all engines tho? The ideal is that everything applies to everyone equally; having something only apply to a subset of impls is a suboptimal outcome.

ben-allen:
updated to apply restriction to all hosts

ljharb (approved):

ljharb:
(please don't land this until 402 editors have reviewed)

ljharb:
(please don't land this until 402 editors have reviewed)

spec/overview.html
```diff
@@ -72,6 +72,9 @@ <h1>Implementation Dependencies</h1>
         <em>Subsets of Unicode:</em> Some operations, such as collation, operate on strings that can include characters from the entire Unicode character set. However, both the Unicode Standard and the ECMAScript standard allow implementations to limit their functionality to subsets of the Unicode character set. In addition, locale conventions typically don't specify the desired behaviour for the entire Unicode character set, but only for those characters that are relevant for the locale. While the Unicode Collation Algorithm combines a default collation order for the entire Unicode character set with the ability to tailor for local conventions, subsets and tailorings still result in differences in behaviour.
       </li>
     </ul>
+    <emu-note>
+      The set of locales made available by ECMAScript hosts must not change as the result of user behaviour, and the set of available locales must not produce observable differences between two users using the same version of the same host on the same platform. As a result, ECMAScript hosts must not allow on-demand installation of new locales.
```
sffc, line 76: Set of locales, yes, but also the set of all enumerable items including currencies, numbering systems, calendars, etc.

Also, please explain why we're adding this constraint. I imagine that it may be desirable in the future to relax this constraint, and we need to understand why it existed.

aphillips:
The W3C I18N Working Group discuss this PR today [in our teleconference](https://www.w3.org/2023/05/11-i18n-minutes.html#t06). I am adding this comment on our behalf. 

We are concerned that this prohibition will disadvantage smaller language/cultural communities who might rely on installation of support to enable locale-based APIs in the browser or JS host. 

We feel that precluding the ability to install a locale or the parts of a locale (such as dictionaries for spell check/breaking/etc.) that assist with high-quality presentation on the Web and in JS applications has the potential to negatively impact those communities that cannot depend on support from browser or system vendors. If there is a "fingerprinting" risk associated with such installation, providing a warning to the user might be the best appropriate response.

Also, note that currently we are not aware of runtimes that allows the list of locales to be updated (other than by updating the entire underlying ICU build), so this strikes us as preventing a feature from existing that might be useful. Also, we note that CLDR releases include new locales twice each year, so presumably browsers would change their list of available locales as updates propagate.

sffc:
CC @codehag for feedback on @aphillips' comment above (see issue #588 for a reminder of the problem this PR is trying to solve)

ljharb:
@aphillips doesnt this prohibition only require refreshing the page after installing a new locale?

aphillips:
@ljharb asked:

> doesnt this prohibition only require refreshing the page after installing a new locale?

Not as far as I can tell? It seems to require that any two distinct users on the same version on the same platform (_not_ the "same host") should always get the same enumerable set of available values. The change in cb6449b extends this to include anything (numbering systems, calendars, etc.). I note that one frequent cause of such patching would be time zone data (which is not listed but is the regular source of runtime patching outside the normal release cycle).

If I understand the threat here, it's to prevent a bad actor from installing a locale into a user's browser and then using that locale ID (perhaps using a well-formed locale ID like: `bad-actr-06b3c`) to track the specific user (or to differentiate groups of users). That suggests that there should be no _silent_ installation of locales or locale data, not necessarily an outright prohibition?

There is a similar issue (w3c/css-drafts#4055) related to fingerprinting based on fonts (which have a much higher level of per-installation variability and which, unlike locales, can actually be installed currently 😃). The challenge here is to support under-served communities of users--particularly when the threat is more-or-less theoretical--without exposing large groups of Web users to abusive behavior.

justingrant:
> I note that one frequent cause of such patching would be time zone data (which is not listed but is the regular source of runtime patching outside the normal release cycle).

Which hosts and engines currently do runtime patching of time zone data? Are these OS patches or browser patches?

ljharb:
@Constellation can lockdown mode come into effect without reloading the page? If it forces a reload, then this note wouldn't apply at all, since it wouldn't be observable within the lifetime of a program.

sffc:
TG2 notes: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-01.md#editorial-added-note-about-sets-of-locales-for-web-browser-implementations-needing-to-not-change-as-a-result-of-user-behaviour-780

Seeking feedback from @dminor

dminor:
Our concern is that on-demand installation of locale data could provide an easy fingerprinting vector for members of smaller language/cultural communities who may face discrimination or persecution, for example by dominant cultural groups or by the government of the country in which they live.

Our position is that it is better to ship data for all locales in a single bundle, which ensures that data for smaller communities is available, without exposing them to a fingerprinting risk.

I believe it would be very difficult to create a user warning that would explain the potential risk in a way that would allow a user to make an informed decision about accepting extra locale data. I suspect most people would ignore these warnings. I'd also point out that there's no guarantee that for a small linguistic community the text of the warning would be localized, which would decrease the likelihood of making an informed decision.

The key point is that the set of locales should not change as a result of user behaviour. We're not trying to prevent vendors from shipping a new bundle of locale data to users as part of an update, just data for individual locales. In the case of Firefox, CLDR and timezone updates are done as part of our normal release cycle anyway.

dminor (approved):

srl295:
> The W3C I18N Working Group discussed this PR today [in our teleconference](https://www.w3.org/2023/05/11-i18n-minutes.html#t06). I am adding this comment on our behalf.
> 
> We are concerned that this prohibition will disadvantage smaller language/cultural communities who might rely on installation of support to enable locale-based APIs in the browser or JS host.

This type of feature was recently requested to me, in the context of ICU4C, and explicitly for the purpose of supporting minority, disadvantaged languages.   (Which, yes, could potentially be at-risk for fingerprinting of various kinds.) 

> We feel that precluding the ability to install a locale or the parts of a locale (such as dictionaries for spell check/breaking/etc.) that assist with high-quality presentation on the Web and in JS applications has the potential to negatively impact those communities that cannot depend on support from browser or system vendors. If there is a "fingerprinting" risk associated with such installation, providing a warning to the user might be the best appropriate response.

Agreed. Default breaking for `Thai` script, for example, in current implementations causes problems for minority script users.  more [here](https://unicode-org.atlassian.net/browse/ICU-13219)

> Also, note that currently we are not aware of runtimes that allows the list of locales to be updated (other than by updating the entire underlying ICU build), so this strikes us as preventing a feature from existing that might be useful. Also, we note that CLDR releases include new locales twice each year, so presumably browsers would change their list of available locales as updates propagate.

CLDR releases include certain locales as "basic" and above, but there are other locales not included.

ICU4C default build includes certain locales, but not others.

Vendors include certain locales, but not others.

In short, certain locales are *already* excluded from web implementations. I'm concerned that requiring that these locales cannot be added on the fly could end up negatively impacting users of already-digitally disadvantaged languages.

srl295:
> Is there any reason _not_ to apply the same restrictions to all engines tho? The ideal is that everything applies to everyone equally; having something only apply to a subset of impls is a suboptimal outcome.

On another topic,  Node.js has, from the first versions that included Intl by default, had [the ability to customize at build and runtime](https://nodejs.org/api/intl.html#providing-icu-data-at-runtime) the set of locales available, and also to supplement the locales depending on the startup environment. It's also been requested to have some way to add locales at runtime there as well. This language seems to make Node.js v0.12 onwards potentially noncompliant.  I don't see the argument or this type of environment at all.

ljharb:
@srl295 that doesn't imply to me that it can be changed during the lifetime of a program, only at program start time.

My understanding of this requirement is that *once a JS program has started*, it can't observe *further* changes to the list of available locales. To that end, anything that requires refreshing or navigating a page, or, restarting an application or launching a process, in order to observe a different set of locales seems to me that it complies with this requirement.

srl295:
> @srl295 that doesn't imply to me that it can be changed during the lifetime of a program, only at program start time.

OK. So "user behaviour" is scoped to the JS runtime? That's helpful…  I then don't see how fingerprinting is mitigated. 

> My understanding of this requirement is that _once a JS program has started_, it can't observe _further_ changes to the list of available locales. To that end, anything that requires refreshing or navigating a page, or, restarting an application or launching a process, in order to observe a different set of locales seems to me that it complies with this requirement.

That would be very different.  And wouldn't then bring as much concern.  Adding locales while running has been discussed as well, but certainly has a lot of other challenges.

ryzokuken (approved):
LGTM

gibson042 (approved):
LGTM

sffc (changes requested):
Leaving Requested Changes to reflect that we should try to get alignment on @aphillips point

ben-allen:
With regard to @aphillips's point: as @dminor has observed, the problem goes beyond the "bad-actr" case, since it would be possible for sites to individually identify users from politically disadvantaged/persecuted groups by checking if they have locale data associated with those groups installed.

I've been in touch with @manishearth, who suggested the following as a strategy for avoiding this problem while still providing localized content for users of non-default locales:

* Have an API for loading extra locale data
* The first time any particular website uses that API with a particular non-default locale, it always behaves as if the locale needs to be downloaded. Once it's been downloaded, state is set indicating that that website knows that the locale has been downloaded. This state persists until cleared by user.
* On all other websites, the locale loading API behaves as if the locale hasn't been downloaded and then pretends to download it. 

It's possible that the "pretends to download the data" stage can be observed by observing bandwidth changes, which could be a problem.  And actually always re-downloading the data is a possibility, but would be not great for people on slow internet connections.

ljharb:
Why is locale data needed on the fly without a page refresh?

srl295:
> Why is locale data needed on the fly without a page refresh?

If I understand your question, a possible answer is: because the page's UI might allow you to change the locale used via a popup menu?

ljharb:
Right, but if it requires a refresh to take effect when a new locale is needed, this requirement is satisfied.

Manishearth:
> Have an API for loading extra locale data


To add to this, I would imagine this API would be something like `ensureLocaleLoaded(locale)` where it returns a Promise that resolves when the locale is loaded (immediately if the locale was already loaded on this webpage in the past, otherwise in some time), and rejects if the locale could not be found or there was a problem with the internet connection (which will likely have identical results across a particular browser version)

> Right, but if it requires a refresh to take effect when a new locale is needed, this requirement is satisfied.

Is this something we wish to require of users? I think that's rather onerous.

Especially if the use of this API would mostly be of the form "check if locale is loaded and then proceed" which works really well with promise chaining.

Manishearth:
Another useful tool to provide may be a special header value that webpages can set that asks the user agent to preload a locale. They'd still need to use a promise-based API since the loading may not happen immediately

(This would be at the level of a web specification, though, since we don't do headers here. But it fits in the larger picture.)

ljharb:
you can refresh it for them with window.refresh(); that doesn’t seem too onerous.

Manishearth:
Yes, my comment is not about the refresh being manual, it is that it is a pretty onerous thing to ask web developers to refresh their page halfway through; and it strongly impacts how they may design their webpage, alongside having a noticeable effect from the user's perspective.

spec/overview.html
```diff
@@ -72,6 +72,13 @@ <h1>Implementation Dependencies</h1>
...
       </li>
     </ul>
+    <emu-note>
+      The set of locales, currencies, calendars, numbering systems, and other enumerable items made available by ECMAScript hosts must not change as the result of user behaviour, or otherwise result in two users using the same version of the same host on the same platform becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
+    </emu-note>
+
+    <emu-note>
+      As a result of this constraint, browser implementations that allow on-demand installation of new locales should not disclose whether non-standard locales and locale components are already installed.
```
Manishearth, line 80: nit: perhaps explicitly call out something about origins?

ljharb:
```suggestion
      As a result of this constraint, browser implementations that allow on-demand installation of new locales should not disclose whether non-standard locales and locale components are already installed, or have been installed during the lifetime of the program.
```

ljharb:
It might be worth noting that one potential way to satisfy this constraint is to pretend that all on-the-fly-available locales are already installed.

sffc:
@ben-allen to update the spec text to incorporate the remainder of @Manishearth's feedback.

spec/overview.html
ben-allen: I've altered the wording to clarify that the constraint is no longer against implementations installing new locales, but is instead against revealing upon first request whether or not new locales have been installed.

spec/overview.html
```diff
@@ -73,6 +73,14 @@ <h1>Implementation Dependencies</h1>
       </li>
     </ul>
 
+    <emu-note>
+      Changes to the set of locales, currencies, calendars, numbering systems, and other enumerable items made available by ECMAScript hosts must not result in two users using the same version of the same host on the same platform becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
```
Manishearth, line 77: Should this be normative? The line below should be informative, but it sounds like this is expected to be normative?

ben-allen: This should indeed be normative, though it's in that fuzzy space where it's a normative change that doesn't impact any extant implementation.

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-09-07.md#normative-added-note-about-sets-of-locales-for-web-browser-implementations-needing-to-not-change-as-a-result-of-user-behaviour-780

spec/overview.html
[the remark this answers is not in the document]
ben-allen: Much clearer, will use this

spec/overview.html
```diff
@@ -73,6 +73,14 @@ <h1>Implementation Dependencies</h1>
       </li>
     </ul>
 
+    <emu-note>
```
bakkot, line 76: If this is normative, it maybe shouldn't be a NOTE?

spec/overview.html
```diff
@@ -73,6 +73,14 @@ <h1>Implementation Dependencies</h1>
...
+      The initial set of locales, currencies, calendars, numbering systems, and other enumerable items visible to a particular origin must be the same for all users sharing the same user agent string (engine and platform version). Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
+    </emu-note>
+
+    <emu-note>
+      Non-normative: As a result of this constraint, the first time a browser implementation that allows on-demand locale installation receives a request from a particular origin that could require installing a new locale, it must not reveal whether or not that locale is already installed.
```
michaelficarra, line 81: Are notes in 402 not all non-normative?

spec/overview.html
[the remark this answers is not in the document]
ByteEater-pl: The notion of "user agent string" is foreign to ECMAScript and not a universal among implementations. @ljharb, please take a look, it seems to be going contrary to your comments above.

As something far from matters typically expected to be specified in a programming language (or API closely related to it, being an optional extension of a language's standard library) and pertaining to only a subset of implementations, namely Web browsers, shouldn't it live in a spec for those? Like the one defining what a Web browser is and what requirements it has to satisfy as an ECMAScript implementation (if supported) in addition to ECMA-262?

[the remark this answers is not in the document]
ljharb: I agree the concept and mention of a user agent doesn’t make any sense in an ecma specification.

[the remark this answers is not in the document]
Manishearth: In that case I think we need to step back and consider the premise here. The fingerprinting concerns are largely only relevant for web browser environments. 

_Either_ it is ECMA402's job to address these fingerprinting concerns and thus it _must_ be allowed to refer to mechanisms available in those contexts, _or_ it is not ECMA402's job, and we don't need to handle this at all. We can't have our fingerprinting cake if we're not planning on eating it.

I suspect the framing here can be refined a bit to be clear that it is talking in a web browser context only. Alternatively, a more general point can be made about "systems where fingerprinting is a concern, like web browsers", and instead of saying UA strings talk about "already distinguishable bits of information (in the case of browsers, this is platform/version/UA string)". Something like that.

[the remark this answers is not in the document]
ByteEater-pl: IMO it really belongs elsewhere. But if a critical mass insists it's desirable to make the spec longer and draw the attention also therein to this concern present in some implementations, an **informative** note with sufficiently generic wording could be added.

ben-allen:
Most recent push is a minimal change based on TC39 feedback &mdash; all I did was move the text that was in notes into the main text of the Implementation Dependencies section.

ryzokuken (approved):

sffc (approved):

spec/overview.html
```diff
@@ -72,6 +72,12 @@ <h1>Implementation Dependencies</h1>
...
       </li>
     </ul>
+    <p>
+      In browser implementations the initial set of locales, currencies, calendars, numbering systems, and other enumerable items visible to a particular origin must be the same for all users sharing the same user agent string (engine and platform version). Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. This constraint is imposed to reduce the fingerprinting risk inherent in internationalization, and may be relaxed in future revisions.
+    </p>
+    <p>
+      As a result of this constraint, the first time a browser implementation that allows on-demand locale installation receives a request from a particular origin that could require installing a new locale, it must not reveal whether or not that locale is already installed.
```
michaelficarra, line 79: This shouldn't be a separate paragraph from the one before it.

dminor:
To follow on from the discussion at TC39 yesterday, the SpiderMonkey team still considers this to be very important. I will ask for review on the current text from our privacy team before the next plenary.

dminor:
Sorry about the long delay here, I've checked with our privacy people and the proposed text looks good to us.

sffc:
TG2 approval in May 2025: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-05-08.md#normative-added-note-about-sets-of-locales-for-web-browser-implementations-needing-to-not-change-as-a-result-of-user-behaviour-780

Issue #256, opened 2018-08-06
What should happen with the usage: "sort" vs "search" choice in implementations?


littledan:
@gsathya pointed out that, in implementations, this choice is often *not* passed down to ICU (he was looking at V8, and I see something similar in SpiderMonkey). It's unclear to me what the ICU API is, but [UTS #35 mentions](http://unicode.org/reports/tr35/tr35-collation.html#Collation_Type_Fallback) separate data for searching. Does anyone have more background in the difference here, and whether the plumbing is in place to make this reality?

jungshik:
ICU does offer ways to choose between the two iirc.  The way might be (I have to check) to call two different apis (one for search and the other for collation) instead of two different option values passed to a single api, but don't quote me. I may as well misremember.

anba:
SpiderMonkey and V8 both seem to respect the `usage` option.

```js
js> ["AE", "Ä"].sort(new Intl.Collator("de", {usage: "sort"}).compare)
["\xC4", "AE"]
js> ["AE", "Ä"].sort(new Intl.Collator("de", {usage: "search"}).compare)
["AE", "\xC4"]

d8> ["AE", "Ä"].sort(new Intl.Collator("de", {usage: "sort"}).compare)
["Ä", "AE"]
d8> ["AE", "Ä"].sort(new Intl.Collator("de", {usage: "search"}).compare)
["AE", "Ä"]
```

gsathya:
Looks like we need to pass the correct `-u-co-` extension when creating the icu::Locale.

gsathya:
Hmm, looks like this will make icu add the `u-co-search` extension to the language tag, giving us
```
d8> new Intl.Collator("de", {usage: "search"}).resolvedOptions().locale
"de-u-co-search"
```

But according to the [ECMA402 spec](https://tc39.github.io/ecma402/#sec-intl-collator-internal-slots), "search" is not a valid value for the "co" extension. V8 will have to manually trim the `u-co-search` subtag from the locale. What was the reason to ban it in the ECMA402 spec?

anba:
I guess "search" (and "standard") were banned from "co" to force users to use the "usage" option instead. (IIRC there are somewhere notes explaining the motivation for the "usage" option and why it only accepts "search" and "sort". But I don't have a link handy.)

gsathya:
> I guess "search" (and "standard") were banned from "co" to force users to use the "usage" option instead.

[...]

Issue #35, opened 2015-09-16
Relative time/dates


zbraniecki:
There are two major areas where relative time/date formatting could use standardization:
1. Relative time representation

Good example are timers - representing time like "31 hours, 15 minutes and 23 seconds" as "31:15:23".
1. pretty relative date/time.
   A lot of web apps humanize relative date and time with messages like:
- a few seconds ago
- less than a minute ago
- one hour ago
- in one hour
- yesterday
- last week
- in 2 days

All of those patterns are available in CLDR and it would enable web authors to provide better user interfaces if we could incorporate this kind of formatting either into DateTimeFormat or into a separate RelativeDateTimeFormat.

caridy:
@zbraniecki, now that we are ready to start working on the 3rd edition (using this repo), I plan to champion the relative time feature based on the initial work that we did on https://github.com/yahoo/intl-relativeformat. Do you want to help?

zbraniecki:
absolutely. We're currently using a very nasty hack for that[1] and I'd love to see some better, possibly CLDR based, API for that.
I'll be definitely interested in shimming whatever we come up with for Firefox and Firefox OS which should give us some ability to test how it works early on.

[1] https://github.com/mozilla-b2g/gaia/blob/19696d5db8e7e078204b15597b9d01c788e5f04f/shared/js/l10n_date.js#L132

caridy:
I could not find "a few minutes ago" data in CLDR, nor the "one", "less than", etc...

Here is the initial spec: https://github.com/caridy/intl-relative-time-spec

zbraniecki:
That looks awesome! Thanks @caridy 

I'll play with this against our use cases more, but first thing I miss here is pure duration. You have "X ago" and "in X", but you don't have "X". Examples:
* "1 month" "1m"
* "26 hours" / "26h"
* "5 seconds", "5 sec", "5s"
etc.

zbraniecki:
Hmm, the more I think about this the more I feel like there's a direct overlap area between three proposals:
* this one
* #37 - abbreviate numbers
* #32 - Unit formatting
* NumberFormat

First of all, there are three modes - future (in 5 minutes), past (5 minutes ago) and "neutral" (5 minutes).

Second, there may be specific maximum unit to show - "50000" meters is "50km", and "5000000" meters is "5000km". Same for "12000000" minutes being "50000h".
[...]

PR #782, opened 2023-05-09
Editorial: added Properties of Intl.Locale Instances section

### LINKED ISSUE #574: Add "Properties of Intl.Locale Instances" section
There should be a "Properties of Intl.Locale Instances" section describing the internal slots and default prototype of `Intl.Locale` objects.

ben-allen:
fix #574


spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+
+    <p>
+      Intl.Locale instances also have several internal slots that are computed by the constructor:
+    </p>
+
+    <ul>
+      <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
+      <li>[[Calendar]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the calendar used for formatting.</li>
+      <li>[[Collation]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the collation, except that the values *"standard"* and *"search"* are not allowed, while the value *"default"* is allowed.</li>
+      <li>[[HourCycle]] is a String value indicating whether the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*) should be used. *"h11"* and *"h23"* start with hour 0 and go up to 11 and 23 respectively. *"h12"* and *"h24"* start with hour 1 and go up to 12 and 24.</li>
```
ryzokuken, line 373: Not sure how redundant we are with this, maybe we could eventually unify the meaning of all the hourCycle values in the spec somewhere?

ryzokuken (approved):
Looks great, thanks Ben!

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      <li>[[NumberingSystem]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the numbering system used for formatting.</li>
+    </ul>
+    <p>
+      Intl.Locale instances also have the following internal slots that are computed by the constructor if the [[RelevantExtensionKeys]] internal slot of Intl.Locale contains *"kn"* or *"kf"*:
```
gibson042, line 377:
```suggestion
      If the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kn"* and/or *"kf"*, Intl.Locale instances will also have the following internal slots:
```

gibson042 (changes requested):
Thanks! I think we need a few tweaks, but this is definitely a step forward.

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+    </p>
+
+    <p>
+      Intl.Locale instances also have several internal slots that are computed by the constructor:
+    </p>
+
+    <ul>
+      <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
+      <li>[[Calendar]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the calendar used for formatting.</li>
+      <li>[[Collation]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the collation, except that the values *"standard"* and *"search"* are not allowed, while the value *"default"* is allowed.</li>
```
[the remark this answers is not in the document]
ben-allen: I may be either overthinking or underthinking this one. Reasoning:  In Intl.Collator, "search" and "standard" aren't allowed as options for "co" and aren't allowed as values for [[Collation]]. They were banned in Intl.Collator in order to force people to use the `usage` option. Implementations of Intl.Collator appear to respect that constraint. See https://github.com/tc39/ecma402/issues/256#issuecomment-411421731 for a discussion of this.

Since the value of Intl.Collator's [[Collation]] slot can't be "search" or "standard", does it make sense for the analogous slot in Intl.Locale to allow them? I admit that the answer here might just be "well of *course* it should allow those values." 

One argument for not including that constraint for the present moment might be that if the constraint is imposed,  the correction of an editorial oversight would result in a normative change.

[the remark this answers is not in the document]
ben-allen: Upon further investigation: The MDN entry for Intl.Locale.prototype.getCollations() currently has a warning against use of "search" that mentions the constraint in Intl.Collator.  A warning also exists for "standard", but because of collation naming inconsistencies in Swedish, which I presume is what motivated the ban in "standard" in Intl.Collator.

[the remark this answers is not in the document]
gibson042: Presumably yes. But let's keep this PR editorial and open an issue for whether or not Intl.Locale should align with Intl.Collator on this point (for which I can imagine arguments in favor of either resolution).

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      Intl.Locale instances have an [[InitializedLocale]] internal slot.
+    </p>
+
+    <p>
+      Intl.Locale instances also have several internal slots that are computed by the constructor:
+    </p>
+
+    <ul>
+      <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
+      <li>[[Calendar]] is a String value with the *"type"* given in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a> for the calendar used for formatting.</li>
```
FrankYFTang, line 371: " or *undefined*".
Notice, this is different from DateTimeFormat because in DateTimeFormat  `[[Calendar]]`  is from the result of ResolveLocale() which could be all String, but in here the  `[[Calendar]]` could be *undefined* not a String, right.
For example, if we have the following

```
let l = new Intl.Locale("en");
l.calendar; // This should be undefined
```

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      <li>[[Collation]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the collation, except that the values *"standard"* and *"search"* are not allowed, while the value *"default"* is allowed.</li>
```
FrankYFTang, line 372: same as above, it could be *undefined*.

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      <li>[[HourCycle]] is a String value indicating whether the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*) should be used. *"h11"* and *"h23"* start with hour 0 and go up to 11 and 23 respectively. *"h12"* and *"h24"* start with hour 1 and go up to 12 and 24.</li>
```
FrankYFTang, line 373: same as above, it could be *undefined*.

spec/locale.html
```diff
@@ -350,4 +350,35 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      <li>[[NumberingSystem]] is a String value with the *"type"* given in Unicode Technical Standard #35 for the numbering system used for formatting.</li>
```
FrankYFTang, line 374: same as above, it could be *undefined*.

FrankYFTang (changes requested):

FrankYFTang:
as the comments I left, It think the internal slots in Intl.Locale are different from other Intl objects. Most of the fields should be *undefined* (except `[[Locale]`) most of the time but not required to be String. So the description should always include the value *undefined* as a possible value for those fields. (except `[[Locale]]`)

spec/locale.html
ben-allen: I've gone through and verified/fixed these. Thanks for the heads-up!

FrankYFTang (approved):
LGTM

spec/locale.html
```diff
@@ -353,4 +353,36 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+
+    <p>
+      Intl.Locale instances also have several internal slots that are computed by the constructor:
+    </p>
+
+    <ul>
+      <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
+      <li>[[Calendar]] is a String value with the *"type"* given in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a> for the calendar used for formatting, or is *undefined*.</li>
+      <li>[[Collation]] is a String value with the *"type"* given in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a> for the collation, or is the String value *"default"*, or is *undefined*.</li>
+      <li>[[HourCycle]] is a String value indicating whether the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*) should be used, or is *undefined*.</li>
```
FrankYFTang, line 376: Sorry, I just realize the spec (and all the implementation therefore) is currently written in a way that the [[HourCycle]] may be string other than *"h11"*, *"h12"* , *"h23"*, *"h24"*

For example , in the following code
```
let a = new Intl.Locale("en-u-hc-abc")
```
[[HourCycle]] will be "abc".

spec/locale.html
```diff
@@ -353,4 +353,36 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+      <li>[[NumberingSystem]] is a String value with the *"type"* given in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a> for the numbering system used for formatting, or is *undefined*.</li>
+    </ul>
+    <p>
+      If the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kn"* and/or *"kf"*, Intl.Locale instances will also have the following internal slots:
+    </p>
+
+    <ul>
+      <li>[[CaseFirst]] is one of the String values *"upper"*, *"lower"*, or *"false"*, or is *undefined*. This internal slot only exists if the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kf"*.</li>
```
FrankYFTang, line 384: I also just realize [[CaseFirst]] could contains not just *undefined* or String *"upper"*, *"lower"*, or *"false"* but other strings which fit "with the *"type"* given in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a> "

For example, in the following code:
```
let b = new Intl.Locale("en-u-kf-efg")
```
[[CaseFirst]] is "efg".

FrankYFTang (changes requested):

spec/locale.html
ben-allen: ah, thank you! looking at it now

FrankYFTang:
sorry, I should realize this earlier. The design of the Locale is different from other Intl Object because it try to only did syntatic checking (but not sematic checking) of the Locale Identifier, so the value in the Locale string is synatic well formed . The values in the option bag is checked to fit a specific set.  so anything fit with Syntax defined in UTS 35 will not cause the exception throw when passing in the Intl.Locale constructor and value could be preserved since the js may just passing that value in/out from external system without constructing any other functional facility in the system and we do not want to see it got modify. For example, if we have a string "en-u-nu-Abcd" even "Abcd" is not a numberingSystem known by the system, we want to make sure the l.numberingSystem still return "Abcd" even the system does not know how that numberingSystem work

spec/locale.html
```diff
@@ -353,4 +353,37 @@ <h1>get Intl.Locale.prototype.region</h1>
...
+    </ul>
+
+    <p>
+      If the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kn"* and/or *"kf"*, Intl.Locale instances will also have the following internal slots:
+    </p>
+
+    <ul>
+      <li>[[CaseFirst]] is a String value that is a syntactically valid type value as given in <a href="https://unicode.org/reports/tr35/#Unicode_locale_identifier">Unicode Technical Standard #35 Part 1 Core, Section 3.2 Unicode Locale Identifier</a>, or is *undefined*.</li>
+      <li>[[Numeric]] is a Boolean value specifying whether numeric sorting is used by the locale, or is *undefined*. This internal slot only exists if the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kn"*.</li>
+    </ul>
```
gibson042, line 387: Might as well combine the lists and individually annotate the items for optional slots.

```suggestion
      <li>[[CaseFirst]] is a String value that is a syntactically valid type value as given in <a href="https://unicode.org/reports/tr35/#Unicode_locale_identifier">Unicode Technical Standard #35 Part 1 Core, Section 3.2 Unicode Locale Identifier</a>, or is *undefined*. This internal slot only exists if the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kf"*.</li>
      <li>[[Numeric]] is a Boolean value specifying whether numeric sorting is used by the locale, or is *undefined*. This internal slot only exists if the [[RelevantExtensionKeys]] internal slot of %Locale% contains *"kn"*.</li>
    </ul>
```

gibson042 (approved):

FrankYFTang (approved):
LGTM

PR #783, opened 2023-05-11
Normative: Added support for sentence break suppressions to Intl.Segmenter

### LINKED ISSUE #580: Add support for more subtags
Currently, a number of Unicode extension keywords are supported, including `nu` for numbering systems, `ca` for calendaring systems, `hc` for hour cycles, etc. 

This issue tracks all the BCP-47 subtags added by Unicode that we do not yet support. Feel free to add to the list below this:

1. `rg` (Region override): see #370.
2. `fw` (First day of the week): see #6.

ben-allen:
fix #580 

Added support for u-ss to Intl.Segmenter

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -19,7 +19,7 @@
       <emu-alg>
         1. If NewTarget is *undefined*, throw a *TypeError* exception.
         1. Let _relevantExtensionKeys_ be %Locale%.[[RelevantExtensionKeys]].
-        1. Let _internalSlotsList_ be &laquo; [[InitializedLocale]], [[Locale]], [[Calendar]], [[Collation]], [[HourCycle]], [[NumberingSystem]] &raquo;.
+        1. Let _internalSlotsList_ be &laquo; [[InitializedLocale]], [[Locale]], [[Calendar]], [[Collation]], [[HourCycle]], [[NumberingSystem]], [[SentenceBreakSuppressions]] &raquo;.
         1. If _relevantExtensionKeys_ contains *"kf"*, then
           1. Append [[CaseFirst]] as the last element of _internalSlotsList_.
         1. If _relevantExtensionKeys_ contains *"kn"*, then
@@ -52,6 +52,8 @@
         1. If _numberingSystem_ is not *undefined*, then
           1. If _numberingSystem_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
         1. Set _opt_.[[nu]] to _numberingSystem_.
+        1. Let _sentenceBreakSuppressions_ be ? GetOption(_options_, *"sentenceBreakSuppressions"*, ~string~, &laquo;*"none"*, *"standard"* &raquo;, *"none"*).
+        1. Set _opt_.[[ss]] to _sentenceBreakSuppressions_.
         1. Let _r_ be ! ApplyUnicodeExtensionToTag(_tag_, _opt_, _relevantExtensionKeys_).
         1. Set _locale_.[[Locale]] to _r_.[[locale]].
         1. Set _locale_.[[Calendar]] to _r_.[[ca]].
@@ -65,6 +67,7 @@
           1. Else,
             1. Set _locale_.[[Numeric]] to *false*.
         1. Set _locale_.[[NumberingSystem]] to _r_.[[nu]].
+        1. Set _locale_.[[SentenceBreakSuppressions]] to _r_.[[ss]].
         1. Return _locale_.
       </emu-alg>
     </emu-clause>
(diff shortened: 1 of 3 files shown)


spec/segmenter.html
```diff
@@ -74,11 +77,8 @@ <h1>Internal slots</h1>
       </p>
 
       <p>
-        The value of the [[RelevantExtensionKeys]] internal slot is &laquo; &raquo;.
+        The value of the [[RelevantExtensionKeys]] internal slot is &laquo; *"ss"* &raquo;.
       </p>
-      <emu-note>
-        Intl.Segmenter does not have any relevant extension keys.
-      </emu-note>
```
srl295, line 81: There are actually others: lb, lw, dx. See https://www.unicode.org/reports/tr35/#UnicodeLineBreakStyleIdentifier

srl295 (approved):
LGTM

spec/segmenter.html
sffc: lb and lw are only for line break, which Intl.Segmenter doesn't support.

dx is not widely used or implemented, and I've raised questions about its utility.

spec/segmenter.html
```diff
@@ -17,16 +17,19 @@ <h1>Intl.Segmenter ( [ _locales_ [ , _options_ ] ] )</h1>
...
         1. Set _options_ to ? GetOptionsObject(_options_).
         1. Let _opt_ be a new Record.
         1. Let _matcher_ be ? GetOption(_options_, *"localeMatcher"*, ~string~, &laquo; *"lookup"*, *"best fit"* &raquo;, *"best fit"*).
         1. Set _opt_.[[localeMatcher]] to _matcher_.
+        1. Let _sentenceBreakSuppressions_ be ? GetOption(_options_, *"sentenceBreakSuppressions"*, ~string~, &laquo; *"none"*, *"standard"* &raquo;, *"none"*).
+        1. Set _opt_.[[ss]] to _sentenceBreakSuppressions_.
         1. Let _localeData_ be %Segmenter%.[[LocaleData]].
         1. Let _r_ be ResolveLocale(%Segmenter%.[[AvailableLocales]], _requestedLocales_, _opt_, %Segmenter%.[[RelevantExtensionKeys]], _localeData_).
         1. Set _segmenter_.[[Locale]] to _r_.[[locale]].
+        1. Set _segmenter_.[[SentenceBreakSuppressions]] to _sentenceBreakSuppressions_.
```
sffc, line 32: I think you need to read from `_r_` here

```suggestion
        1. Set _segmenter_.[[SentenceBreakSuppressions]] to _r_.[[ss]].
```

ryzokuken (approved):
spec changes LGTM, none of the three mentioned extension keys are supported in `Intl` yet and we shouldn't introduce more than one in a single PR so it looks good to go from my end. Let's see how people feel about this.

FrankYFTang:
Sorry, I think we need additional changes that I forget to mention in TG2. 

A. We should also change 
14.1.1 Intl.Locale ( tag [ , options ] )
https://tc39.es/ecma402/#sec-Intl.Locale
to take option
"sentenceBreakSuppressions" 

B. The internal slot of Locale should also include "ss"
https://tc39.es/ecma402/#sec-intl.locale-internal-slots

C. We should also add a getter in section 14.3 
https://tc39.es/ecma402/#sec-properties-of-intl-locale-prototype-object
14.3.X get Intl.Locale.prototype.sentenceBreakSuppressions

which return either "none" (default) or "standard")

ben-allen:
Updated to incorporate required changes to Intl.Locale & include a note in Annex A. Unclear on needed changes to Annex B.

spec/locale.html
```diff
@@ -52,6 +52,8 @@ <h1>Intl.Locale ( _tag_ [ , _options_ ] )</h1>
         1. If _numberingSystem_ is not *undefined*, then
           1. If _numberingSystem_ does not match the Unicode Locale Identifier `type` nonterminal, throw a *RangeError* exception.
         1. Set _opt_.[[nu]] to _numberingSystem_.
+        1. Let _sentenceBreakSuppressions_ be ? GetOption(_options_, *"sentenceBreakSuppressions"*, ~string~, &laquo;*"none"*, *"standard"* &raquo;, *"none"*).
```
FrankYFTang, line 55: For Intl.Locale constructor, the default value for "sentenceBreakSuppressions" should be undefined. (Comparing to other options, such as "calendar", "collation", "hourCycle", "caseFirst", "numeric", "numberingSystem")

sffc:
Note: @dminor would like to wait until ICU4X supports this feature before merging this PR.

https://github.com/unicode-org/icu4x/issues/3927

PR #786, opened 2023-05-31
Normative: raised minimum/maximum fractional digits from 20 to 100

### LINKED ISSUE #585: Limit for maximumFractionDigits option for decimal strings formatting
This issue was originally reported as https://github.com/tc39/proposal-intl-numberformat-v3/issues/34 by @Yaffle.

@Yaffle said:

> I have some "numeric calculator". I am using "bigdecimal" library and so do the rounding (to specified number of digits) using the methods of that library (BigDecimal#toFixed(fractionDigits)). I am dealing with minus sign and exponents by my code (as I am trying to use MathML anyway). So NumberFormat need only convert the digits, and so maximumFractionDigits should be 2**53-1 or Infinity.
>
> For now, I can still do the converion in JavaScript itself, as the current NumberFormat allows to get information about digits and decimal separator.

@sffc said:

> The main limitation with maximumFractionDigits is that it gives us a reasonable way to limit the maximum length of the string returned from Intl.NumberFormat. If we were to raise the cap on maximumFractionDigits to something extremely large, it could be a security vulnerability, because it would become easy to make your computer run out of memory.
>
> So in my opinion, it should be fairly harmless to increase the cap to, say, 100, 255, or perhaps a little bit larger if we have a clear use case, but increasing it to an extremely large number or Infinity voids the benefits that the cap serves in the first place.

@chicoxyzzy said:

> I never needed maximumFractionDigits to be more than 20 back when I was working in a FinTech company, but I can imagine cases when limit increase could be useful.
>
> https://brilliant.org/wiki/very-small-numbers/ - see "Important Very Small Numbers" part here

@gibson042 said during discussion:

> We should be as capable as 262, but going beyond is not necessary. Looks like 100: https://tc39.es/ecma262/#sec-number.prototype.tofixed

Additional meeting discussion notes: https://github.com/tc39/ecma402/blob/master/meetings/notes-2021-07-01.md#limit-for-maximumfractiondigits-option-for-decimal-strings-formatting

ben-allen:
fix #585 

Unclear what the consensus was on increasing max fractional digits to match 262, but if we're doing it this PR does it.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -146,8 +146,8 @@
             1. Set _intlObj_.[[MaximumSignificantDigits]] to 21.
         1. If _needFd_ is *true*, then
           1. If _hasFd_ is *true*, then
-            1. Set _mnfd_ to ? DefaultNumberOption(_mnfd_, 0, 20, *undefined*).
-            1. Set _mxfd_ to ? DefaultNumberOption(_mxfd_, 0, 20, *undefined*).
+            1. Set _mnfd_ to ? DefaultNumberOption(_mnfd_, 0, 100, *undefined*).
+            1. Set _mxfd_ to ? DefaultNumberOption(_mxfd_, 0, 100, *undefined*).
             1. If _mnfd_ is *undefined*, set _mnfd_ to min(_mnfdDefault_, _mxfd_).
             1. Else if _mxfd_ is *undefined*, set _mxfd_ to max(_mxfdDefault_, _mnfd_).
             1. Else if _mnfd_ is greater than _mxfd_, throw a *RangeError* exception.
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):
spec change LGTM, let's check what the consensus is

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-01.md#normative-raised-maximum-fractional-digits-from-20-to-100-786

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-29.md#normative-raised-minimummaximum-fractional-digits-from-20-to-100-786-1

FrankYFTang:
v8 tracking bug https://bugs.chromium.org/p/v8/issues/detail?id=14170

FrankYFTang:
v8 change cl https://chromium-review.googlesource.com/c/v8/v8/+/4679294

FrankYFTang:
The landing of https://chromium-review.googlesource.com/c/v8/v8/+/4679294 is blocked by the merging of this PR. 
Also, it seems there are currently no test262 test validate this PR yet.

FrankYFTang:
test262 tests in https://github.com/tc39/test262/pull/3875

FrankYFTang (approved):
Since this is already reach consensus in TC39, it will be nice if you can land this ASAP. Thanks

FrankYFTang:
thanks for merging this.

PR #788, opened 2023-06-01
Normative: Allow UTC offset time zones

### LINKED ISSUE #683: ECMA-402 should allow numeric-offset host time zones
Among other things, this would remove the need for ECMA-402 Temporal to override ECMA-262 DefaultTimeZone.

gibson042:
Fixes #683

This aligns with ECMA-262, and therefore removes the DefaultTimeZone override.

Example time zones that will now be accepted: [Temporal |TimeZoneUTCOffsetName|](https://tc39.es/proposal-temporal/#prod-TimeZoneUTCOffsetName)
* "+00"
* "-00"
* "−00"
* "+0000"
* "-0000"
* "−0000"
* "+00:00"
* "-00:00"
* "−00:00"
* "+2359"
* "+23:59"
* "-2359"
* "-23:59"
* "−2359"
* "−23:59"

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -679,7 +679,7 @@
       <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
       <li>[[Calendar]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier">Unicode Calendar Identifier</a> used for formatting.</li>
       <li>[[NumberingSystem]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeNumberSystemIdentifier">Unicode Number System Identifier</a> used for formatting.</li>
-      <li>[[TimeZone]] is a String value that is a time zone identifier from the IANA Time Zone Database used for formatting.</li>
+      <li>[[TimeZone]] is a String value used for formatting that is either a time zone identifier from the IANA Time Zone Database or a UTC offset in ISO 8601 extended format.</li>
       <li>[[Weekday]], [[Era]], [[Year]], [[Month]], [[Day]], [[DayPeriod]], [[Hour]], [[Minute]], [[Second]], [[TimeZoneName]] are each either *undefined*, indicating that the component is not used for formatting, or one of the String values given in <emu-xref href="#table-datetimeformat-components"></emu-xref>, indicating how the component should be presented in the formatted output.</li>
       <li>[[FractionalSecondDigits]] is either *undefined* or a positive, non-zero integer Number value indicating the fraction digits to be used for fractional seconds. Numbers will be rounded or padded with trailing zeroes if necessary.</li>
       <li>[[HourCycle]] is a String value indicating whether the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*) should be used. *"h11"* and *"h23"* start with hour 0 and go up to 11 and 23 respectively. *"h12"* and *"h24"* start with hour 1 and go up to 12 and 24. [[HourCycle]] is only used when [[Hour]] is not *undefined*.</li>
@@ -1150,7 +1150,11 @@
       </dl>
 
       <emu-alg>
-        1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_timeZoneIdentifier_, _epochNs_).
+        1. If IsTimeZoneOffsetString(_timeZoneIdentifier_) is *true*, then
+          1. Let _offsetNs_ be ParseTimeZoneOffsetString(_timeZoneIdentifier_).
+        1. Else,
+          1. Assert: IsValidTimeZoneName(_timeZoneIdentifier_) is *true*.
+          1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_timeZoneIdentifier_, _epochNs_).
         1. Let _tz_ be ℝ(_epochNs_) + _offsetNs_.
         1. If _calendar_ is *"gregory"*, then
           1. Return a record with fields calculated from _tz_ according to <emu-xref href="#table-datetimeformat-tolocaltime-record"></emu-xref>.
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -94,9 +94,13 @@
           1. Set _timeZone_ to DefaultTimeZone().
         1. Else,
           1. Set _timeZone_ to ? ToString(_timeZone_).
-          1. If the result of IsValidTimeZoneName(_timeZone_) is *false*, then
-            1. Throw a *RangeError* exception.
+        1. If IsTimeZoneOffsetString(_timeZone_) is *true*, then
+          1. Let _offsetNanoseconds_ be ParseTimeZoneOffsetString(_timeZone_).
+          1. Set _timeZone_ to FormatTimeZoneOffsetString(_offsetNanoseconds_).
+        1. Else if IsValidTimeZoneName(_timeZone_) is *true*, then
           1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZone_).
+        1. Else,
+          1. Throw a *RangeError* exception.
         1. Set _dateTimeFormat_.[[TimeZone]] to _timeZone_.
         1. Let _formatOptions_ be a new Record.
         1. Set _formatOptions_.[[hourCycle]] to _hc_.
@@ -162,6 +166,31 @@
         1. Set _dateTimeFormat_.[[Pattern]] to _pattern_.
         1. Set _dateTimeFormat_.[[RangePatterns]] to _rangePatterns_.
         1. Return _dateTimeFormat_.
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-formattimezoneoffsetstring" aoid="FormatTimeZoneOffsetString">
+      <h1>FormatTimeZoneOffsetString ( _offsetNanoseconds_ )</h1>
+      <emu-alg>
+        1. Assert: _offsetNanoseconds_ is an integer.
+        1. If _offsetNanoseconds_ &ge; 0, let _sign_ be the code unit 0x002B (PLUS SIGN); otherwise, let _sign_ be the code unit 0x002D (HYPHEN-MINUS).
+        1. Set _offsetNanoseconds_ to abs(_offsetNanoseconds_).
+        1. Let _nanoseconds_ be _offsetNanoseconds_ modulo 10<sup>9</sup>.
+        1. Let _seconds_ be floor(_offsetNanoseconds_ / 10<sup>9</sup>) modulo 60.
+        1. Let _minutes_ be floor(_offsetNanoseconds_ / (6 &times; 10<sup>10</sup>)) modulo 60.
+        1. Let _hours_ be floor(_offsetNanoseconds_ / (3.6 &times; 10<sup>12</sup>)).
+        1. Let _h_ be ToZeroPaddedDecimalString(_hours_, 2).
+        1. Let _m_ be ToZeroPaddedDecimalString(_minutes_, 2).
+        1. Let _s_ be ToZeroPaddedDecimalString(_seconds_, 2).
+        1. If _nanoseconds_ &ne; 0, then
+          1. Let _fraction_ be ToZeroPaddedDecimalString(_nanoseconds_, 9).
+          1. Set _fraction_ to the longest possible substring of _fraction_ starting at position 0 and not ending with the code unit 0x0030 (DIGIT ZERO).
+          1. Let _post_ be the string-concatenation of the code unit 0x003A (COLON), _s_, the code unit 0x002E (FULL STOP), and _fraction_.
+        1. Else if seconds &ne; 0, then
+          1. Let _post_ be the string-concatenation of the code unit 0x003A (COLON) and _s_.
+        1. Else,
+          1. Let _post_ be the empty String.
+        1. Return the string-concatenation of _sign_, _h_, the code unit 0x003A (COLON), _m_, and _post_.
       </emu-alg>
     </emu-clause>
   </emu-clause>
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -95,8 +95,13 @@
         1. Else,
           1. Set _timeZone_ to ? ToString(_timeZone_).
         1. If IsTimeZoneOffsetString(_timeZone_) is *true*, then
+          1. Let _parseResult_ be ParseText(StringToCodePoints(_timeZone_), |UTCOffset|).
+          1. Assert: _parseResult_ is a Parse Node.
+          1. If _parseResult_ contains more than one |MinuteSecond| Parse Node, throw a *RangeError* exception.
           1. Let _offsetNanoseconds_ be ParseTimeZoneOffsetString(_timeZone_).
-          1. Set _timeZone_ to FormatTimeZoneOffsetString(_offsetNanoseconds_).
+          1. Let _offsetMinutes_ be _offsetNanoseconds_ / (6 × 10<sup>10</sup>).
+          1. Assert: _offsetMinutes_ is an integer.
+          1. Set _timeZone_ to FormatOffsetTimeZoneIdentifier(_offsetMinutes_).
         1. Else if IsValidTimeZoneName(_timeZone_) is *true*, then
           1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZone_).
         1. Else,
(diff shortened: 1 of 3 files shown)


spec/datetimeformat.html
```diff
@@ -142,6 +146,31 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
         1. Return _dateTimeFormat_.
       </emu-alg>
     </emu-clause>
+
+    <emu-clause id="sec-formattimezoneoffsetstring" aoid="FormatTimeZoneOffsetString">
+      <h1>FormatTimeZoneOffsetString ( _offsetNanoseconds_ )</h1>
+      <emu-alg>
+        1. Assert: _offsetNanoseconds_ is an integer.
```
ptomato, line 153: I'd recommend a structured header (and then there's no need for the assertion). We can also limit the type of _offsetNanoseconds_ to "an integer in the range of -86,400,000,000,000 to 86,400,000,000,000". (I forget whether the range is inclusive or exclusive at both ends.)

spec/datetimeformat.html
```diff
@@ -91,9 +91,13 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
           1. Set _timeZone_ to DefaultTimeZone().
         1. Else,
           1. Set _timeZone_ to ? ToString(_timeZone_).
-          1. If the result of IsValidTimeZoneName(_timeZone_) is *false*, then
-            1. Throw a *RangeError* exception.
+        1. If IsTimeZoneOffsetString(_timeZone_) is *true*, then
+          1. Let _offsetNanoseconds_ be ParseTimeZoneOffsetString(_timeZone_).
```
justingrant, line 95: There's a new CanonicalizeTimeZoneOffsetString AO in Temporal. Should this be brought into 402 to replace this line and the one below it?

justingrant (approved):
Looks great!

spec/datetimeformat.html
gibson042: This operation is defined in Temporal and will be removed from ECMA-402 and/or this PR once added to ECMA-262. I will propagate any upstream changes.

gibson042:
> so.... where is the nanosecond requirement?

It will be discussed upstream at https://github.com/tc39/ecma262/pull/3087 and https://github.com/tc39/proposal-temporal/issues/2593 .

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-01.md#normative-allow-utc-offset-time-zones-788

We need to resolve the minutes/seconds/centiseconds/nanoseconds first, but it seems okay to the group besides that point, which was somewhat controversial for a number of reasons.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-06-29.md#normative-allow-utc-offset-timezones

We'd like to wait until the Temporal changes land and then revisit this PR after it has been updated.

FrankYFTang:
BTW, I do not understand why Justin try to support UTC offset time zone for "system default time zone". I totally understand the need for UTC offset time zone for other usage, such as passin in into Intl.DateTimeFormat as the timeZone value, but I do not believe we know any OS output that and any other software support that as system timezone value, right? Where is that need come from? (I am specifically talking about the fact that Temporal change DefaultTimeZone  as defined as "a valid ([6.5.1](https://tc39.es/ecma402/#sec-isvalidtimezonename)) and canonicalized ([6.5.2](https://tc39.es/ecma402/#sec-canonicalizetimezonename)) time zone name." to add "or an [offset time zone](https://tc39.es/proposal-temporal/#sec-time-zone-identifiers) identifier" to it in
https://tc39.es/proposal-temporal/#sec-systemtimezoneidentifier

gibson042:
@sffc Let's please also discuss it at the next TG2 meeting regardless.

FrankYFTang:
> @sffc Let's please also discuss it at the next TG2 meeting regardless.

added to https://github.com/tc39/ecma402/projects/2

ryzokuken (approved):

sffc:
Do we support formatting these with timeZoneName "shortOffset" and "longOffset"?

We should think about the formatting fallback given the six time zone format options: "short", "long", "shortOffset", "longOffset", "shortGeneric", "longGeneric". We discussed the fallback between these at length in the time zone format options proposal a couple of years ago. It shows up in BasicFormatMatcher: https://tc39.es/ecma402/#sec-basicformatmatcher

FrankYFTang:
Please write some example of the timeZone which previous throw but after this PR will be accepted . Thanks

gibson042:
PR updated to remove support for UTC offsets with sub-minute precision, just like Temporal.

gibson042:
> Please write some example of the timeZone which previous throw but after this PR will be accepted . Thanks

@FrankYFTang added to the PR description.

sffc:
TG2 approval pending additional review of the spec text: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-09-07.md#normative-allow-utc-offset-time-zones-788

sffc (approved):

spec/datetimeformat.html
```diff
@@ -94,9 +94,18 @@ <h1>
           1. Set _timeZone_ to DefaultTimeZone().
         1. Else,
           1. Set _timeZone_ to ? ToString(_timeZone_).
-          1. If the result of IsValidTimeZoneName(_timeZone_) is *false*, then
-            1. Throw a *RangeError* exception.
+        1. If IsTimeZoneOffsetString(_timeZone_) is *true*, then
+          1. Let _parseResult_ be ParseText(StringToCodePoints(_timeZone_), |UTCOffset|).
+          1. Assert: _parseResult_ is a Parse Node.
+          1. If _parseResult_ contains more than one |MinuteSecond| Parse Node, throw a *RangeError* exception.
```
FrankYFTang, line 100: I am not sure how does "contains more than one |MinuteSecond| Parse Node" work here.
How is the "contains" operation defined for Parse Node. I cannot find a concrete definition. 

Look at the UTCOffset grammar in https://tc39.es/ecma262/#sec-time-zone-offset-string-format
```
UTCOffset :::
TemporalSign Hour
TemporalSign Hour HourSubcomponents[+Extended]
TemporalSign Hour HourSubcomponents[~Extended]
```

So if the concept of "contains" is shallow, then it would be impossible for UTCOffset to contain any MinuteSecond.
since 1 or 2 MinuteSecond could be contain inside the HourSubcomponents that contained by UTCOffset. However, the require a "deep containment" interpretation and I am not sure that is what ECMA262 specify. (since I cannot find a clear definition of what  "contains" mean for Parse Node.

gibson042: @FrankYFTang This use of "contains" is prose of the sort that is also currently in Temporal [ParseTimeZoneIdentifier](https://tc39.es/proposal-temporal/#sec-parsetimezoneidentifier)—as suggested by the lack of automatic hyperlinking, there is no concrete definition. However, as I thought would be clear from context in both documents, it is intended to be understood as entailing deep inspection of the tree of Parse Nodes. Can you suggest a change here that would make that more clear?

I could define a precise syntax-directed operation, but **really** don't want to do that because it would apply cross-document to the [ECMA-262 |UTCOffset| nonterminal](https://tc39.es/ecma262/multipage/numbers-and-dates.html#prod-UTCOffset) that will be _replaced_ by [Temporal |TimeZoneUTCOffsetName|](https://tc39.es/proposal-temporal/#prod-TimeZoneUTCOffsetName) (creating a situation where this part of ECMA-402 would be incoherent until fixed).

It's also worth noting that this explicit check for multiple |MinuteSecond| Parse Nodes will itself be removed when Temporal lands, along with many of the other steps here that are subsumed by [ParseTimeZoneIdentifier](https://tc39.es/proposal-temporal/#sec-parsetimezoneidentifier):
```patch
-        1. If IsTimeZoneOffsetString(_timeZone_) is *true*, then
+        1. Let _timeZoneRecord_ be ? ParseTimeZoneIdentifier(_timeZone_).
+        1. If _timeZoneRecord_.[[Name]] is ~empty~, then
-          1. Let _parseResult_ be ParseText(StringToCodePoints(_timeZone_), |UTCOffset|).
-          1. Assert: _parseResult_ is a Parse Node.
-          1. If _parseResult_ contains more than one |MinuteSecond| Parse Node, throw a *RangeError* exception.
-          1. Let _offsetNanoseconds_ be ParseTimeZoneOffsetString(_timeZone_).
-          1. Let _offsetMinutes_ be _offsetNanoseconds_ / (6 × 10<sup>10</sup>).
-          1. Assert: _offsetMinutes_ is an integer.
+          1. Let _offsetMinutes_ be _timeZoneRecord_.[[OffsetMinutes]].
           1. Set _timeZone_ to FormatOffsetTimeZoneIdentifier(_offsetMinutes_).
-        1. Else if IsValidTimeZoneName(_timeZone_) is *true*, then
+        1. Else if IsValidTimeZoneName(_timeZoneRecord_.[[Name]]) is *true*, then
           1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZoneRecord_.[[Name]]).
         1. Else,
           1. Throw a *RangeError* exception.
```

spec/datetimeformat.html
FrankYFTang: I do not change the definition of HourSubcomponents in ECMA262 instead
Currently, it is 
```
HourSubcomponents[Extended] :::
  TimeSeparator[?Extended] MinuteSecond
  TimeSeparator[?Extended] MinuteSecond TimeSeparator[?Extended] MinuteSecond TemporalDecimalFractionopt
```
why not change it to just
```
HourSubcomponents[Extended] :::
  TimeSeparator[?Extended] MinuteSecond
```

?

FrankYFTang: The only place reference to  it is  UTCOffset and the only places refence to UTCOffset are 
21.4.1.33.1 IsTimeZoneOffsetString ( offsetString )
and
21.4.1.33.2 ParseTimeZoneOffsetString ( offsetString )

which are exactly the two places we need to change and both places are only used by place dealing with system timezone.

FrankYFTang:
I found another problem,  UTCOffset ::: require the supporting of U+2212 as the minus sign. beside the ASCII '+' and '-'

What is the reason for such requirement?
https://tc39.es/ecma262/#sec-time-zone-offset-strings
states
'ECMAScript defines a string interchange format for UTC offsets, derived from ISO 8601. '

But in https://www.ietf.org/archive/id/draft-ietf-sedate-datetime-extended-09.html#section-4.1
it point the grammar to https://www.rfc-editor.org/rfc/rfc3339#section-5.6
as
```
   time-numoffset  = ("+" / "-") time-hour ":" time-minute
```
which does not include U+2212 in the syntax

Can someone explain to me where is the requirement of U+2212 come from?

[some doc pointed to 
](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar) point to 
https://datatracker.ietf.org/doc/html/rfc3339#appendix-A 
But I see nowhere in RFC3339 mention U+2212


Could someone explain to me where is the requirement of  U+2212 came from?

FrankYFTang:
'−00:03'
* "-00"
* "−00"
* "+0000"
* "-0000"
* "−0000"
* "+00:00"
* "-00:00"
* "−00:00"
* "+2359"
* "+23:59"
* "-2359"
* "-23:59"
* "−2359"
* "−23:59"

FrankYFTang:
sorry, I click the wrong bottom.

FrankYFTang:
Trying to implement in v8 https://chromium-review.googlesource.com/c/v8/v8/+/4854730

spec/datetimeformat.html
gibson042: @FrankYFTang An ECMA-262 grammar cannot be changed from an ECMA-402 pull request. But all of this _will_ be simplified by/in response to Temporal, as demonstrated in https://github.com/tc39/ecma402/pull/788#discussion_r1320356161 .

gibson042:
> I found another problem, UTCOffset ::: require the supporting of U+2212 as the minus sign. beside the ASCII '+' and '-'
> 
> What is the reason for such requirement? https://tc39.es/ecma262/#sec-time-zone-offset-strings states 'ECMAScript defines a string interchange format for UTC offsets, derived from ISO 8601. '
> 
> But in https://www.ietf.org/archive/id/draft-ietf-sedate-datetime-extended-09.html#section-4.1 it point the grammar to https://www.rfc-editor.org/rfc/rfc3339#section-5.6 as
> 
> ```
>    time-numoffset  = ("+" / "-") time-hour ":" time-minute
> ```
> 
> which does not include U+2212 in the syntax
> 
> Can someone explain to me where is the requirement of U+2212 come from?

@FrankYFTang see https://github.com/tc39/proposal-temporal/issues/814

spec/datetimeformat.html
FrankYFTang: One thing I do not understand is why do we have this PR for ECMA402 instead of just let Temporal to change both ECMA262 and 402. What is the urgent need for that to happen before Temporal reach stage 4?

FrankYFTang:
How about this, we do not throw based on the UTSOffset parsing result, but throw on the length of the result from calling FormatOffsetTimeZoneIdentifier ? if the length of result from FormatOffsetTimeZoneIdentifier > 6 we throw?

spec/datetimeformat.html
justingrant: AFAIK, the main reason is that offset time zones are currently supported in (pre-Temporal) ECMA-262. My assumption is that all ECMA-262 features should be supported when ECMA-402 is also implemented.

Also, do we know yet whether ICU and/or CLDR changes will be needed to enable formatting of offset time zones?  If no, that's great news. If yes, then where do changes need to be made?

FrankYFTang:
ICU SimpleTimeZone created by TimeZone::createInstance will handle the offset timezone but use id prefix with "GMT"
so it will be "GMT+01:30" or "GMT-05:33" etc. My v8 implementation just check if it is starting with "GMT" and then remove that "GMT" prefix

FrankYFTang:
One problem we have with the implementation is we can take the input timezone as
"+00"
"-00"
"−00"
"+0000"
"-0000"
"−0000"
"+00:00"
"-00:00"
"−00:00"

but we it will be inefficient to distinguish it from "UTC" from the resolvedOptions. Could we treat them as simply as "UTC" in the spec text? (so resolvedOptions().timeZone will return "UTC" for these cases.

FrankYFTang:
please take a look at https://github.com/tc39/test262/pull/3917 
I didn't add tests for 

"+00"
"-00"
"−00"
"+0000"
"-0000"
"−0000"
"+00:00"
"-00:00"
"−00:00"

It need more look into these zero offset one . I am not sure we can easily efficiently implement that in v8 and make it still distinguishable from "UTC". Could we change the spec to fold them into UTC if the offset is just 0?

spec/datetimeformat.html
```diff
@@ -94,9 +94,18 @@ <h1>
...
+          1. Let _offsetNanoseconds_ be ParseTimeZoneOffsetString(_timeZone_).
+          1. Let _offsetMinutes_ be _offsetNanoseconds_ / (6 × 10<sup>10</sup>).
+          1. Assert: _offsetMinutes_ is an integer.
+          1. Set _timeZone_ to FormatOffsetTimeZoneIdentifier(_offsetMinutes_).
```
FrankYFTang, line 104: I can no longer find FormatOffsetTimeZoneIdentifier in ECMA262 now.

FrankYFTang: What I mean is if I have the following
```
let output = (new Intl.DateTimeFormat("en", {timeZone})).resolvedOptions().timeZone
```

If timeZone is "+00:00" "-0000", it will hard for us to get it back as "+00:00" "-0000" in v8 now. Since we depends on information inside ICU to be efficiently hold the nessary and useful information the output will all be "UTC" for them.  I see no practical reason why these could not be fold into "UTC"

FrankYFTang:
Is my understanding correct? for the following code
```
(new Intl.DateTimeFormat(undefined, {timeZone}).resolvedOptions().timeZone
```
should get the following result:
"+00" => "+00:00"
"-00" => "-00:00"
"−00" => "-00:00"
"+0000" => "+00:00"
"-0000" => "-00:00"
"−0000" => "-00:00"
"+00:00" => "+00:00"
"-00:00" => "-00:00"
"−00:00" => "-00:00"
"+2359" => "+23:59"
"+23:59" => "+23:59"
"-2359" => "-23:59"
"-23:59" => "-23:59"
"−2359" => "-23:59"
"−23:59" => "-23:59"

justingrant:
> "-00" => "-00:00"

Nope. -00, -00:00, -0000, +00, and +0000 should all be normalized to +00:00. Here's why:

Time zone ID strings output by ECMAScript methods should always be normalized. For offset zones, normalized means ±HH:MM, with -00:00 not allowed. For IANA time zones, normalized means matching the letter case used in the IANA time zone database.  ECMAScript methods should accept non-normalized formats like +00 or -00:00, but should never output a time zone ID string except in a normalized format.

This normalization behavior is specified in https://tc39.es/proposal-temporal/#sec-time-zone-identifiers, as well as in AO spec text in both Temporal and this PR. (If you find a case where an ECMAScript method is allowed to output a non-normalized ID for a built-in offset or built-in named time zone, then it's a spec bug.)

The reason why we require normalized outputs is so that implementations won't have to store the original ID strings that users provided. Instead, implementations are free to store timezone ID slots in a more optimized way, as long as the normalized ID string can be reconstituted later as needed. (e.g. in `id` or `timeZoneId` getters).

For example, an implementation could optimize storage of a timezone ID slot using a 16-bit union:
* one bit to mark the ID as an offset time zone ID or IANA ID
* if offset, a 12-bit signed number of minutes (from -1439 representing -23:59 to +1439 representing +23:59)
* if IANA, a 10-bit unsigned index into a cached array of the ~600 IANA ID strings

Hopefully this explains why "-00" => "+00:00" is the expected behavior.

spec/datetimeformat.html
justingrant: > I can no longer find FormatOffsetTimeZoneIdentifier in ECMA262 now.

This AO is only in Temporal (https://tc39.es/proposal-temporal/#sec-temporal-formatoffsettimezoneidentifier) and in this PR. This AO serves two purposes:
* Formatting offset IDs when they're needed, e.g. in the TimeZone.p.id getter
* When paired with ID parsing, enables normalization of offset time zone IDs

Once Temporal makes it to Stage 4, we'd file another 402 PR to remove FormatOffsetTimeZoneIdentifier from 402 as part of the refactor described in https://github.com/tc39/ecma402/pull/788#discussion_r1320356161.

But in the meantime, we need it here to support offset time zone ID normalization.

FrankYFTang: Sorry. My bad, I thought FormatOffsetTimeZoneIdentifier is in ECMA262 and forget to search this PR.

justingrant:
> it will be inefficient to distinguish it from "UTC" from the resolvedOptions. Could we treat them as simply as "UTC" in the spec text? (so resolvedOptions().timeZone will return "UTC" for these cases.

I assume that  we want  `new Intl.DateTimeFormat('en', {timeZone}).resolvedOptions().tineZone` to always match `Temporal.TimeZone.from(timeZone).id`. Making them different would be confusing for users. Assuming they should behave the same, then:

1. Normalizing the offset time zone ID  "+00:00" into the IANA time zone ID "UTC" would be a normative change to Temporal and I'm pretty sure also a normative change to pre-Temporal ECMA-262. Without digging into the specs it's hard to know how much churn this change would cause, but the bar for normative changes like this is very high so the default should be to avoid making changes like this. 

2. There's also a user-facing advantage of leaving them separate, because UTC and +00:00 do have a slight semantic difference.  UTC has a special meaning as a "computers-only" time zone; no humans are in a place where "UTC" is their time zone ID.  On the other hand, +00:00 is interpreted to mean an offshore time zone where humans are, like a boat 500km south of Liberia.  So even though those two zones return identical results, it can be helpful for a userland app to know the difference. For example, an app may choose to show a user a warning message, e.g. "No humans live in a place whose time zone ID is 'UTC'. Was this a mistake?". 

So unless there's a very large performance advantage of making this change, then I'd like to leave the current spec as-is so that +00:00 and UTC are treated as distinct time zones, even though they always result in the same numeric results.

FrankYFTang:
I drop my early concern about "+00:00" vs "UTC". It is not a problem.

Louis-Aime:
I understand this PR is about letting an author specify a time zone with a simple string, as an offset to UTC, in the context of Temporal. I withdraw the concern I had during [last meeting](https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-09-07.md). Temporal.TimeZone has all functions for the use cases I had in mind. 

However we should then stick to the documentation for `timeZone.getOffsetString`:

> This method is similar to timeZone.getOffsetNanosecondsFor(), but returns the offset formatted as a string, with sign, hours, and minutes.

The present polyfill gives also the seconds: 
`mytz = new Temporal.TimeZone('Europe/Paris');`
`mytz.getOffsetStringFor('1900-02-01T12:00Z'); // '00:09:21' `

justingrant:
> I understand this PR is about letting an author specify a time zone with a simple string, as an offset to UTC, in the context of Temporal. I withdraw the concern I had during [last meeting](https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-09-07.md). Temporal.TimeZone has all functions for the use cases I had in mind.
> 
> However we should then stick to the documentation for `timeZone.getOffsetString()`:
> 
> > This method is similar to timeZone.getOffsetNanosecondsFor(), but returns the offset formatted as a string, with sign, hours, and minutes.
> 
> The present polyfill gives also the seconds: `mytz = new Temporal.TimeZone('Europe/Paris');` `mytz.getOffsetStringFor('1900-02-01T12:00Z'); // '00:09:21' `

The documentation is wrong. If a built-in IANA time zone (or a custom time zone that inherits from a built-in IANA zone) has sub-minute offsets, then `TimeZone.p.getOffsetStringFor` and `ZonedDateTime.p.offset` should return the string including sub-minute values. The limitation to minutes precision applies only to offset time zones, not to IANA zones. 

I'm working on a PR of those docs to fix the problem.

FrankYFTang:
This PR reach consensus on 2023-09-26 meeting

ryzokuken:
Is this still blocked on something or should I hit merge?

FrankYFTang:
please merge

justingrant:
Congratulations everyone for getting this over the finish line!

PR #792, opened 2023-06-12
Editorial: added explanatory note on appropriate usage of `search` collation

### LINKED ISSUE #256: What should happen with the usage: "sort" vs "search" choice in implementations?

ben-allen:
Fix #256 

(see also corresponding MDN PR: https://github.com/mdn/content/pull/27274)

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -80,7 +80,7 @@
     </emu-clause>
 
     <emu-note>
-      The collation activated by the `search` value for the `usage` option should only be used to find matching strings, since this collation is not guaranteed to be in any particular order.
+      The collation associated with the *"search"* value for the *"usage"* option should only be used to find matching strings, since this collation is not guaranteed to be in any particular order (as described in [Unicode Technical Standard #35 Part 1 Core, Unicode Collation Identifier](https://unicode.org/reports/tr35/#UnicodeCollationIdentifier) and [Unicode Technical Standard #10 Unicode Collation Algorithm, Searching and Matching](https://unicode.org/reports/tr10/#Searching)).
     </emu-note>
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):

spec/collator.html
```diff
@@ -78,6 +78,10 @@ <h1>InitializeCollator ( _collator_, _locales_, _options_ )</h1>
         1. Return _collator_.
       </emu-alg>
     </emu-clause>
+
+    <emu-note>
+      The collation activated by the `search` value for the `usage` option should only be used to find matching strings, since this collation is not guaranteed to be in any particular order.
```
gibson042, line 83:
```suggestion
      The collation associated with the *"search"* value for the *"usage"* option should only be used to find matching strings, since this collation is not guaranteed to be in any particular order (as described in [Unicode Technical Standard #35 Part 1 Core, Unicode Collation Identifier](https://unicode.org/reports/tr35/#UnicodeCollationIdentifier) and [Unicode Technical Standard #10 Unicode Collation Algorithm, Searching and Matching](https://unicode.org/reports/tr10/#Searching)).
```

gibson042 (approved):
Thanks! A few suggested tweaks, then LGTM.

PR #796, opened 2023-06-26
Editorial: Removed several unnecessary double negatives

### LINKED ISSUE #458: Editorial: If x is not undefined, then A  / Else B
There are several places in the ECMA402 spec that we have the following pattern

```
1. If X is not undefined, then
  a. AAAA
2. Else,
 b. BBBB
```

I think that double negative is just harder to read. Could we change them to
```
1. If X is undefined, then
  a. BBBB
2. Else,
 b. AAAA
```

ben-allen:
Rebased 3-1-24, and updated to include @gibson042 suggestion to improve readability of `SetNumberFormatDigitOptions` AO in `Intl.NumberFormat`
Fix #458 

Refactored several cases from  

1. If not *undefined*, then 
  a. X
2 Else,
  a. Y

with

1. If *undefined*, then 
  a. Y
2. Else,
  a. X

and 

1. If not *undefined*, then
  a. Return X.
2. Return Y.

to 
1. If *undefined*, then
  a. Return Y.
2. Return X.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -130,16 +130,7 @@
         1. Set _dateTimeFormat_.[[DateStyle]] to _dateStyle_.
         1. Let _timeStyle_ be ? GetOption(_options_, *"timeStyle"*, ~string~, &laquo; *"full"*, *"long"*, *"medium"*, *"short"* &raquo;, *undefined*).
         1. Set _dateTimeFormat_.[[TimeStyle]] to _timeStyle_.
-        1. If _dateStyle_ is not *undefined* or _timeStyle_ is not *undefined*, then
-          1. If _hasExplicitFormatComponents_ is *true*, then
-            1. Throw a *TypeError* exception.
-          1. If _required_ is ~date~ and _timeStyle_ is not *undefined*, then
-            1. Throw a *TypeError* exception.
-          1. If _required_ is ~time~ and _dateStyle_ is not *undefined*, then
-            1. Throw a *TypeError* exception.
-          1. Let _styles_ be _resolvedLocaleData_.[[styles]].[[&lt;_resolvedCalendar_&gt;]].
-          1. Let _bestFormat_ be DateTimeStyleFormat(_dateStyle_, _timeStyle_, _styles_).
-        1. Else,
+        1. If _dateStyle_ is *undefined* and _timeStyle_ is *undefined*, then
           1. Let _needDefaults_ be *true*.
           1. If _required_ is ~date~ or ~any~, then
             1. For each property name _prop_ of &laquo; *"weekday"*, *"year"*, *"month"*, *"day"* &raquo;, do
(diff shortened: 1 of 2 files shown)


ryzokuken (approved):

spec/datetimeformat.html
```diff
@@ -803,10 +803,10 @@ <h1>DateTimeStyleFormat ( _dateStyle_, _timeStyle_, _styles_ )</h1>
...
             1. Set _format_.[[rangePatterns12]] to _dateTimeRangeFormat_.[[rangePatterns12]].
           1. Return _format_.
-        1. If _timeStyle_ is not *undefined*, then
-          1. Return _timeFormat_.
-        1. Assert: _dateStyle_ is not *undefined*.
-        1. Return _dateFormat_.
+        1. If _timeStyle_ is *undefined*, then
+          1. Assert: _dateStyle_ is not *undefined*.
+          1. Return _dateFormat_.
+        1. Return _timeFormat_.
```
gibson042, line 809: I think I just disagree with #458 here. The "fallbacks" pattern more ergonomically translates to comprehensible code[^1] IMO.
```suggestion
        1. Else if _dateStyle_ is not *undefined*, then
          1. Return _dateFormat_.
        1. Else,
          1. Assert: _timeStyle_ is not *undefined*.
          1. Return _timeFormat_.
```

[^1]: e.g.,
      ```js
      if (dateStyle !== undefined && timeStyle !== undefined) {
        …
      } else if (dateStyle !== undefined) {
        return dateFormat;
      } else {
        assert(timeStyle !== undefined);
        return timeFormat;
      }
      ```

spec/datetimeformat.html
```diff
@@ -944,8 +944,8 @@ <h1>FormatDateTimePattern ( _dateTimeFormat_, _patternParts_, _x_, _rangeFormatO
...
             1. Append a new Record { [[Type]]: _p_, [[Value]]: _fv_ } as the last element of the list _result_.
           1. Else if _p_ matches a Property column of the row in <emu-xref href="#table-datetimeformat-components"></emu-xref>, then
-            1. If _rangeFormatOptions_ is not *undefined*, let _f_ be the value of _rangeFormatOptions_'s field whose name matches _p_.
-            1. Else, let _f_ be the value of _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the matching row.
+            1. If _rangeFormatOptions_ is *undefined*, let _f_ be the value of _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the matching row.
+            1. Else, let _f_ be the value of _rangeFormatOptions_'s field whose name matches _p_.
```
gibson042, line 948: Here too.
```suggestion
            1. If _rangeFormatOptions_ is not *undefined*, let _f_ be _rangeFormatOptions_.[[<_p_>]]. Otherwise, let _f_ be the value of _dateTimeFormat_'s internal slot whose name is the Internal Slot column of the matching row.
```

spec/datetimeformat.html
```diff
@@ -109,17 +109,17 @@ <h1>InitializeDateTimeFormat ( _dateTimeFormat_, _locales_, _options_ )</h1>
         1. Set _dateTimeFormat_.[[DateStyle]] to _dateStyle_.
         1. Let _timeStyle_ be ? GetOption(_options_, *"timeStyle"*, ~string~, &laquo; *"full"*, *"long"*, *"medium"*, *"short"* &raquo;, *undefined*).
         1. Set _dateTimeFormat_.[[TimeStyle]] to _timeStyle_.
-        1. If _dateStyle_ is not *undefined* or _timeStyle_ is not *undefined*, then
-          1. If _hasExplicitFormatComponents_ is *true*, then
-            1. Throw a *TypeError* exception.
-          1. Let _styles_ be _dataLocaleData_.[[styles]].[[&lt;_resolvedCalendar_&gt;]].
-          1. Let _bestFormat_ be DateTimeStyleFormat(_dateStyle_, _timeStyle_, _styles_).
-        1. Else,
+        1. If _dateStyle_ is *undefined* and _timeStyle_ is *undefined*, then
```
gibson042, line 112: I'm not sure whether or not abandoning the "_If \<special case>, then \<handle>. Else, \<fallback>._" pattern is an improvement. Maybe we just need an alias describing the special case?
```suggestion
        1. If _dateStyle_ is *undefined* and _timeStyle_ is *undefined*, let _compositeStyle_ be *false*. Else, let _compositeStyle_ be *true*.
        1. If _compositeStyle_ is *true*, then
          1. If _hasExplicitFormatComponents_ is *true*, then
            1. Throw a *TypeError* exception.
          1. Let _styles_ be _dataLocaleData_.[[styles]].[[&lt;_resolvedCalendar_&gt;]].
          1. Let _bestFormat_ be DateTimeStyleFormat(_dateStyle_, _timeStyle_, _styles_).
        1. Else,
```

gibson042 (changes requested):

sffc (commented):
Fairly neutral on this; smaller changes are better.

gibson042 (approved):

PR #735, opened 2022-12-14
Editorial: Align locale handling with UTS 35

gibson042:
Canonicalization has moved to [Annex C](https://unicode.org/reports/tr35/#LocaleId_Canonicalization), and there's not really a sharp distinction between canonicalizing _syntax_ vs. _form_, nor is there a dedicated section for canonicalizing keyword values.

TODO/followup:
* [ ] Make operations to consolidate Unicode grammar references, e.g. IsWellFormedLocale{UValue,Script,Region,LanguageId} (the last probably prompting a comprehensive rename of "language tag" to something more accurate such as "locale id"—note in UTS 35 "_A Unicode **locale identifier** is composed of a Unicode **language identifier** plus (optional) locale extensions_").
* [ ] Make an operation to split a locale identifier into { languageId, extensions, private-use } parts and an operation to split languageId into { baseLanguageCode, script, region, variants } rather than scattered incomplete parsing throughout the spec. In the same vein, consider an operation to split a string into subtags, possibly a generic one to also replace use of StringIndexOf for splitting on ["-x-"](https://tc39.es/ecma402/#sec-insert-unicode-extension-and-canonicalize) and ["-per-"](https://tc39.es/ecma402/#sec-iswellformedunitidentifier) and ["."](https://tc39.es/ecma402/#sec-partitionnotationsubpattern) ([twice](https://tc39.es/ecma402/#sec-getoperands))—maybe even in ECMA-262.
* [ ] Introduce a new `<dfn>CLDR</dfn>` that holds the sole reference to cldr.unicode.org.
* [ ] [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) (e.g., merge IsStructurallyValidLanguageTag into CanonicalizeUnicodeLocaleId)

PR #801, opened 2023-07-10
Editorial: Correct the Intl.Locale.prototype.{script,region} algorithms

### LINKED ISSUE #799: Use unicode_script_subtag and unicode_region_subtag in Locale.prototype.{script,region}
The current definition is hard to read and at least for `Intl.Locale.prototype.script` also incorrect.

[get Intl.Locale.prototype.script](https://tc39.es/ecma402/#sec-Intl.Locale.prototype.script)
> 5. Let lang be the longest prefix of locale matched by the unicode_language_id [Unicode locale nonterminal](https://tc39.es/ecma402/#sec-language-tags).
> 6. If lang contains a [subtag](https://tc39.es/ecma402/#bcp-47-subtag) of length 4, return that [subtag](https://tc39.es/ecma402/#bcp-47-subtag).

`new Intl.Locale("de-1996").script` should return the variant subtag `"1996"` according to that definition, which is clearly wrong.

gibson042:
Fixes #799
Ref #735

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -347,7 +347,7 @@
         1. Assert: _locale_ can be matched by the <code>unicode_locale_id</code> Unicode locale nonterminal.
         1. Let _lang_ be the longest prefix of _locale_ matched by the <code>unicode_language_id</code> Unicode locale nonterminal.
         1. Let _subtags_ be a List of the subtags of _lang_.
-        1. NOTE: A <code>unicode_region_subtag</code> subtag is only valid immediately after an initial <code>unicode_language_subtag</code> subtag that itself might be followed by a <code>unicode_script_subtag</code> subtag, and <code>unicode_region_subtag</code> cannot be confused with <code>unicode_variant_subtag</code> because their productions are disjoint.
+        1. NOTE: A <code>unicode_region_subtag</code> subtag is only valid immediately after an initial <code>unicode_language_subtag</code> subtag, optionally with a single <code>unicode_script_subtag</code> subtag between them. In that position, <code>unicode_region_subtag</code> cannot be confused with any other valid subtag because all their productions are disjoint.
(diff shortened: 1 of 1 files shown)


spec/locale.html
```diff
@@ -347,8 +347,9 @@ <h1>get Intl.Locale.prototype.region</h1>
         1. Assert: _locale_ can be matched by the <code>unicode_locale_id</code> Unicode locale nonterminal.
         1. Let _lang_ be the longest prefix of _locale_ matched by the <code>unicode_language_id</code> Unicode locale nonterminal.
         1. Let _subtags_ be a List of the subtags of _lang_.
-        1. If the length of _subtags_ is greater than 1 and the length of _subtags_[1] is 4, let _i_ be 2. Otherwise, let _i_ be 1.
-        1. If the length of _subtags_ is greater than _i_ and the length of _subtags_[_i_] is in the inclusive range from 2 to 3, return _subtags_[_i_].
+        1. NOTE: A <code>unicode_region_subtag</code> subtag is only valid immediately after an initial <code>unicode_language_subtag</code> subtag that itself might be followed by a <code>unicode_script_subtag</code> subtag, and <code>unicode_region_subtag</code> cannot be confused with <code>unicode_variant_subtag</code> because their productions are disjoint.
```
ben-allen, line 350: This wording feels a little unclear to me. What about:

1. NOTE: A <code>unicode_region_subtag</code> subtag is only valid immediately after an initial <code>unicode_language_subtag</code>, or immediately after a <code>unicode_script_subtag</code> that is itself immediately after an initial <code>unicode_language_subtag</code>

ben-allen (commented):
Thank you so much for doing this -- it's a vast improvement to readability

spec/locale.html
gibson042: Acknowledged, and tweaked slightly.

ben-allen (approved):
LGTM!

sffc:
@gibson042 Is this ready to merge?

gibson042:
I think so, but didn't want to do it unilaterally.

PR #803, opened 2023-07-10
Editorial: Replaced all non-standard 'increase _x_ by _y_' and 'decrease _x_ by …' statements 

### LINKED ISSUE #677: Replace "{increase,decrease} _alias_ by $value" algorithm steps
Rather than using "increase _alias_ by $value" or "decrease _alias_ by $value", we should consistently use the more standard "set _alias_ to _alias_ {+,-} $value".

ben-allen:
Replaced all non-standard 'increase _x_ by _y_' and 'decrease _x_ by …' statements with standard 'Set _x_ to _x_ + _y_' and 'set _x_ to _x_ - _y_' statements. fix #677.


gibson042 (approved):
Thanks!

PR #804, opened 2023-07-11
Editorial: Clarify operations related to merging locale data

gibson042:
* Rename ApplyOptionsToTag to UpdateLanguageId, move its locale id syntax validation out to the (sole) call site, and handle each overriding datum in its own single block.
* Rename ApplyUnicodeExtensionToTag to MergeLocaleData.
* Refactor InsertUnicodeExtensionAndCanonicalize(_locale_, _extension_) into InsertUnicodeExtensionAndCanonicalize(_locale_, _attributes_, _keywords_) (constructing the new -u- extension inline).
* Fix a spec bug in CanonicalizeUnicodeLocaleId.

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -30,7 +30,9 @@
         1. Else,
           1. Let _tag_ be ? ToString(_tag_).
         1. Set _options_ to ? CoerceOptionsToObject(_options_).
-        1. Set _tag_ to ? ApplyOptionsToTag(_tag_, _options_).
+        1. If IsStructurallyValidLanguageTag(_tag_) is *false*, throw a *RangeError* exception.
+        1. Set _tag_ to CanonicalizeUnicodeLocaleId(_tag_).
+        1. Set _tag_ to ? UpdateLanguageId(_tag_, _options_).
         1. Let _opt_ be a new Record.
         1. Let _calendar_ be ? GetOption(_options_, *"calendar"*, ~string~, ~empty~, *undefined*).
         1. If _calendar_ is not *undefined*, then
@@ -51,7 +53,7 @@
         1. If _numberingSystem_ is not *undefined*, then
           1. If _numberingSystem_ cannot be matched by the <code>type</code> Unicode locale nonterminal, throw a *RangeError* exception.
         1. Set _opt_.[[nu]] to _numberingSystem_.
-        1. Let _r_ be ApplyUnicodeExtensionToTag(_tag_, _opt_, _relevantExtensionKeys_).
+        1. Let _r_ be MergeLocaleData(_tag_, _opt_, _relevantExtensionKeys_).
         1. Set _locale_.[[Locale]] to _r_.[[locale]].
         1. Set _locale_.[[Calendar]] to _r_.[[ca]].
         1. Set _locale_.[[Collation]] to _r_.[[co]].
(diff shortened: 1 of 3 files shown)


spec/locale.html
```diff
@@ -68,33 +69,33 @@ <h1>Intl.Locale ( _tag_ [ , _options_ ] )</h1>
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-apply-options-to-tag" aoid=ApplyOptionsToTag>
-      <h1>ApplyOptionsToTag ( _tag_, _options_ )</h1>
-
+    <emu-clause id="sec-updatelanguageid" type="abstract operation" oldids="sec-apply-options-to-tag">
+      <h1>
+        UpdateLanguageId (
```
ben-allen, line 74: I'm a big fan of getting LanguageId into the name of this AO, given that the `unicode_language_id` is what it works with

spec/locale.html
```diff
@@ -51,7 +52,7 @@ <h1>Intl.Locale ( _tag_ [ , _options_ ] )</h1>
         1. If _numberingSystem_ is not *undefined*, then
           1. If _numberingSystem_ cannot be matched by the <code>type</code> Unicode locale nonterminal, throw a *RangeError* exception.
         1. Set _opt_.[[nu]] to _numberingSystem_.
-        1. Let _r_ be ! ApplyUnicodeExtensionToTag(_tag_, _opt_, _relevantExtensionKeys_).
+        1. Let _r_ be MergeLocaleData(_tag_, _opt_, _relevantExtensionKeys_).
```
ben-allen, line 55: I feel like I don't _want_ to like this name, but that I nevertheless like it because it's such a big improvement over the old name and because I can't think of anything that seems better.

spec/negotiation.html
```diff
@@ -173,16 +173,24 @@ <h1>
     <emu-clause id="sec-insert-unicode-extension-and-canonicalize" type="abstract operation">
       <h1>
         InsertUnicodeExtensionAndCanonicalize (
-          _locale_: a Unicode canonicalized locale identifier,
-          _extension_: a Unicode locale extension sequence,
+          _locale_: a language tag,
+          _attributes_: a List of Strings,
```
ben-allen, line 177: The thing I'm wondering about is what this gets us over the old version -- is it just so that InsertUnicodeExtensionAndCanonicalize can do in a more explicit way what this line from the old ApplyUnicodeExtensionsToLocale did?

```8. Let newExtension be a Unicode BCP 47 U Extension based on attributes and keywords.```

If all the changes in ResolveLocale were just made to account for the change in the parameters to InsertUnicodeExtensionAndCanonicalize I feel like all else being equal leaving InsertUnicodeExtensionAndCanonicalize unchanged is better.

File: spec/locale.html [hunk 4]
```diff
@@ -103,18 +104,25 @@ <h1>ApplyOptionsToTag ( _tag_, _options_ )</h1>
               * _region_
               * _variants_
           1. Else,
-            1. Set _languageId_ to _languageId_ with the <emu-not-ref>substring</emu-not-ref> matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal replaced with the string _region_.
-        1. Set _tag_ to _tag_ with the <emu-not-ref>substring</emu-not-ref> matched by the <code>unicode_language_id</code> Unicode locale nonterminal replaced by the string _languageId_.
-        1. Return CanonicalizeUnicodeLocaleId(_tag_).
+            1. Set _languageId_ to _languageId_ with the <emu-not-ref>substring</emu-not-ref> matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal replaced with _region_.
+        1. Set _tag_ to _tag_ with the substring matched by the <code>unicode_language_id</code> Unicode locale nonterminal replaced with _languageId_.
+        1. Return _tag_.
```
ben-allen, line 109: Making sure I'm getting this one right: it's not necessary to canonicalize here, because the new version of MergeLocaleData is guaranteed to canonicalize it anyway.

ben-allen (commented):
This is excellent -- I've got some questions about the refactored InsertUnicodeExtensionAndCanonicalize, but that may be because I'm missing what's going on there.

spec/locale.html
```diff
@@ -103,18 +104,25 @@ <h1>ApplyOptionsToTag ( _tag_, _options_ )</h1>
...
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-apply-unicode-extension-to-tag" aoid=ApplyUnicodeExtensionToTag>
-      <h1>ApplyUnicodeExtensionToTag ( _tag_, _options_, _relevantExtensionKeys_ )</h1>
-
+    <emu-clause id="sec-mergelocaledata" type="abstract operation" oldids="sec-apply-unicode-extension-to-tag">
+      <h1>
+        MergeLocaleData (
```
anba, line 115: [LocaleData](https://tc39.es/ecma402/#sec-internal-slots) is already a term used in the spec, reusing it here can cause confusion.

anba (changes requested):

spec/locale.html
gibson042: Correct.

spec/negotiation.html
gibson042: > The thing I'm wondering about is what this gets us over the old version -- is it just so that InsertUnicodeExtensionAndCanonicalize can do in a more explicit way what this line from the old ApplyUnicodeExtensionsToLocale did?
> 
> `8. Let newExtension be a Unicode BCP 47 U Extension based on attributes and keywords.`

That is a goal, but so is isolating construction of Unicode locale extension sequences to one place—and both calling algorithms already have a list of attributes (trivially in the case of ResolveLocale) and handle keywords one by one. I also foresee similar convergence between CanonicalizeUnicodeLocaleId and InsertUnicodeExtensionAndCanonicalize, although I'm not yet taking that step here.

> If all the changes in ResolveLocale were just made to account for the change in the parameters to InsertUnicodeExtensionAndCanonicalize I feel like all else being equal leaving InsertUnicodeExtensionAndCanonicalize unchanged is better.

No, I do also find that replacing iterative string concatenation in ResolveLocale with use of keyword Records increases its comprehensibility. For example, compare:
```diff
-1. Let _supportedExtensionAddition_ be the string-concatenation of *"-"*, _key_, *"-"*, and _value_.
+1. Set _supportedKeyword_ to the Record { [[Key]]: _key_, [[Value]]: _value_ }.
```

spec/locale.html
```diff
@@ -68,33 +69,33 @@ <h1>Intl.Locale ( _tag_ [ , _options_ ] )</h1>
...
         1. If _language_ is not *undefined*, then
           1. If _language_ cannot be matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
+          1. Set _languageId_ to _languageId_ with the <emu-not-ref>substring</emu-not-ref> matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal replaced with _language_.
         1. Let _script_ be ? GetOption(_options_, *"script"*, ~string~, ~empty~, *undefined*).
         1. If _script_ is not *undefined*, then
           1. If _script_ cannot be matched by the <code>unicode_script_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _region_ be ? GetOption(_options_, *"region"*, ~string~, ~empty~, *undefined*).
-        1. If _region_ is not *undefined*, then
-          1. If _region_ cannot be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Set _tag_ to CanonicalizeUnicodeLocaleId(_tag_).
```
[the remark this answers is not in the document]
gibson042: Nice catch, and thanks for the clear example! I have restored the early canonicalization, and placed it at the call site in anticipation of the oft-repeated `IsStructurallyValidLanguageTag` check merging into `CanonicalizeUnicodeLocaleId`.

spec/locale.html
gibson042: Good point. Renamed to MakeLocaleRecord.

sffc:
For possible inspiration: https://github.com/unicode-org/icu4x/pull/1833

sffc:
@gibson042 Is this ready for @anba to re-review?

gibson042:
Yes.

sffc:
I suggest @ben-allen to verify that @gibson042's latest changes address @anba's comments.

ryzokuken (approved):
LGTM overall, confirmed that @anba's two comments were appropriately addressed, marking them as resolved.

ryzokuken:
@gibson042 please rebase this whenever you could but this looks ready to merge.

PR #807, opened 2023-07-19
Normative: BestAvailableLocale AO now removes extensions before setting initial value of candidate locale.

### LINKED ISSUE #685: `BestAvailableLocale` operation requires additional clarifications/changes to handle extensions
Hello! I'm part of the development team of [boa](https://github.com/boa-dev/boa), where we are developing an ECMAScript engine with `Intl` support, and we have a bit of a problem with how the [`BestAvailableLocale`](https://tc39.es/ecma402/#sec-bestavailablelocale) operation is described.

For context, I had a discussion with some of the [`icu4x`](https://github.com/unicode-org/icu4x) folks (https://github.com/boa-dev/boa/pull/2072#issuecomment-1141625219) and we discovered that the `BestAvailableLocale` operation, as currently described in the latest revision, tries to match invalid Unicode BCP 47 locale identifiers when executing the algorithm described.

https://github.com/boa-dev/boa/pull/2072#issuecomment-1141836914
> I don't think Language Identifiers are the problem per se; I can easily manipulate LanguageIdentifiers using the provided API, and a canonicalized Language Identifier cannot become invalid by following the spec algorithm. The problem are extensions, which could become invalid if that algorithm is followed as stated e.g. with `hi-t-en-h0-hybrid` it would try to match with `hi-t-en-h0`, which has an invalid t extension since the h0 key ought to have a value.

The team kindly recommended us to treat the input `locale` as a simple Language Identifier instead, which should work but doesn't match the algorithm described. Knowing that, the question about extensions still remains, and I would like to open this issue as a request to add additional clarifications on `BestAvailableLocale`. Some of our questions are:

- Should we treat `locale` as a plain Language Identifier?
- Should we extract the Language Identifier part from `locale` and try to match with that, then append the removed extensions to the matched locale?
- Is there another way to describe this algorithm that doesn't generate invalid locales and doesn't treat them as simple string literals?

Thanks!

cc @zbraniecki @sffc

ben-allen:
BestAvailableLocale AO now removes extensions before setting initial value of `candidate` locale. Ensures that attempts to match invalid locales (i.e. ending in extension keys without associated types)  don't occur. fix #685.

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -83,9 +83,8 @@
             1. Let _requestedLocale_ be _requestedLocales_[0].
           1. Else,
             1. Let _requestedLocale_ be ! DefaultLocale().
-          1. Let _noExtensionsLocale_ be the String value that is _requestedLocale_ with any Unicode locale extension sequences removed.
           1. Let _availableLocales_ be a List with language tags that includes the languages for which the Unicode Character Database contains language sensitive case mappings. Implementations may add additional language tags if they support case mapping for additional locales.
-          1. Let _locale_ be BestAvailableLocale(_availableLocales_, _noExtensionsLocale_).
+          1. Let _locale_ be BestAvailableLocale(_availableLocales_, _requestedLocale_).
           1. If _locale_ is *undefined*, set _locale_ to *"und"*.
           1. Let _codePoints_ be StringToCodePoints(_S_).
           1. If _targetCase_ is ~lower~, then
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -88,10 +88,15 @@
       </dl>
       <emu-alg>
         1. Let _candidate_ be the String value that is _locale_ with any Unicode locale extension sequences removed.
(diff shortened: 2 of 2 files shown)


anba:
This change means callers to `BestAvailableLocale` no longer have to remove Unicode extension sequences. This allows to simplify `LookupSupportedLocales` and `TransformCase` a bit.

It also means locales with extensions can no longer be resolved, for example private-use extensions. I think it'd make sense to update [Internal slots of Service Constructors](https://tc39.es/ecma402/#sec-internal-slots) to restrict `[[AvailableLocales]]` to a list of language tags matching exactly `unicode_language_id`, i.e. no extensions are allowed.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-08-03.md#normative-bestavailablelocale-ao-now-removes-extensions-before-setting-initial-value-of-candidate-locale-807

Conclusion: The spec follows RFC 4647 already. We should continue using that algorithm. Improve docs to be clear that it is not possible to return a non-canonical locale ID, even though they could exist transiently as part of the AO.

ben-allen:
moved to #824

PR #823, opened 2023-08-09
Editorial: Document behaviour of U extension values locale slots

### LINKED ISSUE #311: strip out u-extension in the locale returned by resolvedOptions()

ben-allen:
Fix #311 by adding a non-normative note explaining the behaviour of the [[locale]] slot in ResolveLocale AO with regard to what U Extension tags appear or don't appear in the string stored in this slot.

Replaces #817


spec/negotiation.html
```diff
@@ -260,7 +260,7 @@ <h1>ResolveLocale ( _availableLocales_, _requestedLocales_, _options_, _relevant
...
+        Non-normative summary: Two algorithms are available to match the locales: the Lookup algorithm described in RFC 4647 section 3.4, and an implementation dependent best-fit algorithm. Independent of the locale matching algorithm, options specified through Unicode locale extension sequences are negotiated separately, taking the caller's relevant extension keys and locale data into consideration as well as the client-provided options. The abstract operation returns a Record with a field for each key in _relevantExtensionKeys_, containing the selected values for those keys. This Record also contains a [[locale]] field which stores the language tag, but with whatever U Extensions tags that are not in _relevantExtensionKeys_ removed. If a property is set by an option and also specified in the locale extension sequence, the value from the option is used to set the property's value and the overridden extension tag is removed from the string stored in [[locale]].
```
gibson042, line 263:
```suggestion
        Non-normative summary: Two algorithms are available to match the locales: the Lookup algorithm described in RFC 4647 section 3.4, and an implementation dependent best-fit algorithm. Independent of the locale matching algorithm, options specified through Unicode locale extension sequences are negotiated separately, taking the caller's relevant extension keys and locale data into consideration as well as the client-provided options. The abstract operation returns a Record with a field for each key in _relevantExtensionKeys_, containing the selected values for those keys. This Record also contains a [[locale]] field which stores the language tag of the selected locale with Unicode locale extension sequence <code>key</code> subtags restricted to those in _relevantExtensionKeys_. If an extension is specified in both the Unicode locale extension sequence and in an option, the value from the option is used to set the value of the corresponding field and the overridden extension keyword (the <code>key</code> subtag along with any immediately following <code>type</code> subtags and associated separators) is removed from the string stored in [[locale]].
```

gibson042 (approved):

sffc (commented):
Also please note that we only remove the extension tag if it is _different_ than the options bag.

gibson042 (approved):

PR #824, opened 2023-08-09
Editorial: Best available locale documentation

### LINKED ISSUE #685: `BestAvailableLocale` operation requires additional clarifications/changes to handle extensions

ben-allen:
Explanatory note added to address concerns raised in #685 

fix #685

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -94,6 +94,10 @@
           1. If _pos_ &ge; 2 and the character *"-"* occurs at index _pos_ - 2 of candidate, decrease _pos_ by 2.
           1. Let _candidate_ be the substring of _candidate_ from position 0, inclusive, to position _pos_, exclusive.
       </emu-alg>
+
+      <emu-note>
+        In cases where Unicode T Extensions are included in _locale_, this algorithm may temporarily generate invalid language tags. However, it is not possible for the abstract operation to return non-canonical or otherwise invalid language tags.
+      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-lookupmatcher" aoid="LookupMatcher">
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -96,7 +96,7 @@
       </emu-alg>
 
       <emu-note>
-        In cases where Unicode T Extensions are included in _locale_, this algorithm may temporarily generate invalid language tags. However, it is not possible for the abstract operation to return non-canonical or otherwise invalid language tags.
+        In cases where Unicode T Extensions are included in _locale_, this algorithm may temporarily generate invalid language tags. However, it is not possible in any case for the abstract operation to return invalid language tags.
       </emu-note>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec/negotiation.html
```diff
@@ -94,6 +94,10 @@ <h1>
           1. If _pos_ &ge; 2 and the character *"-"* occurs at index _pos_ - 2 of candidate, decrease _pos_ by 2.
           1. Let _candidate_ be the substring of _candidate_ from position 0, inclusive, to position _pos_, exclusive.
       </emu-alg>
+
+      <emu-note>
+        In cases where Unicode T Extensions are included in _locale_, this algorithm may temporarily generate invalid language tags. However, it is not possible in any case for the abstract operation to return invalid language tags.
```
gibson042, line 99:
```suggestion
        When _locale_ includes a <a href="https://unicode.org/reports/tr35/#BCP47_T_Extension">Unicode Technical Standard #35 Part 1 Core BCP 47 T Extension</a> subtag sequence, the truncation in this algorithm may temporarily generate invalid language tags. However, none of them will be returned because _availableLocales_ contains only valid language tags.
```

sffc (approved):

sffc:
CC @jedel1043

jedel1043:
I think this works too. This clarification would allow us to directly skip generating the invalid locales and remove the extensions one by one instead of having to convert between `ICU4X`'s `Locale` and its `String` representation.

Issue #578, opened 2021-06-04
Define sort order of "pluralCategories" array


anba:
[Intl.PluralRules.prototype.resolvedOptions ( )](https://tc39.es/ecma402/#sec-intl.pluralrules.prototype.resolvedoptions) should define the sort order of the "pluralCategories" array, e.g. `["one", "zero"]` or `["zero", "one"]`. ICU (and therefore all browsers) return the array in alphabetical order, i.e. `["few", "many" "one", "other", "two", "zero"]`.

sffc:
This should be a small normative PR, since it reflects web reality.

zbraniecki:
My preference if I were to ignore web reality would be to order them the way UTS 35 does: 

```xml
<!ATTLIST pluralRule count (zero | one | two | few | many | other) #REQUIRED >
```

ryzokuken:
Can we do that? I really like that idea, if we can choose to ignore web reality...

sffc:
The set of six plural forms could grow in the future (unlikely but possible).

I think ICU prefers sorting alphabetically because it's predictable and algorithmic. If you have a custom sort order, then you need to have a special code path.

My preference is to reflect web reality instead of introducing a special case sort order.

zbraniecki:
Alphabetic is a developer paper cut - it makes API clearly return "one" before "zero" while semantically the incremental order relation to numerical values is reverse.

I think if in the future we'll add new categories.

The ICU4C behavior seems like optimal for API implementation author, not API user, which I believe is the wrong API design model.

FrankYFTang:
1) I am worry that changing the order from the current web reality may break pre-existing code 
2) if the result of a fix order is not useful for the developer to depend on, then mandating a specific order will force the implemetnation to run unnecessary cpu power for something useless and slow down the implementation unnecessary.

ljharb:
Why wouldn't it be useful to depend on? Deterministic ordering is always useful.

jswalden:
The main reason to deterministically order, is to make it impossible for sites to inadvertently depend on implementation-specific behavior.  No more and no less.

I doubt any existing code depends on the ordering, myself.  At worst it's not more than extremely few (one or two) users.

Aesthetically I like the zero/one/two/etc. ordering.  But this is for programmatic consumption.  It's not unreasonable to say that users (e.g. UIs) who need the aesthetic ordering can postprocess accordingly.  And implementations (save the polyfill) ship alphabetical already.

[...]

PR #831, opened 2023-08-31
Meta: Update style guide to discuss preferred orderings of elements

ben-allen:
Toward #578

1795b9d9db31c28c869da6d859889b2f39b4c9c6 inspired by https://github.com/tc39/ecma402/issues/578 
29e183bd51c30fc5082b846ec8abcf42d7492caa inspired by https://github.com/tc39/proposal-intl-duration-format/issues/163

### DIFF
--- a/docs/style-guide.md
+++ b/docs/style-guide.md
@@ -4,6 +4,7 @@ This document is an evolving list of recommendations for the ECMA-402 specificat
 
 This document contains background on how the style decisions were reached.  The actual recommendations are highlighted with a :star2: emoji.
 
+
 ## Table of Contents
 
 - [Casing Conventions](#casing-conventions)
@@ -27,7 +28,9 @@ This document contains background on how the style decisions were reached.  The 
             - [Pros](#pros-3)
             - [Cons](#cons-3)
             - [Decision](#decision-3)
-
+    - [Element Ordering](#element-ordering)
+	    - [General Guidelines](#general-guidelines)
+	    - [resolvedOptions](#resolvedoptions)
 
 *Table of Contents generated using https://magnetikonline.github.io/markdown-toc-generate/*
 
@@ -199,3 +202,22 @@ We could enforce a camel case convention on these strings, such as the following
 ##### Decision
 
 - Defer the decision on the syntax for identifiers to the other specification when possible.
+
+## Element Ordering
+
+This section concerns the observable order for elements of arrays and properties of objects.
+
+### General Guidelines
+
+:star2: *ECMA-402 must provide a deterministic order for the elements of all arrays and the properties of all objects. This order should be lexicographic except in cases for which there is a clearly preferable semantic ordering. :star2:*
+
+For example, an array holding the time zone identifiers "Asia/Tokyo", "Asia/Ho_Chi_Minh", and "Asia/Dubai" should use the order `["Asia/Dubai", "Asia/Ho_Chi_Minh", "Asia/Tokyo"]`
+
+For an example of when *not* to use lexicographic order, consider an array holding calendar time scale units. This array could be ordered by descending magnitude as `["years", "months", "weeks", "days"]` or its reverse ascending magnitude as `["days", "weeks", "months", "years"]`, but should not use the lexicographic `["days", "months", "weeks", "years"]` order.
+
+### `resolvedOptions`
+:star2: *Properties in the output from `resolvedOptions()` of an Intl object should start with `locale` and otherwise use lexicographic order. Deviations (such as those motivated by semantic order involving a subset of properties) must be documented.* :star2:
+
+Spec changes that add properties should do so in accordance with this recommendation, rather than automatically placing them at the end—relative enumeration order of properties should remain stable over time, but there is no such expectation regarding adjacency.
+#### Examples
+`Object.keys(new Intl.Segmenter().resolvedOptions())` returns `[ 'locale', 'granularity' ]`. If support for the [Unicode BCP 47 U Extension "dx" key](https://unicode.org/reports/tr35/#UnicodeDictionaryBreakExclusionIdentifier) were added and exposed as `dictionaryBreakExcludedScripts`, that property would belong before `granularity`.


docs/style-guide.md
```diff
@@ -199,3 +201,22 @@ We could enforce a camel case convention on these strings, such as the following
...
+
+## Element Ordering
+
+This section concerns the order in which containers store elements. 
+
+### General Guidelines
+
+:star2: ECMA-402 spec must provide a deterministic order for the contents of all containers. This order should be lexicographic except in cases wherein there is a clearly preferable semantic ordering. :star2:
+  
+For an example of when *not* to use lexicographic order, consider an array holding days of the week. This array should store those days in the order `["monday", "tuesday", "wednesday", ...]` instead of the lexicographic `["friday", "monday", "saturday", "sunday", ...]` order.
```
sffc, line 213: Choose something that is actually in the spec:

```suggestion
For an example of when *not* to use lexicographic order, consider an array holding days of the week. This array should store those days in the order `["years", "months", "weeks", "days"]` instead of the lexicographic `["days", "months", "weeks", "years"]` order.
```

docs/style-guide.md
```diff
@@ -199,3 +201,22 @@ We could enforce a camel case convention on these strings, such as the following
...
+
+### `resolvedOptions`
+
+:star2: The `resolvedOptions` of Intl objects should appear in the following order: 
+
+1. `locale`
+2. All properties (given in lexicographical order) that can be set by extension keys and that are guaranteed to exist
+3. Properties (in lexicographical order) that are not set by extension keys and that are guaranteed to exist 
+4. All properties (in lexicographical order) that exist conditionally. :star2:
```
sffc, line 222: @gibson042 for feedback on this ordering

docs/style-guide.md
gibson042:
```suggestion
For an example of when *not* to use lexicographic order, consider an array holding calendar time scale units. This array could use descending magnitude as `["years", "months", "weeks", "days"]` or its reverse ascending magnitude, but should not use the lexicographic `["days", "months", "weeks", "years"]` order.
```

Please also provide at least one positive example.

docs/style-guide.md
```diff
@@ -199,3 +201,22 @@ We could enforce a camel case convention on these strings, such as the following
 ##### Decision
 
 - Defer the decision on the syntax for identifiers to the other specification when possible.
+
+## Element Ordering
+
+This section concerns the order in which containers store elements.
```
gibson042, line 207: What is the definition of "container" in this context?

docs/style-guide.md
```diff
@@ -199,3 +201,22 @@ We could enforce a camel case convention on these strings, such as the following
...
 
 - Defer the decision on the syntax for identifiers to the other specification when possible.
+
+## Element Ordering
+
+This section concerns the order in which containers store elements. 
+
+### General Guidelines
+
+:star2: ECMA-402 spec must provide a deterministic order for the contents of all containers. This order should be lexicographic except in cases wherein there is a clearly preferable semantic ordering. :star2:
```
gibson042, line 211:
```suggestion
:star2: *ECMA-402 spec must provide a deterministic order for the contents of all containers. This order should be lexicographic except in cases wherein there is a clearly preferable semantic ordering.* :star2:
```

docs/style-guide.md
gibson042: It seems insufficiently stable to me, because we are currently contemplating adding support for -u- extension keys that can currently only be set by options, and could definitely promote a conditional property to mandatory in the future.

```suggestion
### `resolvedOptions`

:star2: *Properties in the output from `resolvedOptions()` of an Intl object should start with `locale` and otherwise use lexicographic order. Deviations (such as those motivated by semantic order involving a subset of properties) must be documented.* :star2:

Spec changes that add properties should do so in accordance with this recommendation, rather than automatically placing them at the end—relative enumeration order of properties should remain stable over time, but there is no such expectation regarding adjacency.

#### Examples

`Object.keys(new Intl.Segmenter().resolvedOptions())` returns `[ 'locale', 'granularity' ]`. If support for the [Unicode BCP 47 U Extension "dx" key](https://unicode.org/reports/tr35/#UnicodeDictionaryBreakExclusionIdentifier) were added and exposed as `dictionaryBreakExcludedScripts`, that property would belong before `granularity`.
```

gibson042 (changes requested):

docs/style-guide.md
```diff
@@ -7,27 +7,29 @@ This document contains background on how the style decisions were reached.  The
...
+		- [Alternative: Use kebab case but also accept camel case](#alternative-use-kebab-case-but-also-accept-camel-case)
+			- [Pros](#pros-2)
+			- [Cons](#cons-2)
+			- [Decision](#decision-2)
+	- [Identifiers defined outside ECMA-402](#identifiers-defined-outside-ecma-402)
+		- [Examples](#examples-1)
+		- [Alternative: Convert identifiers to camel case](#alternative-convert-identifiers-to-camel-case)
+			- [Pros](#pros-3)
+			- [Cons](#cons-3)
+			- [Decision](#decision-3)
```
ryzokuken, line 29: Are the whitespace changes necessary?

ben-allen: changed it back to the previous spacing

ryzokuken: The previous spacing was 4 spaces and your new version has 2 spaces :see_no_evil:

ryzokuken (approved):
Some old comments and some new ones, let's clean this one up!

docs/style-guide.md
gibson042: Consistency with existing conventions at [Identifiers defined by ECMA-402](https://github.com/tc39/ecma402/blob/ff6c9c68d629d1cdc3edf45755262e15d17e84b1/docs/style-guide.md#identifiers-defined-by-ecma-402) and [Identifiers defined outside ECMA-402](https://github.com/tc39/ecma402/blob/ff6c9c68d629d1cdc3edf45755262e15d17e84b1/docs/style-guide.md#identifiers-defined-outside-ecma-402).

docs/style-guide.md
```diff
@@ -199,3 +201,22 @@ We could enforce a camel case convention on these strings, such as the following
 ##### Decision
 
 - Defer the decision on the syntax for identifiers to the other specification when possible.
+
```
sffc, line 204: @ben-allen Please make sure this fully fixes #578

sffc: Actually we should follow this up with a normative PR for PluralRules.

docs/style-guide.md
```diff
@@ -199,3 +202,24 @@ We could enforce a camel case convention on these strings, such as the following
...
+For an example of when *not* to use lexicographic order, consider an array holding calendar time scale units. This array could be ordered by descending magnitude as `["years", "months", "weeks", "days"]` or its reverse ascending magnitude as `["days", "weeks", "months", "years"]`, but should not use the lexicographic `["days", "months", "weeks", "years"]` order.
+
+### `resolvedOptions`
+
+:star2: <em>The `resolvedOptions` of Intl objects should appear in the following order: 
+
+1. `locale`
+2. All properties (given in lexicographical order) that can be set by extension keys and that are guaranteed to exist
+3. Properties (in lexicographical order) that are not set by extension keys and that are guaranteed to exist 
+4. All properties (in lexicographical order) that exist conditionally.</em> :star2:
```
ryzokuken, line 225: Is this whole section purposefully italicized?

ryzokuken (approved):
Overall LGTM, thanks!

ben-allen:
@gibson042 what do you think about this version?

docs/style-guide.md
ben-allen: Yes, to match the italicization in the other starred blocks. I do admit that italicizing the list might be a situation where the result of insisting on consistency is ugly enough that it's the wrong thing to do.

ryzokuken: I think one way to do it would be to rephrase the first sentence into a summarized guideline, star+italicize it and follow that with the full list. An entire list being starred/italicized doesn't really say much and looks weird.

docs/style-guide.md
```diff
@@ -27,7 +28,9 @@ This document contains background on how the style decisions were reached.  The
             - [Pros](#pros-3)
             - [Cons](#cons-3)
             - [Decision](#decision-3)
-
+    - [Element Ordering](#element-ordering)
+	      - [General Guidelines](#general-guidelines)
+	      - [resolvedOptions](#resolvedoptions)
```
gibson042, line 33: The extra indentation here breaks Markdown list output: https://github.com/tc39/ecma402/blob/97ba211cddb8a4a22f67e327af05372b545f2bcd/docs/style-guide.md#table-of-contents:~:text=Element%20Ordering
```suggestion
    - [Element Ordering](#element-ordering)
	    - [General Guidelines](#general-guidelines)
	    - [resolvedOptions](#resolvedoptions)
```

docs/style-guide.md
```diff
@@ -199,3 +202,25 @@ We could enforce a camel case convention on these strings, such as the following
...
 
 - Defer the decision on the syntax for identifiers to the other specification when possible.
+
+## Element Ordering
+
+This section concerns the order in which containers store elements. 
+
+### General Guidelines
+
+:star2: *ECMA-402 spec must provide a deterministic order for the contents of all containers. This order should be lexicographic except in cases wherein there is a clearly preferable semantic ordering. :star2:*
```
gibson042, line 212:
```suggestion
## Element Ordering

This section concerns the observable order for elements of arrays and properties of objects.

### General Guidelines

:star2: *ECMA-402 must provide a deterministic order for the elements of all arrays and the properties of all objects. This order should be lexicographic except in cases for which there is a clearly preferable semantic ordering. :star2:*
```

docs/style-guide.md
```diff
@@ -199,3 +202,25 @@ We could enforce a camel case convention on these strings, such as the following
...
+
+For example, an array holding the time zone identifiers `Asia/Tokyo`, `Asia/Ho_Chi_Minh`, and `Asia/Dubai` should use the order `["Asia/Dubai", "Asia/Ho_Chi_Minh", "Asia/Tokyo"]`
```
gibson042, line 214:
```suggestion
For example, an array holding the time zone identifiers "Asia/Tokyo", "Asia/Ho_Chi_Minh", and "Asia/Dubai" should use the order `["Asia/Dubai", "Asia/Ho_Chi_Minh", "Asia/Tokyo"]`
```

docs/style-guide.md
```diff
@@ -199,3 +202,25 @@ We could enforce a camel case convention on these strings, such as the following
...
+
+### `resolvedOptions`
+
+:star2: *The `resolvedOptions` of `Intl` objects should appear with `locale` first, followed by properties guaranteed to exist that can be set via extension keys, followed by properties guaranteed to exist that are not set by extension keys, and then finally all properties that exist conditionally. Elements in each of these categories should appear in lexicographical order.* :star2:
+
+Order of `resolvedOptions`:
+1. `locale`
+2. All properties (given in lexicographical order) that can be set by extension keys and that are guaranteed to exist
+3. Properties (in lexicographical order) that are not set by extension keys and that are guaranteed to exist 
+4. All properties (in lexicographical order) that exist conditionally.
```
gibson042, line 226: Reiterating https://github.com/tc39/ecma402/pull/831#discussion_r1330814261 :
```suggestion
### `resolvedOptions`
:star2: *Properties in the output from `resolvedOptions()` of an Intl object should start with `locale` and otherwise use lexicographic order. Deviations (such as those motivated by semantic order involving a subset of properties) must be documented.* :star2:
Spec changes that add properties should do so in accordance with this recommendation, rather than automatically placing them at the end—relative enumeration order of properties should remain stable over time, but there is no such expectation regarding adjacency.
#### Examples
`Object.keys(new Intl.Segmenter().resolvedOptions())` returns `[ 'locale', 'granularity' ]`. If support for the [Unicode BCP 47 U Extension "dx" key](https://unicode.org/reports/tr35/#UnicodeDictionaryBreakExclusionIdentifier) were added and exposed as `dictionaryBreakExcludedScripts`, that property would belong before `granularity`.
```

gibson042 (changes requested):

docs/style-guide.md
ben-allen: I'm revising this one again and though the reasoning in [this comment](https://github.com/tc39/ecma402/pull/831#discussion_r1330814261) seems sound to me, I wanted to verify that we've got agreement that the "put conditionally existing properties after guaranteed to exist properties" language is what we want to go with. @gibson042 @ryzokuken @sffc

sffc: No strong opinion between the two orderings. I guess I slightly favor putting the extension keywords adjacent in the list to the locale and in such a way that they could be sliced out and fed to the Locale constructor, but that is an error-prone operation, so Richard's ordering is fine, too.

ben-allen:
@gibson042 Does this look good to merge?

gibson042 (approved):

PR #840, opened 2023-11-14
Editorial: Refactor PartitionPattern for simplicity and comprehensibility

gibson042:
* Adds a prose description explaining treatment of brace-wrapped placeholders and adjacent literal text in converting a String pattern into an equivalent List of Records.
* Renames spec aliases (_beginIndex_ → _placeholderStart_, _endIndex_ → _placeholderEnd_, _p_ → _placeholderName_).
* Adopts ECMA-262 phrasing conventions (e.g., "the [substring](https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#substring) of _S_ from _inclusiveStart_ [to _exclusiveEnd_]" and `<`/`≤`/`≥`/`>` operators).
* Reduces unnecessary repetition (e.g., "If _beginIndex_ \> _nextIndex_, then Let _literal_ be the substring of _pattern_ from _nextIndex_ to _beginIndex_ and Append the Record { [[Type]]: **"literal"**, [[Value]]: _literal_ } to _result_." → "Let _literal_ be the substring of _pattern_ from _placeholderEnd_ + 1 to _placeholderStart_ and If _literal_ is not the empty String, then Append the Record { [[Type]]: **"literal"**, [[Value]]: _literal_ } to _result_.").
* Makes more obvious the omission of zero-length literal Records.
* Increases alias reference locality (e.g., checking for a literal part between a previous _placeholderEnd_ `}` and current _placeholderStart_ `{` before advancing _placeholderEnd_ for the current placeholder part).

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -435,29 +435,26 @@
       <dl class="header">
         <dt>description</dt>
         <dd>
-          The [[Value]] field will be a String value if [[Type]] is *"literal"*, and *undefined* otherwise.
+          It interprets _pattern_ as containing an arbitrary number of brace-wrapped "placeholder" <emu-not-ref>substrings</emu-not-ref> and returns a List in which each such placeholder is represented as a Record with [[Type]] equal to the unwrapped <emu-not-ref>substring</emu-not-ref> and [[Value]] equal to *undefined*, and each non-empty <emu-not-ref>substring</emu-not-ref> adjacent to or between such placeholders is represented as a Record with [[Type]] equal to *"literal"* and [[Value]] equal to the <emu-not-ref>substring</emu-not-ref>, preserving the relative order of appearance such that _pattern_ could be recovered by concatenating the value from the appropriate field of each Record in turn.
           The syntax of the abstract pattern strings is an implementation detail and is not exposed to users of ECMA-402.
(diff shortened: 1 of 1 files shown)


gibson042:
@ryzokuken ping

ben-allen (approved):
This looks great! the spec steps are much clearer now, and the new description is something I wish had been in 402 when I started reading it.

A couple of ideas that I'm not sure are good or not: 

1. I realized while reviewing this that the word "placeholder" isn't actually used that often in 402, and thought maybe it would be worthwhile to provide a definition. I searched around a little for a link to a definition (thinking I'd find it in documentation for CLDR or ICU) but without much success. _Would_ it be useful to define the term, or can we assume that anyone who's reading it is either already familiar or can easily figure it out from context?
2. The description refers to placeholders as placeholders in the first clause, and then later as substrings, i.e. brace-wrapped placeholders become represented as unwrapped substrings. To my eye the switch between one word and the other carries an implication that the placeholder and the substring that's yielded when you remove the braces are different things. Would it be clearer to start the sentence with "It represents each brace-wrapped placeholder substring in `pattern`" instead of just "It represents each brace-wrapped placeholder in `pattern`"?

PR #843, opened 2023-11-21
Editorial: Prefix names of intrinsics

### LINKED ISSUE #842: Should intrinsics be scoped?
TL;DR: Is there anything speaking for, or against, renaming the intrinsics established in ECMA-402 to follow the form `%Intl.Foo%`? 

This may not be hugely important within the spec document itself, but personally I would appreciate if any references to those intrinsics in engine code can be more easily identified as `Intl`-related.

This may need input from the ECMA-262 editors - the closest relevant thing I could find is https://github.com/tc39/ecma262/commit/8f67a13e1af599fb31047c6f8e35acd7586d1bd1, which possibly sets a precedent for the names of intrinsics to match what they are referenced as from the outside world. The current Temporal draft spec uses `%Temporal.Foo%` for each constructor, opposed to a plain `%Foo%` name.

linusg:
Closes #842.

Not sure if the same naming scheme is appropriate for `%SegmentIteratorPrototype%` and `%SegmentsPrototype%`, but happy update those as well if wanted.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -4,7 +4,7 @@
   <emu-clause id="sec-the-intl-collator-constructor">
     <h1>The Intl.Collator Constructor</h1>
     <p>
-      The Intl.Collator constructor is the <dfn>%Collator%</dfn> intrinsic object and a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in <emu-xref href="#sec-internal-slots"></emu-xref>.
+      The Intl.Collator constructor is the <dfn>%Intl.Collator%</dfn> intrinsic object and a standard built-in property of the Intl object. Behaviour common to all service constructor properties of the Intl object is specified in <emu-xref href="#sec-internal-slots"></emu-xref>.
     </p>
 
     <emu-clause id="sec-intl.collator">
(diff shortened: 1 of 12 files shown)


ljharb:
The rule is that anything that's accessible off of a top-level intrinsic, should be - so if `%Segments%` exists, then i'd expect `%Segments.prototype%`, etc.

linusg:
`%Segments%` doesn't exist, the object is created directly in an AO (https://tc39.es/ecma402/#sec-createsegmentsobject):

```
1. Let internalSlotsList be « [[SegmentsSegmenter]], [[SegmentsString]] ».
2. Let segments be OrdinaryObjectCreate(%SegmentsPrototype%, internalSlotsList).
```

Same for `%SegmentIteratorPrototype%` (https://tc39.es/ecma402/#sec-createsegmentiterator):

```
1. Let internalSlotsList be « [[IteratingSegmenter]], [[IteratedString]], [[IteratedStringNextSegmentCodeUnitIndex]] ».
2. Let iterator be OrdinaryObjectCreate(%SegmentIteratorPrototype%, internalSlotsList).
```

ljharb:
Gotcha; then those two would remain as-is.

gibson042 (commented):
Thanks, this is great! I'm thinking it also needs to include [`oldids`](https://tc39.es/ecmarkup/#oldids) attributes for preserving any current inbound links, and possibly renaming %SegmentsPrototype% to %IntlSegmentsPrototype% and %SegmentIteratorPrototype% to %IntlSegmentIteratorPrototype% as well.

linusg:
Done! I don't think any IDs are changed by the first commit (all links reference the public names already), but I've included `oldids` in the second change.

gibson042 (approved):

PR #846, opened 2023-12-14
Normative: Canonicalize value of extension in ApplyUnicodeExtensionToTag

### LINKED ISSUE #707: `ApplyUnicodeExtensionToTag` and `ResolveLocale` set the `result` record's internal slots to non-canonical values
In `ApplyUnicodeExtensionToTag` and `ResolveLocale`, we loop over the provided relevant extension keys, and set the corresponding internal slot of the `result` record to the value present in either the locale string or `options` object:
```
6. For each element key of relevantExtensionKeys, do
    ...
    g. Set result.[[<key>]] to value.
```

This happens *before* the values are canonicalized via `InsertUnicodeExtensionAndCanonicalize`:
```
9. If newExtension is not the empty String, then
    a. Let locale be ! InsertUnicodeExtensionAndCanonicalize(locale, newExtension)    .
10. Set result.[[locale]] to locale.
11. Return result.
```

Thus, the locale string of the corresponding Intl object is canonicalized, but the extension key internal slots are non-canonical.

The [formatjs](https://github.com/formatjs/formatjs) polyfill shows this issue:
```js
import '@formatjs/intl-locale/polyfill.js'

const loc = new Intl.Locale('en', {calendar: 'islamicc'});
console.log(loc.toString());
console.log(loc.calendar);
```

That prints:
```js
en-u-ca-islamic-civil
islamicc
```

test262 expects `loc.calendar` to be canonicalized (and that is how v8 and spidermonkey already behave).

FrankYFTang:
Close #707

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -136,7 +136,8 @@
           1. Let _optionsValue_ be _options_.[[&lt;_key_&gt;]].
           1. If _optionsValue_ is not *undefined*, then
             1. Assert: Type(_optionsValue_) is String.
-            1. Set _value_ to _optionsValue_.
+            1. Let _optionsUValue_ be the ASCII-lowercase of _optionsValue_.
+            1. Set _value_ to the String value resulting from canonicalizing _optionsUValue_ as a value of key _key_ per <a href="https://unicode.org/reports/tr35/#processing-localeids">Unicode Technical Standard #35 Part 1 Core, Annex C LocaleId Canonicalization Section 5 Canonicalizing Syntax, Processing LocaleIds</a>.
             1. If _entry_ is not ~empty~, then
               1. Set _entry_.[[Value]] to _value_.
             1. Else,


FrankYFTang:
@trflynn89

trflynn89:
Thanks! I tested this change over in [Serenity's LibJS](https://github.com/SerenityOS/serenity) and it looks good to me:
```js
> const loc = new Intl.Locale('en', {calendar: 'islamicc'});

> loc.toString()
"en-u-ca-islamic-civil"

> loc.calendar
"islamic-civil"
```

We also now pass [`test/intl402/Locale/constructor-options-canonicalized.js`](https://github.com/tc39/test262/blob/main/test/intl402/Locale/constructor-options-canonicalized.js) with this change, which was previously failing on this `islamicc` case.

spec/locale.html
```diff
@@ -136,7 +136,8 @@ <h1>
           1. Let _optionsValue_ be _options_.[[&lt;_key_&gt;]].
           1. If _optionsValue_ is not *undefined*, then
             1. Assert: Type(_optionsValue_) is String.
-            1. Set _value_ to _optionsValue_.
+            1. Let _optionsUValue_ be the ASCII-lowercase of _optionsValue_.
+            1. Set _value_ to the String value resulting from canonicalizing _optionsUValue_ as a value of key _key_ per <a href="https://unicode.org/reports/tr35/#processing-localeids">Unicode Technical Standard #35 Part 1 Core, Annex C LocaleId Canonicalization Section 5 Canonicalizing Syntax, Processing LocaleIds</a>.
```
anba, line 140: [ResolveLocale](https://tc39.es/ecma402/#sec-resolvelocale) also explicitly converts `key` to ASCII lower-case before this step. Should we do the same here? (`key` is already in ASCII lower-case, so it's not strictly required to perform any case conversions, neither here nor in `ResolveLocale`.)

anba (approved):
:+1:  LGTM.

spec/locale.html
FrankYFTang: In here, Inside 14.1.3 ApplyUnicodeExtensionToTag ( tag, options, relevantExtensionKeys )
https://tc39.es/ecma402/#sec-apply-unicode-extension-to-tag
5. For each element key of relevantExtensionKeys, do

and only place call it is in 
14.1.1 Intl.Locale ( tag [ , options ] )
https://tc39.es/ecma402/#sec-Intl.Locale
29. Let r be [ApplyUnicodeExtensionToTag](https://tc39.es/ecma402/#sec-apply-unicode-extension-to-tag)(tag, opt, relevantExtensionKeys).

where relevantExtensionKeys is from 
2. Let relevantExtensionKeys be [%Locale%](https://tc39.es/ecma402/#sec-intl-locale-constructor).[[RelevantExtensionKeys]].

and  [%Locale%](https://tc39.es/ecma402/#sec-intl-locale-constructor).[[RelevantExtensionKeys]] is defined in 
https://tc39.es/ecma402/#sec-intl.locale-internal-slots
"The value of the [[RelevantExtensionKeys]] internal slot is « "ca", "co", "hc", "kf", "kn", "nu" ». If [%Collator%](https://tc39.es/ecma402/#sec-the-intl-collator-constructor).[[RelevantExtensionKeys]] does not contain "kf", then remove "kf" from [%Locale%](https://tc39.es/ecma402/#sec-intl-locale-constructor).[[RelevantExtensionKeys]]. If [%Collator%](https://tc39.es/ecma402/#sec-the-intl-collator-constructor).[[RelevantExtensionKeys]] does not contain "kn", then remove "kn" from [%Locale%](https://tc39.es/ecma402/#sec-intl-locale-constructor).[[RelevantExtensionKeys]]."

I agree with you that the "converts key to ASCII lower-case" in [ResolveLocale](https://tc39.es/ecma402/#sec-resolvelocale) is also unnecessary. But that is another editoral issue we can change separately

gibson042 (approved):

littledan:
So the rule here is, "canonicalize if we recognize something, but if there's an unrecognized value, let it continue as uninterpreted text" right? And because of this latter part, we can't convert things like first day of the week into a number in this place, right? I'm OK with this PR, but if you could summarize how these things fit together, that'd help my understanding.

FrankYFTang:
Ben- why do you merge this PR now. This PR is on the agenda for TC39 discussion tomorrow https://github.com/tc39/agendas/blob/main/2024/02.md we should not merge it until TC39 agree with it first.

FrankYFTang:
This is on the agenda for discussion on 2024-02-06 TC39 meeting https://github.com/tc39/agendas/blob/main/2024/02.md

ben-allen:
oh my goodness, sorry about that! I've pushed a revert commit for it.

FrankYFTang:
Has concensus from TG1 this morning around 11:25am PST (2024-02-06)

 Please reland this. @ben-allen

PR #847, opened 2023-12-15
Editorial: Remove a paragraph for the length property of String.prototype.localeCompare

arai-a:
This style follows other "length" properties in this spec and the ECMAScript spec.

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -31,10 +31,6 @@
         1. Return CompareStrings(_collator_, _S_, _thatValue_).
       </emu-alg>
 
-      <p>
-        The value of the *"length"* property of the `localeCompare` method is 1.
-      </p>
-
       <emu-note>
         The `localeCompare` method itself is not directly suitable as an argument to `Array.prototype.sort` because the latter requires a function of two arguments.
       </emu-note>


spec/locale-sensitive-functions.html
```diff
@@ -32,7 +32,7 @@ <h1>String.prototype.localeCompare ( _that_ [ , _locales_ [ , _options_ ] ] )</h
       </emu-alg>
 
       <p>
-        The value of the *"length"* property of the `localeCompare` method is 1.
+        The value of the *"length"* property of the `localeCompare` method is *1*<sub>𝔽</sub>.
```
gibson042, line 35: Even more consistency:
```suggestion
        The *"length"* property of the `localeCompare` method is *1*<sub>𝔽</sub>.
```
But actually, it shouldn't even be necessary at all because the length of **String.prototype.localeCompare ( that [ , locales [ , options ] ] )** is already 1 per [ECMAScript Standard Built-in Objects](https://tc39.es/ecma262/multipage/ecmascript-standard-built-in-objects.html#sec-ecmascript-standard-built-in-objects):
> Every built-in function object, including constructors, has a **"length"** property whose value is a non-negative integral Number. Unless otherwise specified, this value is the number of required parameters shown in the subclause heading for the function description. Optional parameters and rest parameters are not included in the parameter count.
```suggestion
```

gibson042 (approved):

spec/locale-sensitive-functions.html
arai-a: Thanks!
Indeed, the paragraph isn't necessary.
Updated the PR to just remove the paragraph.

PR #848, opened 2023-12-15
Editorial: Remove unnecessary lowercase conversion of key

FrankYFTang:
Replaced with assertion since all the key come from internal slots that are already all in lowercase ASCII.

This issue is mentioned by @anba in https://github.com/tc39/ecma402/pull/846#r1428263375


ben-allen:
@gibson042 any problems with merging this one?

ryzokuken (approved):
LGTM. @anba could you take a quick look before we hit merge?

anba (approved):

anba:
> LGTM. @anba could you take a quick look before we hit merge?

Looks good to me, too.

gibson042 (approved):

PR #849, opened 2023-12-15
Editorial: Refactor ResolveLocale to simplify calls

gibson042:
This has been bugging me for quite a while.

* Rename _dataLocaleData_ aliases to the more accurate _resolvedLocaleData_
* Update the return value to include locale data rather than its lookup key
* Capitalize the return value static field names

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -70,15 +70,15 @@
         1. If _relevantExtensionKeys_ contains *"kf"*, then
           1. Set _collator_.[[CaseFirst]] to _r_.[[kf]].
         1. Let _dataLocale_ be _r_.[[dataLocale]].
-        1. Let _dataLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
+        1. Let _resolvedLocaleData_ be _localeData_.[[&lt;_dataLocale_&gt;]].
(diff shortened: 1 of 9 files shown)


ben-allen (approved):
I am entirely for this change -- I've found the name `_dataLocaleData_` to be borderline unreadable -- though I'm concerned that others might dislike losing the relationship between the name of the alias and the name of the `[[DataLocale]]` slot. @ryzokuken @sffc @FrankYFTang

sffc:
The word "resolved" to me means that the data went through some resolution algorithm, which I guess it did because of dataLocale. I think better would be to rename the variables to "resolvedLocale" and "resolvedLocaleData".

anba:
I wonder if we could remove the term "dataLocale" from the spec. I've created #853 as a possible alternative to this PR.

ryzokuken (approved):

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-02-22.md#editorial-rename-datalocaledata-aliases-to-the-more-accurate-resolvedlocaledata-849

gibson042:
@ben-allen @ryzokuken This has extended to change the shape of the ResolveLocale return value and refactor an internal slot of NumberFormat and RelativeTimeFormat instances, so I'm requesting re-review before merging.

ryzokuken (approved):

Issue #710, opened 2022-09-22
publish ECMA402 biblio to npm


FrankYFTang:
Per https://github.com/tc39/ecmarkup/pull/425#issuecomment-1254085705
it will be nice if we can publish ECMA402 biblio to npm  like [262 does](https://github.com/tc39/ecma262/blob/e72a6b2134ab1d545d9014c85a52a5a10d26625f/.github/workflows/publish-biblio.yml)

@bakkot   @leobalter @gibson042 @ryzokuken

FrankYFTang:
@sffc

ljharb:
Tokens are stored at the repo level, not the org level, so we’d need to add the token to this repo - we’d also need to manually publish the first version (which we did for 262 as well); i can help with both of those.

ryzokuken:
Hi @ljharb want to pick this up again? I'd be happy to help in any way you might require :D

ljharb:
whoops, sure :-) let me take a look at it this afternoon.

ljharb:
The `NPM_TOKEN_FOR_TC39_USER` secret is already on the repo (i must have done it back in September 2022), so I think all that's needed is a PR with the workflow and script, and just prior to landing it, I can publish the first version of the package.

PR #854, opened 2024-01-24
Meta: publish biblio in github actions

bakkot:
Fixes https://github.com/tc39/ecma402/issues/710. This is just a copy-paste of the relevant files from ecma262 with a find-replace of 262 with 402. It also adds a `LICENSE.md` to the top level of the repo so that it can use the same license for the published biblio package.

In the linked issue @ljharb said he'd need to publish it manually first, but I'm not sure why; I think automation tokens can publish the first version of a package?

### DIFF
--- a/scripts/publish-biblio.sh
+++ b/scripts/publish-biblio.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+
+npx ecmarkup --verbose spec.html --write-biblio biblio/biblio.json /dev/null
+
+cp LICENSE.md biblio/
+
+cd biblio
+
+COMMIT_COUNT=$(git rev-list --count HEAD)
+npm version --no-git-tag-version "2.1.${COMMIT_COUNT}"
+
+SHORT_COMMIT=$(git rev-parse --short HEAD)
+LONG_COMMIT=$(git rev-parse --verify HEAD)
+echo "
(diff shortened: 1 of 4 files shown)


ljharb:
@bakkot in my experience a scoped package always needs to be manually published first, but might as well try it and see.

ctcpip (approved):

ryzokuken (approved):

scripts/publish-biblio.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+
+npx ecmarkup --verbose spec.html --write-biblio biblio/biblio.json /dev/null
```
ljharb, line 6: is perhaps this supposed to be > /dev/null?

ljharb (commented):
the publish script is failing: https://github.com/tc39/ecma402/actions/runs/8083286891/job/22086029500

scripts/publish-biblio.sh
ryzokuken: no, I missed it but `spec.html` does not exist. We need to change it to `spec/index.html`.

bakkot: Ugh, this should be `spec/index.html`. I forgot this project puts its files in a different place than ecma262.

(And no, just `/dev/null` is correct.)

ryzokuken: actually, maybe both?

bakkot: https://github.com/tc39/ecma402/pull/866

PR #855, opened 2024-01-31
Editorial: use new IteratorStepValue AO

bakkot:
https://github.com/tc39/ecma262/pull/3268 added a new AO which lets us simplify use of the iteration protocol. This was the only use in 402.

### DIFF
--- a/spec/listformat.html
+++ b/spec/listformat.html
@@ -383,16 +383,14 @@ Output (List of parts records):
           1. Return a new empty List.
         1. Let _iteratorRecord_ be ? GetIterator(_iterable_, ~sync~).
         1. Let _list_ be a new empty List.
-        1. Let _next_ be *true*.
-        1. Repeat, while _next_ is not *false*,
-          1. Set _next_ to ? IteratorStep(_iteratorRecord_).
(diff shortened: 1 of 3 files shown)


spec/listformat.html
```diff
@@ -383,16 +383,14 @@ <h1>
...
-            1. If Type(_nextValue_) is not String, then
-              1. Let _error_ be ThrowCompletion(a newly created *TypeError* object).
-              1. Return ? IteratorClose(_iteratorRecord_, _error_).
-            1. Append _nextValue_ to _list_.
-        1. Return _list_.
+        1. Repeat,
+          1. Let _next_ be ? IteratorStepValue(_iteratorRecord_).
+          1. If _next_ is ~done~, then
+            1. Return _list_.
+          1. If Type(_next_) is not String, then
```
ljharb, line 390:
```suggestion
          1. If _next_ is not a String, then
```
? or should that wait til it's done holistically

bakkot: Yeah that pattern shows up all over this spec; I don't think it makes sense to change just that one in this PR.

gibson042 (approved):
Thanks, this is an improvement.

PR #868, opened 2024-02-29
Editorial: Document that DefaultLocale() returns must not include a "-u-" extension

gibson042:
This is already required:
* LookupMatcher can return { [[locale]]: DefaultLocale(), ... }.
* ResolveLocale uses that result's [[locale]] as the name of a field in its _localeData_ argument and asserts that the corresponding value is a Record.
* The _localeData_ argument to ResolveLocale is generally provided from the value of a service constructor's [[LocaleData]] slot.
* "Internal slots of Service Constructors" constrains those slots to be Records "that have fields for each locale contained in [[AvailableLocales]]", and constrains their [[AvailableLocales]] slots to include only language tags that "must not have a Unicode locale extension sequence".

Therefore, if DefaultLocale() *were* to return a value with a Unicode locale extension sequence, use of it in ResolveLocale would fail to find the required Record at a field with that name in e.g. the service constructor's [[LocaleData]].

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -118,7 +118,7 @@
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>The returned String value represents the structurally valid (<emu-xref href="#sec-isstructurallyvalidlanguagetag"></emu-xref>) and canonicalized (<emu-xref href="#sec-canonicalizeunicodelocaleid"></emu-xref>) language tag for the host environment's current locale.</dd>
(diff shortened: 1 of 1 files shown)


anba (approved):
This restriction is already present through the definitions in [Internal slots of Service Constructors](https://tc39.es/ecma402/#sec-internal-slots):

> [[AvailableLocales]] is a List that contains structurally valid (6.2.1) and canonicalized (6.2.2) language tags identifying the locales for which the implementation provides the functionality of the constructed objects. Language tags on the list must **not have a Unicode locale extension sequence**. The list must include the **value returned by the DefaultLocale** abstract operation (6.2.3), and [...]

Additionally adding this note to DefaultLocale helps to clarify this restriction, though.

ryzokuken (approved):

PR #870, opened 2024-02-29
Editorial: Better align BestAvailableLocale with BCP 47

gibson042:
Per [RFC 4647 section 3.4 - Lookup](https://www.rfc-editor.org/rfc/rfc4647.html#section-3.4),
> Single letter or digit subtags (including both the letter 'x', which introduces private-use sequences, and the subtags that introduce extensions) are removed at the same time as their closest trailing subtag.

This is relevant for a requested locale such as "en-x-y-z-foo", which should be truncated first to "en-x-y-z" and then to "en" ("y" is removed at the same time as its closest trailing subtag "z" and "x" is removed at the same time as its closest trailing subtag "y", which means all three are removed in one step).

However, the change is still editorial because BestAvailableLocale is never called with _availableLocales_ that would contain a language tag like "en-x-y".

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -92,7 +92,8 @@
         1. Repeat,
           1. If _availableLocales_ contains _candidate_, return _candidate_.
           1. Let _pos_ be the index into _candidate_ of the last occurrence of *"-"* (code unit 0x002D HYPHEN-MINUS). If that character does not occur, return *undefined*.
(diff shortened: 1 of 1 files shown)


sffc:
Related: https://github.com/tc39/ecma402/pull/824

ben-allen:
looks good to me!

ben-allen (approved):

PR #875, opened 2024-03-25
Editorial: update for String{,Last}IndexOf returning not-found

bakkot:
Update for https://github.com/tc39/ecma262/pull/3300. Marked as draft until that lands; I will update this PR to pull in the extra biblio when it does (though the type-checking isn't yet able to do anything with that).

I'm not absolute sure about the changes PartitionPattern (what does  and PartitionNotationSubPattern and would appreciate extra scrutiny there.

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -264,7 +264,7 @@
         1. If IsSanctionedSingleUnitIdentifier(_unitIdentifier_) is *true*, then
           1. Return *true*.
         1. Let _i_ be StringIndexOf(_unitIdentifier_, *"-per-"*, 0).
-        1. If _i_ is -1 or StringIndexOf(_unitIdentifier_, *"-per-"*, _i_ + 1) is not -1, then
+        1. If _i_ is ~not-found~ or StringIndexOf(_unitIdentifier_, *"-per-"*, _i_ + 1) is not ~not-found~, then
           1. Return *false*.
         1. Assert: The five-character substring *"-per-"* occurs exactly once in _unitIdentifier_, at index _i_.
         1. Let _numerator_ be the substring of _unitIdentifier_ from 0 to _i_.
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -165,7 +165,7 @@
         1. Let _k_ be 3.
         1. Repeat, while _k_ &lt; _size_,
           1. Let _e_ be StringIndexOf(_extension_, *"-"*, _k_).
-          1. If _e_ = -1, let _len_ be _size_ - _k_; else let _len_ be _e_ - _k_.
+          1. If _e_ is ~not-found~, let _len_ be _size_ - _k_; else let _len_ be _e_ - _k_.
           1. Let _subtag_ be the substring of _extension_ from _k_ to _k_ + _len_.
           1. NOTE: A <emu-not-ref>keyword</emu-not-ref> is a sequence of subtags in which the first is a key of length 2 and any subsequent ones (if present) have length in the inclusive interval from 3 to 8, collectively constituting a value along with their medial *"-"* separators. An attribute is a single subtag with length in the inclusive interval from 3 to 8 that precedes all <emu-not-ref>keywords</emu-not-ref>.
           1. Assert: _len_ &ge; 2.
@@ -197,7 +197,7 @@
       <emu-alg>
         1. Assert: _locale_ does not contain a Unicode locale extension sequence.
         1. Let _privateIndex_ be StringIndexOf(_locale_, *"-x-"*, 0).
-        1. If _privateIndex_ = -1, then
+        1. If _privateIndex_ is ~not-found~, then
           1. Let _newLocale_ be the string-concatenation of _locale_ and _extension_.
         1. Else,
           1. Let _preExtension_ be the substring of _locale_ from 0 to _privateIndex_.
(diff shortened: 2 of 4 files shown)


bakkot:
Updated the biblio; this is ready for review/landing.

ben-allen:
This looks entirely good to me.

Note, though: As a result of changes made in #840, merging the changes in `PartitionPattern` with the current main is somewhat tricky. I've rebased it on [a branch in my own fork](https://github.com/ben-allen/ecma402/tree/bakkot-875-string-not-found), but getting another set of eyes on it would be prudent. [Here's a direct link to the relevant AO](https://github.com/ben-allen/ecma402/blob/b7395e8188a0dcc13d1113818fe4d04579de1187/spec/negotiation.html#L429)
@gibson042 @bakkot

bakkot:
@ben-allen looks like `_placeholderEnd_` still needs updating in the AO you linked

ben-allen:
> @ben-allen looks like `_placeholderEnd_` still needs updating in the AO you linked

The `Let _placeholderEnd_ be -1` line? That's included (and left unchanged) because `_placeholderEnd_` can't be `~not-found~` &mdash; as I understand it, whenever there's a `{` there's guaranteed to be a matching `}`, and so there's no way inside the `Repeat, while _placeholderStart_ is not ~not-found~` loop for `_placeholderEnd_` to get set to `~not-found~`. 

It's set to -1 before the loop and compared to -1 after the loop so that `_tail_` doesn't get included in `_result_` twice in the case that `_pattern_` has no placeholders.

bakkot:
Ah. In that case it should use a different value than `-1`, like `~none~`, and the `1. Assert: _placeholderStart_ < _placeholderEnd_` line should be changed `Assert: _placeholderEnd_ is not ~not-found~ and _placeholderStart_ < _placeholderEnd_`.

ben-allen:
> Ah. In that case it should use a different value than `-1`, like `~none~`, and the `1. Assert: _placeholderStart_ < _placeholderEnd_` line should be changed `Assert: _placeholderEnd_ is not ~not-found~ and _placeholderStart_ < _placeholderEnd_`.

Ah, brilliant &mdash; changing now

ben-allen:
[link to updated `PartitionPattern`](https://github.com/ben-allen/ecma402/blob/c34038b1dd9162d97fbd3e1bc9a5c9383b6ca4cc/spec/negotiation.html#L431)

bakkot:
Ah, hm, it didn't occur to me that this does mess up the first line of the loop.

```
1. Let _literal_ be the substring of _pattern_ from _placeholderEnd_ + 1 to _placeholderStart_.
```

I guess using -1 is fine and just updating the assert.

ben-allen:
Apologies for all the fixups! [Here's](https://github.com/ben-allen/ecma402/blob/675e0364a453174bd148c2c9d8a6b975dc564152/spec/negotiation.html#L431) a rewritten version using `~none~` rather than `-1` to initialize `_placeholderEnd_`. This is slightly messy &mdash; I'm sure there's a more elegant way &mdash; but perhaps it's less opaque than the `-1` version.

bakkot:
That version of `PartitionPattern` LGTM, thanks for taking care of that!

ben-allen:
@bakkot with your permission I'll push what I've got [here](https://github.com/tc39/ecma402/compare/main...ben-allen:ecma402:bakkot-875-string-not-found?expand=1) to your branch and merge.

bakkot:
Go for it.

gibson042:
FWIW, I think it would be more readable for _placeholderEnd_ to still be initialized to -1 and remain integer-valued.
```
1. Let _placeholderEnd_ be -1.
1. Let _placeholderStart_ be StringIndexOf(_pattern_, *"{"*, 0).
1. Repeat, while _placeholderStart_ is not ~not-found~,
  1. Let _literal_ be the substring of _pattern_ from _placeholderEnd_ + 1 to _placeholderStart_.
  1. If _literal_ is not the empty String, then
    1. Append the Record { [[Type]]: *"literal"*, [[Value]]: _literal_ } to _result_.
  1. Set _placeholderEnd_ to StringIndexOf(_pattern_, *"}"*, _placeholderStart_).
  1. Assert: _placeholderEnd_ is not ~not-found~ and _placeholderStart_ < _placeholderEnd_.
  1. Let _placeholderName_ be the substring of _pattern_ from _placeholderStart_ + 1 to _placeholderEnd_.
  1. Append the Record { [[Type]]: _placeholderName_, [[Value]]: *undefined* } to _result_.
  1. Set _placeholderStart_ to StringIndexOf(_pattern_, *"{"*, _placeholderEnd_).
1. Let _tail_ be the substring of _pattern_ from _placeholderEnd_ + 1.
1. If _tail_ is not the empty String, then
  1. Append the Record { [[Type]]: *"literal"*, [[Value]]: _tail_ } to _result_.
1. Return _result_.
```

ben-allen:
> FWIW, I think it would be more readable for _placeholderEnd_ to still be initialized to -1 and remain integer-valued.
> 
> ```
> 1. Let _result_ be a new empty List.
> 1. Let _placeholderEnd_ be -1.
> 1. Let _placeholderStart_ be StringIndexOf(_pattern_, *"{"*, 0).
> 1. Repeat, while _placeholderStart_ is not ~not-found~,
>   1. Let _literal_ be the substring of _pattern_ from _placeholderEnd_ + 1 to _placeholderStart_.
>   1. If _literal_ is not the empty String, then
>     1. Append the Record { [[Type]]: *"literal"*, [[Value]]: _literal_ } to _result_.
>   1. Set _placeholderEnd_ to StringIndexOf(_pattern_, *"}"*, _placeholderStart_).
>   1. Assert: _placeholderEnd_ is not ~not-found~ and _placeholderStart_ < _placeholderEnd_.
>   1. Let _placeholderName_ be the substring of _pattern_ from _placeholderStart_ + 1 to _placeholderEnd_.
>   1. Append the Record { [[Type]]: _placeholderName_, [[Value]]: *undefined* } to _result_.
>   1. Set _placeholderStart_ to StringIndexOf(_pattern_, *"{"*, _placeholderEnd_).
> 1. Let _tail_ be the substring of _pattern_ from _placeholderEnd_ + 1.
> 1. If _tail_ is not the empty String, then
>   1. Append the Record { [[Type]]: *"literal"*, [[Value]]: _tail_ } to _result_.
> 1. Return _result_.
> ```

Upon consideration, I'm given to agree. My line of reasoning is roughly "Which is more potentially annoying to an implementer? The use of -1 instead of something self-documenting, or the additional tests for `~none~` / weirdness involved in making sure that a pattern without placeholders not get appended to `_result_` twice?" and thinking that -1 indeed less annoying.

PR #876, opened 2024-03-31
Editorial: align time zone identifier text and AOs with ECMA-262

justingrant:
This PR updates ECMA-402 spec text to align with current ECMA-262 terminology and AOs for time zone identifiers. Changes include:

* Replaces `DefaultTimeZone`,  `CanonicalizeTimeZoneName`, and `IsValidTimeZoneName` with calls to current ECMA-262 abstract operations.
* Refactors `AvailableCanonicalTimeZones`  to call new ECMA-262 AOs, and renames it to `AvailablePrimaryTimeZoneIdentifiers`.
* Adds a `GetAvailableNamedTimeZoneIdentifier` AO Temporal will eventually merge into ECMA-262, but it's not there yet so we need a copy here.
* Replaces `#sec-time-zone-names` with a "Use of the IANA Time Zone Database" section that extends ECMA-262's `#sec-time-zone-identifiers` section.
* Adds a few paragraphs of editorial recommendations (borrowed from Temporal) about managing updates and builds of TZDB.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -107,10 +107,10 @@
           1. Let _offsetMinutes_ be _offsetNanoseconds_ / (6 × 10<sup>10</sup>).
           1. Assert: _offsetMinutes_ is an integer.
           1. Set _timeZone_ to FormatOffsetTimeZoneIdentifier(_offsetMinutes_).
-        1. Else if IsValidTimeZoneName(_timeZone_) is *true*, then
-          1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZone_).
         1. Else,
-          1. Throw a *RangeError* exception.
+          1. Let _timeZoneIdentifierRecord_ be GetAvailableNamedTimeZoneIdentifier(_timeZone_).
+          1. If _timeZoneIdentifierRecord_ is ~empty~, throw a *RangeError* exception.
+          1. Set _timeZone_ to _timeZoneIdentifierRecord_.[[PrimaryIdentifier]].
         1. Set _dateTimeFormat_.[[TimeZone]] to _timeZone_.
         1. Let _formatOptions_ be a new Record.
         1. Set _formatOptions_.[[hourCycle]] to _hc_.
@@ -1266,7 +1266,7 @@
         1. If IsTimeZoneOffsetString(_timeZoneIdentifier_) is *true*, then
           1. Let _offsetNs_ be ParseTimeZoneOffsetString(_timeZoneIdentifier_).
         1. Else,
-          1. Assert: IsValidTimeZoneName(_timeZoneIdentifier_) is *true*.
+          1. Assert: GetAvailableNamedTimeZoneIdentifier(_timeZoneIdentifier_) is not ~empty~.
           1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_timeZoneIdentifier_, _epochNs_).
         1. Let _tz_ be ℝ(_epochNs_) + _offsetNs_.
         1. If _calendar_ is *"gregory"*, then
@@ -1274,10 +1274,6 @@
         1. Else,
           1. Return a ToLocalTime Record with the fields calculated from _tz_ for the given _calendar_. The calculations should use best available information about the specified _calendar_.
       </emu-alg>
-
-      <emu-note>
-        A conforming implementation must recognize *"UTC"* and all Zone and Link names from the IANA Time Zone Database (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.
-      </emu-note>
     </emu-clause>
 
     <emu-clause id="sec-datetimeformat-tolocaltime-records">
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -164,59 +164,111 @@
     </dl>
   </emu-clause>
 
-  <emu-clause id="sec-time-zone-names">
-    <h1>Time Zone Names</h1>
+  <emu-clause id="sec-use-of-iana-time-zone-database" oldids="sec-time-zone-names">
+    <h1>Use of the IANA Time Zone Database</h1>
 
     <p>
-      This specification identifies time zones using the Zone and Link names of the IANA Time Zone Database. Their canonical form is the corresponding Zone name in the casing used in the IANA Time Zone Database except as specifically overridden by CanonicalizeTimeZoneName.
+      Implementations that adopt this specification must be time zone aware: they must use the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a> to supply time zone identifiers and data used in ECMAScript calculations and formatting.
+      This section defines how the IANA Time Zone Database should be used by time zone aware implementations.
+    </p>
+    <p>
+      Except as overridden by AvailableNamedTimeZoneIdentifiers, each Zone in the IANA Time Zone Database must be a primary time zone identifier and each Link name in the IANA Time Zone Database must be a non-primary time zone identifier.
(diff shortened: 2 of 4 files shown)


spec/datetimeformat.html
```diff
@@ -107,10 +107,10 @@ <h1>
           1. Let _offsetMinutes_ be _offsetNanoseconds_ / (6 × 10<sup>10</sup>).
           1. Assert: _offsetMinutes_ is an integer.
           1. Set _timeZone_ to FormatOffsetTimeZoneIdentifier(_offsetMinutes_).
-        1. Else if IsValidTimeZoneName(_timeZone_) is *true*, then
-          1. Set _timeZone_ to CanonicalizeTimeZoneName(_timeZone_).
         1. Else,
-          1. Throw a *RangeError* exception.
+          1. Let _timeZoneIdentifierRecord_ be GetAvailableNamedTimeZoneIdentifier(_timeZone_).
+          1. If _timeZoneIdentifierRecord_ is ~empty~, throw a RangeError exception.
```
ptomato, line 112:
```suggestion
          1. If _timeZoneIdentifierRecord_ is ~empty~, throw a *RangeError* exception.
```

spec/datetimeformat.html
```diff
@@ -1266,7 +1266,7 @@ <h1>
         1. If IsTimeZoneOffsetString(_timeZoneIdentifier_) is *true*, then
           1. Let _offsetNs_ be ParseTimeZoneOffsetString(_timeZoneIdentifier_).
         1. Else,
-          1. Assert: IsValidTimeZoneName(_timeZoneIdentifier_) is *true*.
+          1. Assert: GetAvailableNamedTimeZoneIdentifier(_timeZoneIdentifier_) is not ~empty~.
           1. Let _offsetNs_ be GetNamedTimeZoneOffsetNanoseconds(_timeZoneIdentifier_, _epochNs_).
         1. Let _tz_ be ℝ(_epochNs_) + _offsetNs_.
         1. If _calendar_ is *"gregory"*, then
```
ptomato, line 1272: I think we can remove the note a few lines below here that says

> A conforming implementation must recognize *"UTC"* and all Zone and Link names from the IANA Time Zone Database (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations.

It's now covered in "Use of the IANA Time Zone Database".

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,111 @@ <h1>
...
       </dl>
+      <p>This definition supersedes the definition provided in <emu-xref href="#sec-availablenamedtimezoneidentifiers"></emu-xref>.</p>
       <emu-alg>
-        1. If one of the Zone or Link names of the IANA Time Zone Database is an ASCII-case-insensitive match for _timeZone_, return *true*.
-        1. If _timeZone_ is an ASCII-case-insensitive match for *"UTC"*, return *true*.
-        1. Return *false*.
+        1. Let _identifiers_ be a List containing the String value of each Zone or Link name in the IANA Time Zone Database.
+        1. Assert: No element of _identifiers_ is an ASCII-case-insensitive match for any other element.
+        1. Assert: Every element of _identifiers_ identifies a Zone or Link name in the IANA Time Zone Database.
+        1. Sort _identifiers_ into the same order as if an Array of the same values had been sorted using %Array.prototype.sort% with *undefined* as the argument.
```
ptomato, line 223:
```suggestion
        1. Let _identifiers_ be the List of each Zone and Link name in the IANA Time Zone Database, sorted according to lexicographic code unit order.
        1. Assert: No element of _identifiers_ is an ASCII-case-insensitive match for any other element.
        1. Assert: Every element of _identifiers_ identifies a Zone or Link name in the IANA Time Zone Database.
```

I opened https://github.com/tc39/ecma262/pull/3299 to get rid of this awkward way to specify sorting. The language the 262 editors settled on is "sorted according to lexicographic code unit order." That PR has no further comments and I expect it'll be merged this week, so I think it's safe to use that language here as well.

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,111 @@ <h1>
...
+        1. Let _result_ be a new empty List.
+        1. For each element _identifier_ of _identifiers_, do
+          1. Let _primary_ be _identifier_.
+          1. If _identifier_ is a non-primary time zone identifier and _identifier_ is not *"UTC"*, then
```
ptomato, line 227:
```suggestion
          1. If _identifier_ is a Link name and _identifier_ is not *"UTC"*, then
```
We use "rules for resolving Link names" in the step below, so it might be good to use the same language here.

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,111 @@ <h1>
...
+          1. Let _record_ be the Time Zone Identifier Record { [[Identifier]]: _identifier_, [[PrimaryIdentifier]]: _primary_ }.
+          1. Append _record_ to _result_.
+        1. Assert: _result_ contains a Time Zone Identifier Record _r_ such that _r_.[[Identifier]] is *"UTC"* and _r_.[[PrimaryIdentifier]] is *"UTC"*.
+        1. Return _result_.
       </emu-alg>
-    </emu-clause>
 
-    <emu-note>
-      Any value returned from SystemTimeZoneIdentifier that is not recognized as valid by IsTimeZoneOffsetString must be recognized as valid by IsValidTimeZoneName.
-    </emu-note>
```
ptomato, line 197: Do we want to keep this note, but in a different place?

spec/locales-currencies-tz.html
```diff
@@ -225,14 +277,11 @@ <h1>
...
         1. Let _result_ be a new empty List.
-        1. For each element _name_ of _names_, do
-          1. Assert: IsValidTimeZoneName( _name_ ) is *true*.
-          1. Let _canonical_ be CanonicalizeTimeZoneName( _name_ ).
-          1. If _result_ does not contain _canonical_, then
-            1. Append _canonical_ to _result_.
-        1. [declared="comparefn"] Sort _result_ in order as if an Array of the same values had been sorted using %Array.prototype.sort% using *undefined* as _comparefn_.
+        1. For each element _timeZoneIdentifierRecord_ of _records_, do
+          1. If _timeZoneIdentifierRecord_.[[Identifier]] is _timeZoneIdentifierRecord_.[[PrimaryIdentifier]], then
+            1. Append _timeZoneIdentifierRecord_.[[Identifier]] to the end of _result_.
```
ptomato, line 284:
```suggestion
            1. Append _timeZoneIdentifierRecord_.[[Identifier]] to _result_.
```

ptomato (approved):
Works for me!

spec/locales-currencies-tz.html
justingrant: Well, the second step of [SystemTimeZoneIdentifier](https://tc39.es/ecma262/#sec-systemtimezoneidentifier) in ECMA-262 seems to say the same thing: 

> Let _systemTimeZoneString_ be the String representing the host environment's current time zone, either a primary time zone identifier or an offset time zone identifier.

So IMO we have this one covered, given that SystemTimeZoneIdentifier is no longer overridden in ECMA-402. What do you think?

ptomato: Sounds good.

spec/locales-currencies-tz.html
ptomato: That works for me.

justingrant:
@sffc @gibson042 what is the process for ECMA-402 editorial PRs?  I'd like to ideally get this reviewed before it gets too stale.

sffc:
The process is to get a review from  @gibson042 or @ben-allen.

justingrant:
> The process is to get a review from @gibson042 or @ben-allen.

@gibson042 @ben-allen Possible to review this before it gets too stale?  Thanks!

Issue #806, opened 2023-07-13
Draft plan to align canonical time zone IDs across implementations


justingrant:
This issue proposes some ideas and a draft plan for how implementations can align on a common set of canonical time zone IDs, in order to fix problems like:
* Variation in canonicalization behavior between Chrome/Safari vs Firefox
* Chrome and Safari's returning outdated IDs like Asia/Calcutta and Europe/Kiev, causing [many user complaints](https://docs.google.com/presentation/d/1MVBKAB8U16ynSHmO6Mkt26hT5U-28OjyG9-L-GFdikE/edit#slide=id.g22181d24971_0_102).

There are three questions to answer:

1. Which IDs should be canonical?
2. How should we get these IDs into implementations?
3. When should we ship these changes?

This is an early draft, so please let me know if I made mistakes below or if you see a better way to achieve the goal of using up-to-date canonical IDs in ECMAScript. Note that this plan below is complimentary but unrelated to the now-Stage 3 [proposal-caonical-tz](https://github.com/tc39/proposal-canonical-tz) proposal. 

Feedback is welcome, especially from @sffc @FrankYFTang @anba @Constellation @gibson042 @dminor.

## 1. Which IDs should be canonical?

To avoid messy geopolitical judgement calls, I recommend that we defer to the IANA Time Zone Database to decide which IDs should be canonical, using the following simple rules:
* **a) Every ID in [zone.tab](https://github.com/eggert/tz/blob/main/zone.tab) should be canonical.** zone.tab lists at least one unique time zone for each ISO 3166-1 country code. Time zone changes in a country will not affect any other country. Note that Firefox breaks this rule for 11 country codes that are very closely associated, like Europe/Vatican and Europe/Rome. I'd be curious to hear @anba's opinion about whether these 11 merges are a good idea or not.
* **b) If the current ECMAScript canonical ID is an outdated name for the ID in zone.tab, then the ECMAScript ID should become a non-canonical ID that resolves to the ID in zone.tab.** Chrome and Safari have 19 cases like this, where old ICU IDs like Asia/Calcutta and Europe/Kiev are still being used even though they've been replaced in the IANA TZDB.

The JSON objects below ere generated by a simple JS app you can run and edit at https://codesandbox.io/s/zone-tab-mismatches-mlf93j. The results for Chrome and Safari, which use ICU for canonicalization, are identical. Firefox, which overrides ICU's canonicalization, so it has different results.

[...]

Issue #778, opened 2023-04-27
`Intl.supportedValuesOf('timeZone')` for UTC, Etc/GMT*, and other non-contintent/non-ocean time zone identifiers


justingrant:
As part of work on [proposal-canonical-tz](https://github.com/tc39/proposal-canonical-tz), I noticed that JS engines have very different results from `Intl.supportedValuesOf('timeZone')` for time zones that are not associated with a continent or ocean. 

V8 does not return any of these zones. Safari returns only one: UTC. Firefox returns 40 zones, including "UTC", Etc/ zones like "Etc/GMT+11", and a few others like "EST" and "CET".

Should the 402 spec be more prescriptive about what kinds of canonical time zone identifiers should vs. should not be returned by this API?

```
Intl.supportedValuesOf('timeZone').filter(z => !['Asia', 'Africa', 'Europe', 'Australia', 'America', 'Antarctica', 'Atlantic', 'Pacific', 'Indian', 'Arctic'].some(c => z.startsWith(c)))

// => Chrome: []
// => Safari: ["UTC"]
// => Firefox: [
  "CET",
  "CST6CDT",
  "EET",
  "EST",
  "EST5EDT",
  "Etc/GMT+1",
  "Etc/GMT+10",
  "Etc/GMT+11",
  "Etc/GMT+12",
  "Etc/GMT+2",
  "Etc/GMT+3",
  "Etc/GMT+4",
  "Etc/GMT+5",
  "Etc/GMT+6",
  "Etc/GMT+7",
  "Etc/GMT+8",
  "Etc/GMT+9",
  "Etc/GMT-1",
  "Etc/GMT-10",
  "Etc/GMT-11",
  "Etc/GMT-12",
  "Etc/GMT-13",
  "Etc/GMT-14",
  "Etc/GMT-2",
  "Etc/GMT-3",
  "Etc/GMT-4",
  "Etc/GMT-5",
  "Etc/GMT-6",
  "Etc/GMT-7",
  "Etc/GMT-8",
  "Etc/GMT-9",
  "Factory",
  "HST",
  "MET",
  "MST",
  "MST7MDT",
  "PST8PDT",
  "UTC",
  "WET"
]
```

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2023-05-04.md#intlsupportedvaluesoftimezone-for-utc-etcgmt-and-other-non-contintentnon-ocean-time-zone-identifiers-778

[...]

PR #877, opened 2024-03-31
Normative: specify time zone ID requirements to reduce divergence between engines

### LINKED ISSUE #825: Should ECMA-402 spec text for time zone canonicalization refer to CLDR or to IANA as authoritative?
CLDR in https://github.com/unicode-org/cldr/pull/3105 will soon provide the ability to fetch modern canonical IDs for currently-problematic time zones like Asia/Calcutta and Europe/Kiev. ICU will also be adding an API to expose the IANA canonical ID. This will enable V8 and JSC to finally expose modern IANA canonical names like SpiderMonkey does, but without the separate IANA-based overrides that SpiderMonkey has had to maintain.

Before this change, it didn't make sense to have normative spec text in 402 to define which IDs should be primary (the new 262 term for canonical time zone ID) vs. non-primary. But with this CLDR/ICU change, it's finally practical to specify normative rules for determining which IDs are primary (and, if not, which primary ID they resolve to).

I can draft a PR with this normative text for discussion, but first there's one main question to answer: should we specify the rules solely as using CLDR data, solely as using IANA data, or  should we specify rules for which IANA IDs are canonical that just happens to match what CLDR is doing?

As an illustration, here's two possible directions that we could take this spec text. Don't worry about the particular text used (it's very rough and will change) but the general approach of depending on CLDR vs. depending on IANA is where I'm most looking for feedback.

Which one is better?  @sffc @gibson042 @anba @Constellation @FrankYFTang

### Option A - Defer to CLDR

The Unicode Common Locale Data Repository (CLDR) defines which available named time zone identifiers are primary or non-primary, as well as which non-primary time zone identifiers resolve to which primary time zone identifiers. The following exceptions are applied: 
* For historical reasons, *"UTC"* is a primary time zone identifier while *"Etc/UTC"*, *"Etc/GMT"*, and *"GMT"* (and the identifiers that resolve to them) are non-primary time zone identifiers that resolve to *"UTC"*.
* CLDR identifiers that are not present in the IANA Time Zone Database (*"Etc/Unknown"*, *"Canada/East-Saskatchewan"*, and *"US/Pacific-New"*) are not supported.
*  *"Factory"* is not supported.

The following spec text would replace the steps of https://tc39.es/proposal-temporal/#sup-availablenamedtimezoneidentifiers: 

1. For each `<type>`element that contains an `alias` attribute in `timezone.xml` in the Unicode Common Locale Data Repository (CLDR), do
   1. Let _aliases_ be a new List, populated by splitting the space-delimited `alias` attribute.
   1. If an  `iana` attribute is present, let _primary_ be the String value of that attribute; otherwise, let _primary_ be the first element in _aliases_.
   1. If _primary_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, set _primary_ to *"UTC"*.
   1. If _primary_ is not *"Etc/Unknown"* nor *"Factory"* , then
      1. Let _record_ be the Time Zone Identifier Record { [[Identifier]]: _primary_, [[PrimaryIdentifier]]: _primary_ }.
      1. Append record to _result_.
      1. For each element _identifier_ in _aliases_, do
         1. If _identifier_ is not _primary_ and is not one of *"Canada/East-Saskatchewan"* nor *"US/Pacific-New"*,  then
            1. Set _record_ to  the Time Zone Identifier Record { [[Identifier]]: _primary_, [[PrimaryIdentifier]]: _primary_ }.
            1. Append record to _result_.
1. Sort _result_ lexicographically by UTF-16 code unit of each element's [[Identifier]] field, in ascending order
1. Assert: _result_ contains a Time Zone Identifier Record _r_ such that _r_.[[Identifier]] is "UTC" and _r_.[[PrimaryIdentifier]] is "UTC".
1. Return _result_.

### Option B - Define using IANA only

Each Zone in the IANA Time Zone Database must be a primary time zone identifier and each Link name in the IANA Time Zone Database must be a non-primary time zone identifier that resolves to its corresponding Zone name, with the following exceptions:
* For historical reasons, "UTC" must be a primary time zone identifier.  "Etc/UTC", "Etc/GMT", and "GMT", as well as all Link names that resolve to any of them, must be non-primary time identifiers that resolve to "UTC".
* Any Link name in the `TZ` column of `zone.tab` of the IANA Time Zone Database must be a primary time zone identifier.
* Any Link name that represents a geographical area entirely contained within the territory of a single ISO 3166-2 country code must resolve to a primary identifier that also represents a geographical area entirely contained within the territory of the same ISO 3166-2 country code. For example, "Atlantic/Jan_Mayen" must resolve to "Arctic/Longyearbyen".
* The following legacy POSIX identifiers must resolve to their Continent/City equivalents, as shown in the table below:

| Legacy POSIX Zone Name | Primary Time Zone Identifier |
| -------------------------- | ----------------------------- |
| EST | Etc/GMT+5 |
| MST | Etc/GMT+7 |
| HST | Etc/GMT+10 |
| EST5EDT | America/New_York |
| CST6CDT | America/Chicago |
| MST7MDT | America/Denver |
| PST8PDT | America/Los_Angeles |
| WET | Europe/Lisbon |
| CET | Europe/Berlin |
| MET | Europe/Vienna |
| EET | Europe/Athens |

### Option C - Define IANA rules, but explain how to use `timezone.xml`

We could also merge both (A) and (B): define the IANA \*and\* explain how to use `timezone.xml` data to satisfy those rules. I'm not going to draft text for this yet because I'm unsure if anyone would want it, but I wanted to include this option here for discussion.

justingrant:
This proposed change resolves #825 by adding normative spec text to clarify how ECMA-402 implementations should decide which IANA time zone IDs should be primary vs. non-primary. This will enable more consistency between ECMAScript implementations and prevent future divergence.

This PR also accommodates to web reality by aligning ECMA-402 with CLDR and ICU. This should make it easier for all ECMAScript engines to comply with the spec while still being able to use ICU. 

This PR is stacked on #876, so please ignore the first commit when reviewing this PR.

Note that the problem of out-of-date primary IDs for renamed Zones (like Asia/Calcutta) is out of scope to this PR, because the [spec](https://tc39.es/ecma402/#sec-canonicalizetimezonename) already requires current IDs, and there's already a [plan](https://github.com/tc39/ecma402/issues/806#issuecomment-1904654534) to fix it that requires no spec changes.

### Summary of proposed changes

This PR implements "Option C" in https://github.com/tc39/ecma402/issues/825 by deterministically defining ECMAScript's exceptions from the IANA Time Zone Database's defaults, and then pointing implementers at ICU as a convenient implementation of those exceptions.

We'll start with a baseline of IANA's Zone and Link names and specify a few exceptions:
1. Existing [special-cases](https://tc39.es/ecma402/#sec-canonicalizetimezonename) for Etc/UTC, Etc/GMT, and GMT will be retained.
2. Primary and non-primary IDs must share the same ISO-3166-2 country code, so IANA's time zone merges between countries will be reverted. CLDR and ICU already disallow merges across country boundaries. Note that IANA's merges of time zones \*inside\* of a country (e.g. Asia/Chongquing=>Asia/Shanghai) will be accepted, also matching current ICU/CLDR behavior.

This PR also makes two other smaller text changes that we expect to have zero impact on current engines:
* Changes #876's recommendations into requirements to limit observable dynamic updating of TZDB info. AFAIK, no existing ECMAScript engine updates TZDB (observably or otherwise!) during the lifetime of the surrounding agent, so this is really a future-compatibility change.
* Adds text recommending a two-year waiting period before a newly-renamed ID becomes primary. These are rare (the last was 2022's Europe/Kiev=>Europe/Kyiv) and when the next rename happens, we'll try to convince CLDR to implement this requirement for us. There are no pending renames, so this is also a future-compatibility change.

### Per-engine changes required

Implementing the changes in this PR will impact JS engines differently, given the current divergence between engines:

* For V8 (cc @FrankYFTang) and JSC (cc @Constellation), there should be no change because they rely on ICU and CLDR which already behave like 1-3 above. Note that this PR doesn't affect the [plan](https://github.com/tc39/ecma402/issues/806#issuecomment-1904654534) to fix out-of-date canonicalizations like Asia/Calcutta and Europe/Kiev. This plan is unchanged: as part of landing Temporal Stage 4, switch to use ICU's new [`icu::TimeZone::getIanaID()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1TimeZone.html#a7e182f60aa75c78a623065ad8470d290), which returns the latest IANA IDs instead of out-of-date canonical IDs like Asia/Calcutta.

* For SpiderMonkey (cc @anba), more changes are needed because currently SpiderMonkey conforms to the [spec](https://tc39.es/ecma402/#sec-canonicalizetimezonename) which requires using `backward` in TZDB to determine canonicalization. SM could use  [`icu::TimeZone::getIanaID()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1TimeZone.html#a7e182f60aa75c78a623065ad8470d290) to implement (2) and (3) above, or could implement the same behavior by reading CLDR data or IANA data directly. Also, this PR will reduce SM's [differences](https://github.com/tc39/ecma402/issues/778) in `Intl.supportedValuesOf('timeZone')` vs. V8/JSC.

### Testing

Test262 changes will be needed to validate these normative changes, but I'm not sure how we can run those tests except using the Temporal polyfill. @ptomato I'll be looking for your advice (and perhaps help writing tests!) on this point.

### Feedback requested

Feedback is welcome on any part of this proposal, but I'm most interested in making sure that the spec text actually accomplishes what the summary above claims that it does.

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -170,39 +170,74 @@
     <p>
       Implementations that adopt this specification must be time zone aware: they must use the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a> to supply time zone identifiers and data used in ECMAScript calculations and formatting.
       This section defines how the IANA Time Zone Database should be used by time zone aware implementations.
-    </p>
-    <p>
-      Except as overridden by AvailableNamedTimeZoneIdentifiers, each Zone in the IANA Time Zone Database must be a primary time zone identifier and each Link name in the IANA Time Zone Database must be a non-primary time zone identifier.
       No String may be an available named time zone identifier unless it is a Zone name or a Link name in the IANA Time Zone Database.
       Available named time zone identifiers returned by ECMAScript built-in objects must use the casing found in the IANA Time Zone Database.
     </p>
     <p>
-      In the IANA Time Zone Database, the UTC time zone is represented by the Zone *"Etc/UTC"* which is distinct from the Zone *"Etc/GMT"*.
-      For historical reasons, ECMAScript uses *"UTC"* as the primary identifier for the former Zone and does not recognize the latter Zone as distinct, instead requiring *"Etc/UTC"*, *"Etc/GMT"*, and *"GMT"* (if available) to be non-primary identifiers that resolve to *"UTC"*.
-      This is the only deviation from the IANA Time Zone Database that is required of a time zone aware ECMAScript implementation.
+      Each Zone in the IANA Time Zone Database must be a primary time zone identifier and each Link name in the IANA Time Zone Database must be a non-primary time zone identifier that resolves to its corresponding Zone name, with the following exceptions implemented in AvailableNamedTimeZoneIdentifiers:
     </p>
-    <p>
-      The IANA Time Zone Database is typically updated between five and ten times per year.
-      These updates may add new Zone or Link names, may change Zones to Links, and may change the UTC offsets and transitions associated with any Zone.
-      ECMAScript implementations are recommended to include updates to the IANA Time Zone Database as soon as possible.
-      Such prompt action ensures that ECMAScript programs can accurately perform time-zone-sensitive calculations and can use newly-added available named time zone identifiers supplied by external input or the host environment.
-    </p>
-
-    <p>
-      If implementations revise time zone information during the lifetime of an agent, then it is recommended that the list of available named time zone identifiers, the primary time zone identifier associated with any available named time zone identifier, and the UTC offsets and transitions associated with any available named time zone identifier, be consistent with results previously observed by that agent.
-      Due to the complexity of supporting this recommendation, it is recommended that implementations maintain a fully consistent copy of the IANA Time Zone Database for the lifetime of each agent.
-    </p>
-
-    <p>This section complements but does not supersede <emu-xref href="#sec-time-zone-identifiers"></emu-xref>.</p>
+    <ul>
+      <li>
+        For historical reasons, *"UTC"* must be a primary time zone identifier.
+        *"Etc/UTC"*, *"Etc/GMT"*, and *"GMT"*, as well as all Link names that resolve to any of them, must be non-primary time identifiers that resolve to *"UTC"*.
+      </li>
+      <li>
+        Any Link name that is present in the “TZ” column of file <code>zone.tab</code> must be a primary time zone identifier.
+        For example, both *"Europe/Prague"* and *"Europe/Bratislava"* must be primary time zone identifiers.
+        This requirement guarantees at least one primary time zone identifier for each <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code, and ensures that future changes to time zone rules of one country will not affect ECMAScript programs that use another country's time zone(s), unless those countries' territorial boundaries have also changed.
+      </li>
+      <li>
+        Any Link name that is not listed in the “TZ” column of file <code>zone.tab</code> and that represents a geographical area entirely contained within the territory of a single <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code must resolve to a primary identifier that also represents a geographical area entirely contained within the territory of the same country code.
+        For example, *"Atlantic/Jan_Mayen"* must resolve to *"Arctic/Longyearbyen"*.
+      </li>
+    </ul>
 
     <emu-note>
       <p>
         The IANA Time Zone Database offers build options that affect which time zone identifiers are primary.
-        The default build options merge different countries' time zones, for example *"Atlantic/Reykjavik"* being a Link to the Zone *"Africa/Abidjan"*.
+        The default build options merge different countries' time zones, for example *"Atlantic/Reykjavik"* is built as a Link to the Zone *"Africa/Abidjan"*.
         Geographically and politically distinct locations are likely to introduce divergent time zone rules in a future version of the IANA Time Zone Database.
-        Therefore, it is recommended that ECMAScript implementations instead use build options such as <code>PACKRATDATA=backzone PACKRATLIST=zone.tab</code> or a similar alternative that ensures at least one primary identifier for each <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code.
+        The exceptions above serve to mitigate these future-compatibility issues.
+      </p>
+      <p>
+        The Unicode Common Locale Data Repository (<a href="https://cldr.unicode.org">CLDR</a>) implements most of the exceptions above when determining which available named time zone identifiers are primary or non-primary.
+        Although use of CLDR data is recommended for consistency between implementations, it is not required.
+        Non-CLDR-based implementations can still use CLDR's identifier data in <a href="https://github.com/unicode-org/cldr/blob/main/common/bcp47/timezone.xml"><code>timezone.xml</code></a>.
+        Implementations may also build the IANA Time Zone Database directly, for example by using build options such as <code>PACKRATDATA=backzone PACKRATLIST=zone.tab</code> and performing any post-processing needed to ensure compliance with the requirements above.
       </p>
     </emu-note>
+
+    <p>
+      The IANA Time Zone Database is typically updated between five and ten times per year.
+      These updates may add new Zone or Link names, may change Zones to Links, and may change the UTC offsets and transitions associated with any Zone.
+      Implementations are recommended to include updates to the IANA Time Zone Database as soon as possible.
+      Such prompt action ensures that ECMAScript programs can accurately perform time-zone-sensitive calculations and can use newly-added available named time zone identifiers supplied by external input or the host environment.
+    </p>
+    <p>
+      Although the IANA Time Zone Database maintainers strive for stability, in rare cases (averaging less than once per year) a Zone may be replaced by a new Zone.
+      For example, in 2022 "*Europe/Kiev*" was deprecated to a Link resolving to a new "*Europe/Kyiv*" Zone.
+      The deprecated Link is called a <dfn variants="renamed time zone identifiers">renamed time zone identifier</dfn> and the newly-added Zone is called a <dfn variants="replacement time zone identifiers">replacement time zone identifier</dfn>.
+    </p>
+    <p>
+      To reduce disruption from these infrequent changes, implementations should initially add each replacement time zone identifier as a non-primary time zone identifier that resolves to the existing renamed time zone identifier.
+      This allows ECMAScript programs to recognize both identifiers, but also reduces the chance that an ECMAScript program will send the replacement time zone identifier to another system that does not yet recognize it.
+      After a <dfn id="rename-waiting-period">rename waiting period</dfn>, implementations should promote the new Zone to a primary time zone identifier while simultaneously demoting the renamed time zone identifier to non-primary.
+      To provide ample time for other systems to be updated, the recommended rename waiting period is two years.
+      However, it does not need to be either exact or dynamic.
+      Instead, implementations should make the replacement time zone identifier primary after the waiting period as part of their normal release process for updating time zone data.
+    </p>
+    <p>
+      A waiting period should only apply when a new Zone is added to replace an existing Zone.
+      If an existing Zone and Link are swapped, then no renaming has happened and no waiting period is necessary.
+    </p>
+
+    <p>
+      If implementations revise time zone information during the lifetime of an agent, then it is required that the list of available named time zone identifiers, the primary time zone identifier associated with any available named time zone identifier, and the UTC offsets and transitions associated with any available named time zone identifier, be consistent with results previously observed by that agent.
+      Due to the complexity of supporting this requirement, it is recommended that implementations maintain a fully consistent copy of the IANA Time Zone Database for the lifetime of each agent.
+    </p>
+
+    <p>This section complements but does not supersede <emu-xref href="#sec-time-zone-identifiers"></emu-xref>.</p>
+    </emu-clause>
 
     <emu-clause id="sup-availablenamedtimezoneidentifiers" oldids="sec-availabletimezones" type="implementation-defined abstract operation">
       <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Records</h1>
@@ -219,20 +254,67 @@
       <emu-alg>
         1. Let _identifiers_ be a List containing the String value of each Zone or Link name in the IANA Time Zone Database.
         1. Assert: No element of _identifiers_ is an ASCII-case-insensitive match for any other element.
-        1. Assert: Every element of _identifiers_ identifies a Zone or Link name in the IANA Time Zone Database.
         1. Sort _identifiers_ according to lexicographic code unit order.
         1. Let _result_ be a new empty List.
         1. For each element _identifier_ of _identifiers_, do
           1. Let _primary_ be _identifier_.
-          1. If _identifier_ is a Link name and _identifier_ is not *"UTC"*, then
-            1. Set _primary_ to the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
-            1. NOTE: An implementation may need to resolve _identifier_ iteratively.
+          1. If _identifier_ is a Link name in the IANA Time Zone Database and _identifier_ is not present in the “TZ” column of <code>zone.tab</code> of the IANA Time Zone Database, then
+            1. Let _zone_ be the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
+            1. If _zone_ starts with *"Etc/"*, then
+              1. Set _primary_ to _zone_.
+            1. Else,
+              1. Let _identifierCountryCode_ be the <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code whose territory contains the geographical area corresponding to _identifier_.
+              1. Let _zoneCountryCode_ be the ISO 3166-1 Alpha-2 country code whose territory contains the geographical area corresponding to _zone_.
+              1. If _identifierCountryCode_ is _zoneCountryCode_, then
+                1. Set _primary_ to _zone_.
+              1. Else,
+                1. Let _countryCodeLineCount_ be the number of lines in file <code>zone.tab</code> of the IANA Time Zone Database where the “country-code” column is _identifierCountryCode_.
+                1. If _countryCodeLineCount_ is 1, then
+                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the “country-code” column is _identifierCountryCode_.
+                  1. Set _primary_ to the contents of the “TZ” column of _countryCodeLine_.
+                1. Else,
+                  1. Let _backzone_ be *undefined*.
+                  1. Let _backzoneLinkLines_ be the List of lines in the file <code>backzone</code> of the IANA Time Zone Database that start with either *"Link "* or *"#PACKRATLIST zone.tab Link "*.
+                  1. For each element _line_ of _backzoneLinkLines_, do
+                    1. Let _i_ be StringIndexOf(_line_, *"Link "*, 0).
+                    1. Set _line_ to the substring of _line_ from _i_ + 5.
+                    1. Let _backzoneAndLink_ be StringSplitToList(_line_, *" "*).
+                    1. Assert: _backzoneAndLink_ has at least two elements, and both _backzoneAndLink_[0] and _backzoneAndLink_[1] are available named time zone identifiers.
+                    1. If _backzoneAndLink_[1] is _identifier_, then
+                      1. Assert: _backzone_ is *undefined*.
+                      1. Set _backzone_ to _backzoneAndLink_[0].
+                  1. Assert: _backzone_ is not *undefined*.
+                  1. Set _primary_ to _backzone_.
           1. If _primary_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, set _primary_ to *"UTC"*.
+          1. If _primary_ is a replacement time zone identifier and its rename waiting period has not concluded, then
+            1. Let _renamedIdentifier_ be the renamed time zone identifier that _primary_ replaced.
+            1. Set _primary_ to _renamedIdentifier_.
           1. Let _record_ be the Time Zone Identifier Record { [[Identifier]]: _identifier_, [[PrimaryIdentifier]]: _primary_ }.
           1. Append _record_ to _result_.
         1. Assert: _result_ contains a Time Zone Identifier Record _r_ such that _r_.[[Identifier]] is *"UTC"* and _r_.[[PrimaryIdentifier]] is *"UTC"*.
         1. Return _result_.
       </emu-alg>
+
+      <emu-note>
+        <p>
+          The algorithm above for resolving Links to primary time zone identifiers is intended to correspond to the behaviour of <code>icu::TimeZone::getIanaID()</code> in the International Components for Unicode (<a href="https://icu.unicode.org/">ICU</a>) and the processes for maintaining time zone identifier data in the Unicode Common Locale Data Repository (<a href="https://cldr.unicode.org">CLDR</a>).
+        </p>
+        <p>
+          This algorithm resolves Links to primary time zone identifiers without crossing the boundaries of ISO 3166-1 Alpha-2 country codes, using data from files <code>zone.tab</code> and <code>backzone</code> of the IANA Time Zone Database.
+          If the country code of a Link has only one line in <code>zone.tab</code>, then that line will determine the primary time zone identifier.
+          However, if that country code has multiple lines in <code>zone.tab</code>, then historical mappings in <code>backzone</code> must be used to identify the correct primary time zone identifier.
+        </p>
+        <p>
+          For example, to resolve "Pacific/Truk" (in country code "FM") if the default build options of the IANA Time Zone Database identify it as a Link to "Pacific/Port_Moresby" (in country code "PG"), then the “country-code” column of <code>zone.tab</code> must be checked for lines corresponding with "FM".
+          If there were only one such line, then the “TZ” column of that line would determine the primary time zone identifier associated with "Pacific/Truk".
+          But if there are multiple "FM" lines in <code>zone.tab</code>, then <code>backzone</code> must be inspected and a line such as "Link Pacific/Chuuk Pacific/Truk" would result in using "Pacific/Chuuk" as the primary time zone identifier.
+        </p>
+        <p>
+          Note that <code>zone.tab</code> is the preferred source of mapping data because <code>backzone</code> mappings may, in rare cases, cross the boundaries of ISO 3166-1 Alpha-2 country codes.
+          For example, "Atlantic/Jan_Mayen" (in country code "SJ") is mapped in <code>backzone</code> to "Europe/Oslo" (in country code "NO").
+          As of the 2024a release of the IANA Time Zone Database, "Atlantic/Jan_Mayen" is the only case where this happens.
+        </p>
+      </emu-note>
 
       <emu-note>
         Time zone identifiers in the IANA Time Zone Database can change over time.
@@ -260,9 +342,9 @@
         1. Return ~empty~.
       </emu-alg>
       <emu-note>
-        For any _timeZoneIdentifier_, or any value that is an ASCII-case-insensitive match for it, it is recommended that the resulting Time Zone Identifier Record contain the same field values for the lifetime of the surrounding agent.
-        Furthermore, it is recommended that time zone identifiers not dynamically change from primary to non-primary during the lifetime of the surrounding agent, meaning that if _timeZoneIdentifier_ is an ASCII-case-insensitive match for the [[PrimaryIdentifier]] field of the result of a previous call to GetAvailableNamedTimeZoneIdentifier, then GetAvailableNamedTimeZoneIdentifier(_timeZoneIdentifier_) must return a Record where [[Identifier]] is [[PrimaryIdentifier]].
-        Due to the complexity of supporting these recommendations, it is recommended that the result of AvailableNamedTimeZoneIdentifiers (and therefore GetAvailableNamedTimeZoneIdentifier too) remains the same for the lifetime of the surrounding agent.
+        For any _timeZoneIdentifier_, or any value that is an ASCII-case-insensitive match for it, it is required that the resulting Time Zone Identifier Record contain the same field values for the lifetime of the surrounding agent.
+        Furthermore, it is required that time zone identifiers not dynamically change from primary to non-primary during the lifetime of the surrounding agent, meaning that if _timeZoneIdentifier_ is an ASCII-case-insensitive match for the [[PrimaryIdentifier]] field of the result of a previous call to GetAvailableNamedTimeZoneIdentifier, then GetAvailableNamedTimeZoneIdentifier(_timeZoneIdentifier_) must return a Time Zone Identifier Record where [[Identifier]] is [[PrimaryIdentifier]].
+        Due to the complexity of supporting these requirements, it is recommended that the result of AvailableNamedTimeZoneIdentifiers (and therefore GetAvailableNamedTimeZoneIdentifier too) remains the same for the lifetime of the surrounding agent.
       </emu-note>
     </emu-clause>
 
@@ -285,6 +367,38 @@
         1. Return _result_.
       </emu-alg>
     </emu-clause>
+
+      <emu-clause id="sec-string-split-to-list" type="abstract operation">
+        <h1>
+          StringSplitToList (
+            _S_: a String,
+            _separator_: a String,
+          ): a List of Strings
+        </h1>
+        <dl class="header">
+          <dt>description</dt>
+          <dd>
+            The returned List contains all disjoint substrings of _S_ that do not contain _separator_ but are immediately preceded and/or immediately followed by an occurrence of _separator_.
+            Each such <emu-not-ref>substring</emu-not-ref> will be the empty String between adjacent occurrences of _separator_, before a _separator_ at the very start of _S_, or after a _separator_ at the very end of _S_, but otherwise will not be empty.
+          </dd>
+        </dl>
+        <emu-alg>
+          1. Assert: _S_ is not the empty String.
+          1. Assert: _separator_ is not the empty String.
+          1. Let _separatorLength_ be the length of _separator_.
+          1. Let _substrings_ be a new empty List.
+          1. Let _i_ be 0.
+          1. Let _j_ be StringIndexOf(_S_, _separator_, 0).
+          1. Repeat, while _j_ is not ~not-found~,
+            1. Let _T_ be the substring of _S_ from _i_ to _j_.
+            1. Append _T_ to _substrings_.
+            1. Set _i_ to _j_ + _separatorLength_.
+            1. Set _j_ to StringIndexOf(_S_, _separator_, _i_).
+          1. Let _T_ be the substring of _S_ from _i_.
+          1. Append _T_ to _substrings_.
+          1. Return _substrings_.
+        </emu-alg>
+      </emu-clause>
   </emu-clause>
 
   <emu-clause id="sec-measurement-unit-identifiers">


spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,228 @@ <h1>
...
+            1. Set _primary_ to the value in the second column of the row where the first column is _identifier_.
+          1. Else,
+            1. NOTE: The algorithm steps below are intended to mimic the behaviour of the *icu::TimeZone::getIanaID()* API of the International Components for Unicode (ICU) <a href="https://icu.unicode.org/">https://icu.unicode.org/</a>.
+               The steps in this section are provided for testing conformance of ICU-based ECMAScript implementations, and to ensure compatibility between ICU-based and non-ICU-based implementations.
+            1. If _identifier_ is present in the *TZ* column of *zone.tab* of the IANA Time Zone Database, then
+              1. Set _primary_ to _identifier_.
+            1. Else if _identifier_ is a Link in the IANA Time Zone Database, then
+              1. Let _zone_ be the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
+              1. If _zone_ starts with *"Etc/"*, then
```
justingrant, line 333: @gibson042 is "starts with" and "ends with" the correct spec-ese, and if not what should I use?

sffc:
Thanks for putting this together @justingrant! We can discuss this at the next TG2 meeting. In the mean time, I encourage the listed reviewers to take a look.

ptomato:
> Test262 changes will be needed to validate these normative changes, but I'm not sure how we can run those tests except using the Temporal polyfill.

It's fine to submit tests to test262 that no engine can pass yet, as long as they are correct according to the current snapshot of ECMA-262 or 402 or a Stage 3 proposal.

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,228 @@ <h1>
...
+                1. Set _primary_ to _zone_.
+              1. Else,
+                1. Let _identifierCountyCode_ be the ISO 3166-2 country code <a href="https://www.iso.org/iso-3166-country-codes.html">https://www.iso.org/iso-3166-country-codes.html</a> whose territory contains the geographical area corresponding to _identifier_.
+                1. Let _zoneCountyCode_ be the ISO 3166-2 country code whose territory contains the geographical area corresponding to _zone_.
+                1. If _identifierCountyCode_ is _zoneCountyCode_, then
```
ptomato, line 338:
```suggestion
                1. Let _identifierCountryCode_ be the ISO 3166-2 country code <a href="https://www.iso.org/iso-3166-country-codes.html">https://www.iso.org/iso-3166-country-codes.html</a> whose territory contains the geographical area corresponding to _identifier_.
                1. Let _zoneCountryCode_ be the ISO 3166-2 country code whose territory contains the geographical area corresponding to _zone_.
                1. If _identifierCountryCode_ is _zoneCountryCode_, then
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,228 @@ <h1>
...
+                  1. Set _primary_ to _zone_.
+                1. Else,
+                  1. Let _countryCodeLine_ be the line in *zone.tab* of the IANA Time Zone Database where the *country-code* column is _identifierCountyCode_.
```
ptomato, line 341:
```suggestion
                  1. Let _countryCodeLine_ be the line in *zone.tab* of the IANA Time Zone Database where the *country-code* column is _identifierCountryCode_.
```

ptomato (commented):
I don't feel qualified to review the details — I mostly haven't followed those discussions. At a general level, this all looks very reasonable.

spec/locales-currencies-tz.html
justingrant: Lol, I had to look for several minutes before I could even spot the change. No wonder I missed the typo the first time. Thanks for catching this, fixed now.

spec/locales-currencies-tz.html
justingrant: FWIW, it looks like "starts with" is something that ECMA-262 already uses: https://tc39.es/ecma262/#sec-getsubstitution. So I'll assume that it's OK here unless I hear otherwise.

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+
+    <emu-note>
+      <p>
+        The IANA Time Zone Database offers build options that affect which time zone identifiers are primary.
+        The default build options merge different countries' time zones, for example *"Atlantic/Reykjavik"* is built as a Link to the Zone *"Africa/Abidjan"*.
+        Geographically and politically distinct locations are likely to introduce divergent time zone rules in a future version of the IANA Time Zone Database.
+        The exceptions above serve to mitigate these future-compatibility issues for ECMAScript programmers and end users.
+      </p>
+      <p>
+        International Components for Unicode (ICU) <a href="https://icu.unicode.org/">https://icu.unicode.org/</a> is a widely-used library that exposes IANA Time Zone Database information.
```
srl295, line 263: Refer to cldr first here. That xml file is not "ICU's". 

Then afterwards say for example ICU is widely used etc.

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+    </p>
+    <p>
+      To reduce disruption from these infrequent changes, ECMAScript implementations must initially add each replacement time zone identifier as a non-primary time zone identifier that resolves to the existing renamed time zone identifier.
+      This allows ECMAScript programs to recognize both identifiers, but also reduces the chance that an ECMAScript program will send the replacement time zone identifier to another system that does not yet recognize it.
+      After a waiting period, implementations must promote the new Zone to a primary time zone identifier while simultaneously demoting the renamed time zone identifier to non-primary.
+      This waiting period is two years after the IANA Time Zone Database release containing the changes, to provide ample time for other systems to be updated.
+      After two years, implementations should update their time zone data to make the replacement time zone identifier primary and the renamed time zone identifier non-primary.
+      This two-year period does not need to be exact.
+      For example, it's acceptable to wait until the next ICU release after two years has expired.
```
srl295, line 289: Cldr release?

justingrant: Are they released separately?  I thought (but if I'm mistaken let me know!) that each ICU release bundled a specific version of CLDR, so from the point of view of an ECMAScript engine, it's the ICU release that they will include in their builds and CLDR data comes along with that. But I'm admittedly not familiar with the release process so I might have gotten it wrong.

Would "ICU and CLDR release" be better here?

justingrant: @srl295 Do you know answers to the questions above? Thanks!

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+
+    <emu-table id="table-posix-time-zone-identifier-mapping">
+      <emu-caption><emu-not-ref>Legacy</emu-not-ref> POSIX Time Zone Identifier Mapping</emu-caption>
+      <table class="real-table">
+        <thead style="white-space: nowrap">
+          <tr>
+            <th><emu-not-ref>Legacy</emu-not-ref> POSIX Zone Name</th>
+            <th>Primary Time Zone Identifier</th>
+          </tr>
+        </thead>
```
gibson042, line 207: I don't think this use of "POSIX" is correct; https://data.iana.org/time-zones/theory.html#naming just refers to "legacy names".
```suggestion
        "<emu-not-ref>Legacy names</emu-not-ref>" must be non-primary time zone identifiers that resolve to their closest non-legacy equivalents, as shown in <emu-xref href="#table-legacy-time-zone-identifier-mapping"></emu-xref> below:
      </li>
    </ul>

    <emu-table id="table-legacy-time-zone-identifier-mapping">
      <emu-caption><emu-not-ref>Legacy</emu-not-ref> Time Zone Identifier Mapping</emu-caption>
      <table class="real-table">
        <thead style="white-space: nowrap">
          <tr>
            <th><emu-not-ref>Legacy Name</emu-not-ref></th>
            <th>Primary Time Zone Identifier</th>
          </tr>
        </thead>
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+        The default build options merge different countries' time zones, for example *"Atlantic/Reykjavik"* is built as a Link to the Zone *"Africa/Abidjan"*.
+        Geographically and politically distinct locations are likely to introduce divergent time zone rules in a future version of the IANA Time Zone Database.
+        The exceptions above serve to mitigate these future-compatibility issues for ECMAScript programmers and end users.
+      </p>
+      <p>
+        International Components for Unicode (ICU) <a href="https://icu.unicode.org/">https://icu.unicode.org/</a> is a widely-used library that exposes IANA Time Zone Database information.
+        ICU implements most of the exceptions above when determining which available named time zone identifiers are primary or non-primary.
+        Although use of ICU is recommended for consistency between implementations, it is not required.
+        Non-ICU-based implementations can still use ICU's identifier data, as found in *timezone.xml* in the Unicode Common Locale Data Repository (CLDR) <a href="https://cldr.unicode.org/">https://cldr.unicode.org/</a>.
```
gibson042, line 266:
```suggestion
        Non-ICU-based implementations can still use ICU's identifier data, as found in file <a href="https://github.com/unicode-org/cldr/blob/main/common/bcp47/timezone.xml"><code>timezone.xml</code></a> of the Unicode Common Locale Data Repository (<a href="https://cldr.unicode.org">CLDR</a>).
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+        Implementations may also build the IANA Time Zone Database directly, for example by using build options such as <code>PACKRATDATA=backzone PACKRATLIST=zone.tab</code> and performing any post-processing needed to ensure compliance with the requirements above.
+      </p>
+    </emu-note>
 
     <p>
-      A conforming implementation must recognize *"UTC"* and all other Zone and Link names (and <strong>only</strong> such names), and use best available current and historical information about their offsets from UTC and their daylight saving time rules in calculations. However, the set of combinations of time zone name and language tag for which localized time zone names are available is implementation dependent.
+      The IANA Time Zone Database is typically updated between five and ten times per year.
+      These updates may add new Zone or Link names, may change Zones to Links, and may change the UTC offsets and transitions associated with any Zone.
+      ECMAScript implementations are recommended to include updates to the IANA Time Zone Database as soon as possible.
```
gibson042, line 274:
```suggestion
      Implementations are recommended to include updates to the IANA Time Zone Database as soon as possible.
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+      Such prompt action ensures that ECMAScript programs can accurately perform time-zone-sensitive calculations and can use newly-added available named time zone identifiers supplied by external input or the host environment.
+    </p>
+    <p>
+      Although the IANA Time Zone Database maintainers strive for stability, in rare cases (averaging less than once per year) a Zone may be replaced by a new Zone.
+      For example, in 2022 "*Europe/Kiev*" was deprecated to a Link resolving to a new "*Europe/Kyiv*" Zone.
+      The deprecated Link is called a <dfn variants="renamed time zone identifiers">renamed time zone identifier</dfn> and the newly-added Zone is called a <dfn variants="replacement time zone identifiers">replacement time zone identifier</dfn>.
+    </p>
+    <p>
+      To reduce disruption from these infrequent changes, ECMAScript implementations must initially add each replacement time zone identifier as a non-primary time zone identifier that resolves to the existing renamed time zone identifier.
```
gibson042, line 283: I don't feel comfortable with this guidance being a "must"; implementations may have arbitrarily large gaps between releases.
```suggestion
      To reduce disruption from these infrequent changes, implementations should initially add each replacement time zone identifier as a non-primary time zone identifier that resolves to the existing renamed time zone identifier.
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
-        1. If _timeZone_ is an ASCII-case-insensitive match for *"UTC"*, return *true*.
-        1. Return *false*.
+        1. Let _identifiers_ be a List containing the String value of each Zone or Link name in the IANA Time Zone Database.
+        1. Assert: No element of _identifiers_ is an ASCII-case-insensitive match for any other element.
+        1. Assert: Every element of _identifiers_ identifies a Zone or Link name in the IANA Time Zone Database.
+        1. Sort _identifiers_ according to lexicographic code unit order.
+        1. Let _result_ be a new empty List.
+        1. For each element _identifier_ of _identifiers_, do
+          1. Let _primary_ be _identifier_.
+          1. If _identifier_ is listed in the first column of <emu-not-ref>Legacy</emu-not-ref> POSIX Time Zone Identifier Mapping <emu-xref href="#table-posix-time-zone-identifier-mapping"></emu-xref>, then
```
gibson042, line 324:
```suggestion
          1. If _identifier_ is listed in the first column of <emu-xref href="#table-legacy-time-zone-identifier-mapping"></emu-xref>, then
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+            1. Set _primary_ to the value in the second column of the row where the first column is _identifier_.
+          1. Else,
+            1. NOTE: The algorithm steps below are intended to mimic the behaviour of *icu::TimeZone::getIanaID()* in the International Components for Unicode (ICU) <a href="https://icu.unicode.org/">https://icu.unicode.org/</a>.
```
gibson042, line 327:
```suggestion
            1. NOTE: The algorithm steps below are intended to mimic the behaviour of <code>icu::TimeZone::getIanaID()</code> in the International Components for Unicode (<a href="https://icu.unicode.org/">ICU</a>).
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+               The steps in this section are provided for testing conformance of ICU-based ECMAScript implementations, and to ensure compatibility between ICU-based and non-ICU-based implementations.
+            1. If _identifier_ is present in the *TZ* column of *zone.tab* of the IANA Time Zone Database, then
```
gibson042, line 329:
```suggestion
            1. If _identifier_ is present in the “TZ” column of file <code>zone.tab</code> of the IANA Time Zone Database, then
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+              1. If _zone_ starts with *"Etc/"*, then
+                1. Set _primary_ to _zone_.
+              1. Else,
+                1. Let _identifierCountryCode_ be the ISO 3166-2 country code <a href="https://www.iso.org/iso-3166-country-codes.html">https://www.iso.org/iso-3166-country-codes.html</a> whose territory contains the geographical area corresponding to _identifier_.
+                1. Let _zoneCountryCode_ be the ISO 3166-2 country code whose territory contains the geographical area corresponding to _zone_.
+                1. If _identifierCountryCode_ is _zoneCountryCode_, then
+                  1. Set _primary_ to _zone_.
+                1. Else,
+                  1. Let _countryCodeLine_ be the line in *zone.tab* of the IANA Time Zone Database where the *country-code* column is _identifierCountryCode_.
+                  1. Set _primary_ to the contents of the *TZ* column of _countryCodeLine_.
```
gibson042, line 342:
```suggestion
                1. Let _identifierCountryCode_ be the <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code whose territory contains the geographical area corresponding to _identifier_.
                1. Let _zoneCountryCode_ be the ISO 3166-1 Alpha-2 country code whose territory contains the geographical area corresponding to _zone_.
                1. If _identifierCountryCode_ is _zoneCountryCode_, then
                  1. Set _primary_ to _zone_.
                1. Else,
                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the “country-code” column is _identifierCountryCode_.
                  1. Set _primary_ to the contents of the “TZ” column of _countryCodeLine_.
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+    <p>
+      Although the IANA Time Zone Database maintainers strive for stability, in rare cases (averaging less than once per year) a Zone may be replaced by a new Zone.
+      For example, in 2022 "*Europe/Kiev*" was deprecated to a Link resolving to a new "*Europe/Kyiv*" Zone.
+      The deprecated Link is called a <dfn variants="renamed time zone identifiers">renamed time zone identifier</dfn> and the newly-added Zone is called a <dfn variants="replacement time zone identifiers">replacement time zone identifier</dfn>.
+    </p>
+    <p>
+      To reduce disruption from these infrequent changes, ECMAScript implementations must initially add each replacement time zone identifier as a non-primary time zone identifier that resolves to the existing renamed time zone identifier.
+      This allows ECMAScript programs to recognize both identifiers, but also reduces the chance that an ECMAScript program will send the replacement time zone identifier to another system that does not yet recognize it.
+      After a waiting period, implementations must promote the new Zone to a primary time zone identifier while simultaneously demoting the renamed time zone identifier to non-primary.
```
gibson042, line 285: This should include a `<dfn>` so it can be linked to.
```suggestion
      After a <dfn id="rename-waiting-period">rename waiting period</dfn>, implementations must promote the new Zone to a primary time zone identifier while simultaneously demoting the renamed time zone identifier to non-primary.
```

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+                  1. Let _countryCodeLine_ be the line in *zone.tab* of the IANA Time Zone Database where the *country-code* column is _identifierCountryCode_.
+                  1. Set _primary_ to the contents of the *TZ* column of _countryCodeLine_.
+            1. If _primary_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, set _primary_ to *"UTC"*.
+            1. If _primary_ is a replacement time zone identifier, and it has been less than two years since the release of the IANA Time Zone Database that added _primary_ as a new Zone name, then
+              1. Let _renamedIdentifier_ be the renamed time zone identifier that _primary_ replaced.
+              1. Set _primary_ to _renamedIdentifier_.
+              1. NOTE: This two year waiting period does not need to be exact, and is not required to be applied dynamically, especially in implementations that do not update time zone data between releases.
+                Instead, implementations may make the replacement time zone identifier primary as part of their normal release process for updating time zone data, and to release the change as close as practical to the end of the two year waiting period.
```
gibson042, line 348:
```suggestion
            1. If _primary_ is a replacement time zone identifier, and the corresponding rename waiting period since the release of the IANA Time Zone Database that added _primary_ as a new Zone name has not concluded, then
              1. Let _renamedIdentifier_ be the renamed time zone identifier that _primary_ replaced.
              1. Set _primary_ to _renamedIdentifier_.
```

gibson042 (changes requested):
Mostly editorial comments, but I am not comfortable with making the rename waiting period mandatory.

FrankYFTang:
> 4\. e.g. PST7PDT=>America/Los_Angeles

I assuem you mean to say "e.g. PST8PDT=>America/Los_Angeles" ?
not "PST7PDT=>America/Los_Angeles" right?

FrankYFTang:
I am not sure what these CLDR tool files for but it mention
https://github.com/unicode-org/cldr/blob/a6fbd0ffa1db0c9f60dc0f678ace589366ae7f3a/tools/cldr-code/src/main/resources/org/unicode/cldr/icu/timezone_aliases.txt#L100 
https://github.com/unicode-org/cldr/blob/a6fbd0ffa1db0c9f60dc0f678ace589366ae7f3a/tools/cldr-code/src/main/resources/org/unicode/cldr/util/data/TimeZoneAliases.txt#L86

spec/locales-currencies-tz.html
```diff
@@ -164,59 +164,229 @@ <h1>
...
+    </p>
+    <ul>
+      <li>
+        For historical reasons, *"UTC"* must be a primary time zone identifier.
+        *"Etc/UTC"*, *"Etc/GMT"*, and *"GMT"*, as well as all Link names that resolve to any of them, must be non-primary time identifiers that resolve to "UTC".
+      </li>
+      <li>
+        Any Link name in the TZ column of *zone.tab* of the IANA Time Zone Database must be a primary time zone identifier.
+        For example, both *"Europe/Prague"* and *"Europe/Bratislava"* must be primary time zone identifiers.
+        This requirement guarantees at least one primary time zone identifier for each ISO 3166-2 country code <a href="https://www.iso.org/iso-3166-country-codes.html">https://www.iso.org/iso-3166-country-codes.html</a>.
```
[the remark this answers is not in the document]
justingrant: Whoops, somehow I confused ISO 3166-1 Alpha-2 (which is the correct answer) with ISO 3166-2. Thanks for catching this!  Also, your suggestions below are great, thanks so much! I'll apply them soon and rebase.

spec/locales-currencies-tz.html
justingrant: "Should is fine with me, I'll make this change in the next push.

spec/locales-currencies-tz.html
anba: There are three ICU projects (ICU4C, ICU4J, and ICU4X). CLDR and ICU4C/ICU4J releases are sync'ed, but ICU4X has a different release cycle. ECMAScript implementations use all three ICU projects. For example ICU4C is used by JSC, V8, and SpiderMonkey; ICU4X is used by SpiderMonkey for `Intl.Segmenter`; ICU4J is used by GraalJS. 

So to avoid confusion which ICU project is meant, referring to CLDR is preferable.

ryzokuken: Not only that, the spec explicitly avoids referring to specific ICU/non-ICU implementations in favor of referring to CLDR.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-04-25.md#normative-specify-time-zone-id-requirements-to-reduce-divergence-between-engines-877

spec/locales-currencies-tz.html
justingrant: OK, I'll replace ICU with CLDR here, and elsewhere.

spec/locales-currencies-tz.html
justingrant: Per other comments I ended up just focusing on CLDR here, and removing the reference to ICU because the data in question is wholly contained in CLDR.

justingrant:
I just pushed a new commit that includes what I think resolves all review feedback. @gibson042 @sffc (and anyone else who's interested) do you want to re-review?

spec/locales-currencies-tz.html
```diff
@@ -224,10 +319,29 @@ <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Record
...
-            1. Set _primary_ to the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
-            1. NOTE: An implementation may need to resolve _identifier_ iteratively.
-          1. If _primary_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, set _primary_ to *"UTC"*.
+          1. If _identifier_ is listed in the first column of <emu-xref href="#table-legacy-time-zone-identifier-mapping"></emu-xref>, then
+            1. Set _primary_ to the value in the second column of the row where the first column is _identifier_.
+          1. Else,
+            1. NOTE: The algorithm steps below are intended to mimic the behaviour of <code>icu::TimeZone::getIanaID()</code> in the International Components for Unicode (<a href="https://icu.unicode.org/">ICU</a>).
+               The steps in this section are provided for testing conformance of ICU-based ECMAScript implementations, and to ensure compatibility between ICU-based and non-ICU-based implementations.
+            1. If _identifier_ is present in the "TZ" column of *zone.tab* of the IANA Time Zone Database, then
```
gibson042, line 327:
```suggestion
            1. If _identifier_ is present in the “TZ” column of file <code>zone.tab</code> of the IANA Time Zone Database, then
```

spec/locales-currencies-tz.html
```diff
@@ -224,10 +319,29 @@ <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Record
...
+              1. If _zone_ starts with *"Etc/"*, then
+                1. Set _primary_ to _zone_.
+              1. Else,
+                1. Let _identifierCountryCode_ be the <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code whose territory contains the geographical area corresponding to _identifier_.
+                1. Let _zoneCountryCode_ be the ISO 3166-1 Alpha-2 country code whose territory contains the geographical area corresponding to _zone_.
+                1. If _identifierCountryCode_ is _zoneCountryCode_, then
+                  1. Set _primary_ to _zone_.
+                1. Else,
+                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the "country-code" column is _identifierCountryCode_.
+                  1. Set _primary_ to the contents of the "TZ" column of _countryCodeLine_.
```
gibson042, line 340:
```suggestion
                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the “country-code” column is _identifierCountryCode_.
                  1. Set _primary_ to the contents of the “TZ” column of _countryCodeLine_.
```

spec/locales-currencies-tz.html
```diff
@@ -224,10 +319,29 @@ <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Record
...
+              1. Let _zone_ be the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
+              1. If _zone_ starts with *"Etc/"*, then
+                1. Set _primary_ to _zone_.
+              1. Else,
+                1. Let _identifierCountryCode_ be the <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code whose territory contains the geographical area corresponding to _identifier_.
+                1. Let _zoneCountryCode_ be the ISO 3166-1 Alpha-2 country code whose territory contains the geographical area corresponding to _zone_.
+                1. If _identifierCountryCode_ is _zoneCountryCode_, then
+                  1. Set _primary_ to _zone_.
+                1. Else,
+                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the "country-code" column is _identifierCountryCode_.
```
gibson042, line 339: As of right now, 31 codes have more than one line in zone.tab:

<details>

```console
$ curl -sSL 'https://github.com/eggert/tz/raw/main/zone.tab' \
  | awk '{ if (match($1, "^[A-Z][A-Z]$")) print $1; }' \
  | sort \
  | uniq -c \
  | awk '$1 > 1 { print; i++; } END { print i }'
     10 AQ
     12 AR
     12 AU
     16 BR
     23 CA
      2 CD
      3 CL
      2 CN
      2 CY
      2 DE
      2 EC
      3 ES
      3 FM
      4 GL
      4 ID
      3 KI
      7 KZ
      2 MH
      3 MN
     12 MX
      2 MY
      2 NZ
      3 PF
      2 PG
      2 PS
      3 PT
     26 RU
      2 UA
      2 UM
     29 US
      2 UZ
31
```

</details>

For example, America/Blanc-Sablon (CA) is a Link to America/Puerto_Rico (PR), but CA has a total of 23 lines in <tt>zone.tab</tt> (including one that is specifically for America/Blanc-Sablon), and similar arrangements affect regions as populous as Africa/Kinshasa and Asia/Kuala_Lumpur. So this algorithm is underspecified. What is the actual intent?

spec/locales-currencies-tz.html
gibson042, line 340: This whole block can be simplified a bit.
```suggestion
              1. Set _primary_ to the Zone name that _identifier_ resolves to, according to the rules for resolving Link names in the IANA Time Zone Database.
              1. If _primary_ does not start with *"Etc/"*, then
                1. Let _identifierCountryCode_ be the <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country code whose territory contains the geographical area corresponding to _identifier_.
                1. Let _zoneCountryCode_ be the ISO 3166-1 Alpha-2 country code whose territory contains the geographical area corresponding to _primary_.
                1. If _identifierCountryCode_ is not _zoneCountryCode_, then
                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the “country-code” column is _identifierCountryCode_.
                  1. Set _primary_ to the contents of the “TZ” column of _countryCodeLine_.
```

gibson042 (changes requested):
10 editorial suggestions and one request for better specification.

spec/locales-currencies-tz.html
justingrant: Specs are required to use smart quote characters?

spec/locales-currencies-tz.html
justingrant: Ahh, shoot. Yeah, we'll need a better solution, thanks for catching this. 

The problem to solve here is multi-hop Links like A => B => C where A and B are in the the same country but C is in a different country. This happens when an earlier TZDB release created the A => B Link, then a later release folded B into C, so the link that used to be A => B now was changed to A => C (along with a new Link for B => C).

There are 15 cases like this, conveniently separated into three cases of 5 each. 

**Cases where A is in zone.tab**
- America/Kralendijk ⇨ America/Curacao  ⇨ America/Puerto_Rico
- America/Lower_Princes ⇨ America/Curacao ⇨ America/Puerto_Rico
- America/Marigot ⇨ America/Port_of_Spain ⇨ America/Puerto_Rico
- America/St_Barthelemy ⇨ America/Port_of_Spain  ⇨ America/Puerto_Rico
- Arctic/Longyearbyen ⇨ Europe/Oslo ⇨ Europe/Berlin

**Cases where A's country code has only one line in zone.tab**
- America/Virgin ⇨ America/St_Thomas  ⇨ America/Puerto_Rico
- Atlantic/Jan_Mayen ⇨ Europe/Oslo ⇨ Europe/Berlin
- Iceland ⇨ Atlantic/Reykjavik ⇨ Africa/Abidjan
- Africa/Asmera ⇨ Africa/Asmara ⇨ Africa/Nairobi
- Africa/Timbuktu ⇨ Africa/Bamako ⇨ Africa/Abidjan

**Cases where A's country code has multiple lines in zone.tab (broken case in current PR's spec text)**
- (FM) Pacific/Ponape ⇨ Pacific/Pohnpei ⇨ Pacific/Guadalcanal
- (FM) Pacific/Truk ⇨ Pacific/Chuuk ⇨ Pacific/Port_Moresby
- (FM) Pacific/Yap ⇨ Pacific/Chuuk ⇨ Pacific/Port_Moresby
- (CA) America/Coral_Harbour ⇨ America/Atikokan ⇨ America/Panama
- (AQ) Antarctica/South_Pole ⇨ Antarctica/McMurdo ⇨ Pacific/Auckland 

For that last case, the only way to know what B was is to look in the `backzone` file in TZDB. The old Links are listed there. 

I'll draft some spec text to handle that case, and will let you know once I push it because I'm sure it will require a round or two of polishing. I suspect that the cleanest solution will be to just to use `backzone` for both the second and third cases above, and remove the complex spec text to handle the second case using zone.tab.

Thanks again for catching this.

justingrant: @gibson042 I just pushed a solution to the problem you found, as well as fixing (I think) all the other review feedback you had except the quote characters. Want to review the latest?

For the cases above, it turned out that there wasn't a single algorithmic solution possible, so I used the existing PR's algorithm for Case 2 above, and used `backzone` for Case 3. 

Sadly, we can't use `backzone` for both cases 2 and 3 because a single A => B => C chain in `backzone` requires the zone.tab algorithm. The problematic identifier is Atlantic/Jan_Mayen. This ID is linked to Europe/Oslo (country code "NO") in `backzone`, but it should be linked to Arctic/Longyearbyen which has the same country code ("SJ") as Atlantic/Jan_Mayen.

Let me know what you think of the latest commit.

spec/locales-currencies-tz.html
gibson042: No, but it's the prevailing convention and I see no reason to deviate.

spec/locales-currencies-tz.html
```diff
@@ -219,15 +254,34 @@ <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Record
...
+                1. If _countryCodeLineCount_ is 1, then
+                  1. Let _countryCodeLine_ be the line in file <code>zone.tab</code> of the IANA Time Zone Database where the "country-code" column is _identifierCountryCode_.
+                  1. Set _primary_ to the contents of the "TZ" column of _countryCodeLine_.
+                1. Else,
+                  1. Let _backzoneLinkLines_ be the subset of lines in the file <code>backzone</code> of the IANA Time Zone Database that start with either *"Link"* or *"#PACKRATLIST zone.tab Link"*.
+                  1. Assert: Exactly one line in _backzoneLinkLines_ contains _identifier_ in the second, space-delimited column after *"Link"* or *"#PACKRATLIST zone.tab Link"*.
+                  1. Let _backzoneLinkLine_ be the line in _backzoneLinkLines_ that contains _identifier_ in the second, space-delimited column after *"Link"* or *"#PACKRATLIST zone.tab Link"*.
+                  1. Set _primary_ to the contents of the first, space-delimited column in _backzoneLinkLine_ after *"Link"* or *"#PACKRATLIST zone.tab Link"*.
```
gibson042, line 280: I've wanted a StringSplit operation for a while ([IsWellFormedUnitIdentifier](https://tc39.es/ecma402/#sec-iswellformedunitidentifier) is practically begging for it), but in the absence of such I'd like to be crystal-clear in this kind of algorithm.
```suggestion
                  1. Let _backzone_ be *undefined*.
                  1. Let _backzoneLinkLines_ be the List of lines in the file <code>backzone</code> of the IANA Time Zone Database that start with either *"Link "* or *"#PACKRATLIST zone.tab Link "*.
                  1. For each element _line_ of _backzoneLinkLines_, do
                    1. If _line_ starts with *"#PACKRATLIST zone.tab "*, set _line_ to the substring of _line_ from 22.
                    1. Assert: _line_ starts with *"Link "*.
                    1. Let _i_ be 5.
                    1. Let _j_ be StringIndexOf(_line_, *" "*, _i_).
                    1. Assert: _j_ is not ~not-found~ and _j_ > _i_.
                    1. Let _k_ be StringIndexOf(_line_, *" "*, _j_ + 1).
                    1. If _k_ is ~not-found~, set _k_ to the length of _line_.
                    1. Assert: _k_ > _j_ + 1.
                    1. Let _alias_ be the substring of _line_ from _j_ + 1 to _k_.
                    1. If _alias_ is _identifier_, then
                      1. Assert: _backzone_ is *undefined*.
                      1. Set _backzone_ to the substring of _line_ from _i_ to _j_.
                  1. Assert: _backzone_ is not *undefined*.
                  1. Set _primary_ to _backzone_.
```
A worked example of e.g. Pacific/Truk against current tzdata in an editorial note would also be helpful.

gibson042 (approved):
The algorithm looks good, but I'd like to use more precise phrasing than «the first, space-delimited column in _backzoneLinkLine_ after…».

spec/locales-currencies-tz.html
justingrant: Oh wow that's hard to read. I wonder if it'd be better to introduce a string split AO into 402 in this PR and use it in this PR, then it could be re-used elsewhere in 402 later?

justingrant: Take a look at the commit I just pushed that includes a minimal string-split AO, borrowed from the lower half of the spec text of [`String.prototype.split`](https://tc39.es/ecma262/#sec-string.prototype.split). 

> A worked example of e.g. Pacific/Truk against current tzdata in an editorial note would also be helpful.

When you say "worked example" what do you mean?

spec/locales-currencies-tz.html
justingrant: Want to take a look at the latest commit and let me know if you think that it's simplified enough?

spec/locales-currencies-tz.html
```diff
@@ -219,15 +254,42 @@ <h1>AvailableNamedTimeZoneIdentifiers ( ): a List of Time Zone Identifier Record
...
+                  1. For each element _line_ of _backzoneLinkLines_, do
+                    1. If _line_ starts with *"#PACKRATLIST zone.tab "*, set _line_ to the substring of _line_ from 22.
+                    1. Assert: _line_ starts with *"Link "*.
+                    1. Set _line_ to the substring of _line_ from 5.
+                    1. Let _backzoneAndLink_ be StringSplitToList(_line_, *" "*, 2).
+                    1. Assert: _backzoneAndLink_ has exactly two elements.
+                    1. If _backzoneAndLink_[1] is _identifier_, then
+                      1. Assert: _backzone_ is *undefined*.
+                      1. Set _primary_ to _backzoneAndLink_[0].
+                  1. Assert: _backzone_ is not *undefined*.
```
gibson042, line 288:
```suggestion
                  1. Let _backzone_ be *undefined*.
                  1. Let _backzoneLinkLines_ be the List of lines in the file <code>backzone</code> of the IANA Time Zone Database that start with either *"Link "* or *"#PACKRATLIST zone.tab Link "*.
                  1. For each element _line_ of _backzoneLinkLines_, do
                    1. Let _i_ be StringIndexOf(_line_, *"Link"*, 0).
                    1. Set _line_ to the substring of _line_ from _i_.
                    1. Let _fields_ be StringSplitToList(_line_, *" "*).
                    1. Assert: _fields_ has at least three elements, _fields_[0] is *"Link"*, _fields_[1] is not the empty String, and _fields_[2] is not the empty String.
                    1. If _fields_[2] is _identifier_, then
                      1. Assert: _backzone_ is *undefined*.
                      1. Set _backzone_ to _fields_[1].
                  1. Assert: _backzone_ is not *undefined*.
                  1. Set _primary_ to _backzone_.
```

gibson042 (changes requested):
Looking good! Some final tweaks to StringSplitToList and its use, then I think this is ready.

spec/locales-currencies-tz.html
justingrant: > Take a look at the commit I just pushed that includes a minimal string-split AO, borrowed from the lower half of the spec text of [`String.prototype.split`](https://tc39.es/ecma262/#sec-string.prototype.split).

And while we're at it, I refactored `IsWellFormedUnitIdentifier` to use the new AO.

gibson042: > When you say "worked example" what do you mean?

A note explaining something like this:
> This algorithm attempts to resolve Links to primary time zone identifiers without crossing the boundaries of <a href="https://www.iso.org/glossary-for-iso-3166.html">ISO 3166-1 Alpha-2</a> country codes, using data from files <code>zone.tab</code> and <code>backzone</code> of the IANA Time Zone Database as necessary.
> For example, if *"Pacific/Truk"* (in country code *"FM"*) is a Link to *"Pacific/Port_Moresby"* (in country code *"PG"*) with default configuration, then <code>zone.tab</code> will be checked for lines corresponding with country code *"FM"*. If there is only one such line, then the Zone of that line will be treated as the primary time zone identifier associated with *"Pacific/Truk"*. But otherwise, that data will be taken from the unique Link in <code>backzone</code> with source name *"Pacific/Truk"* (e.g., line *"Link Pacific/Chuuk Pacific/Truk"* will result in primary time zone identifier *"Pacific/Chuuk"*).

(or substitute any other example that requires use of `backzone`)

spec/locales-currencies-tz.html
```diff
@@ -285,6 +347,46 @@ <h1>
...
+          1. Repeat, while _j_ ≠ -1,
+            1. Let _T_ be the substring of _S_ from _i_ to _j_.
+            1. Append _T_ to _substrings_.
+            1. If the number of elements in _substrings_ is _limit_, return _substrings_.
+            1. Set _i_ to _j_ + _separatorLength_.
+            1. Set _j_ to StringIndexOf(_S_, _separator_, _i_).
+          1. Let _T_ be the substring of _S_ from _i_.
+          1. Append _T_ to _substrings_.
+          1. Return _substrings_.
+        </emu-alg>
```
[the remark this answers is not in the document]
justingrant: > But let's keep it super narrow for now to support evolution of e.g. _limit_ (which really should limit the search for separators rather than truncating the results) and empty string handling.

I'm happy to remove _limit_, but could you explain more about "should limit the search for separators rather than truncating the results"?  Do you mean that _limit_ should cause the last substring to contain any remaining content in the string, even if there are separators in it?

I'm asking because the two cases in the spec so far both perform well using truncation because both of those cases only care about the first few substrings.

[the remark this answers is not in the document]
justingrant: Also, if we wanted to promote this AO to 262, `String.prototype.split` uses truncation so a "rest" result wouldn't work for that case.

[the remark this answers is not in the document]
gibson042: > Do you mean that _limit_ should cause the last substring to contain any remaining content in the string, even if there are separators in it?

Yes, that is my position and IIRC there is regret on the committee that `String.prototype.split` itself does not work like that (as it does e.g. [Python `str.split`](https://docs.python.org/3/library/stdtypes.html#str.split)).

> the two cases in the spec so far both perform well using truncation because both of those cases only care about the first few substrings.

I agree, but am not confident that will remain the case and don't want to paint this into a corner. Anything that can be specified with truncation can also be specified without it, and implementations are not obligated to perform unobservable steps (which can even be reiterated in a note if we're concerned about it).

> Also, if we wanted to promote this AO to 262, `String.prototype.split` uses truncation so a "rest" result wouldn't work for that case.

No, it would be fine as e.g.
```
1. Let _substrings_ be StringSplitToList(_S_, _R_, _lim_).
1. If the number of elements in _substrings_ > _lim_, remove the last element from _substrings_.
1. Assert: the number of elements in _substrings_ ≤ _lim_.
1. Return CreateArrayFromList(_substrings_).
```
or even (with a new _maxLength_ parameter for CreateArrayFromList)
```
1. Let _substrings_ be StringSplitToList(_S_, _R_, _lim_).
1. Return CreateArrayFromList(_substrings_, _lim_).
```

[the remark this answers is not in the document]
gibson042: > Do you mean that _limit_ should cause the last substring to contain any remaining content in the string, even if there are separators in it?

Yes, that is my position and IIRC there is regret on the committee that `String.prototype.split` itself does not work like that (as it does e.g. [Python `str.split`](https://docs.python.org/3/library/stdtypes.html#str.split)).

> the two cases in the spec so far both perform well using truncation because both of those cases only care about the first few substrings.

I agree, but am not confident that will remain the case and don't want to paint this into a corner. Anything that can be specified with truncation can also be specified without it, and implementations are not obligated to perform unobservable steps (which can even be reiterated in a note if we're concerned about it).

> Also, if we wanted to promote this AO to 262, `String.prototype.split` uses truncation so a "rest" result wouldn't work for that case.

No, it would be fine as e.g.
```
1. Let _substrings_ be StringSplitToList(_S_, _R_, _lim_).
1. If the number of elements in _substrings_ > _lim_, remove the last element from _substrings_.
1. Assert: the number of elements in _substrings_ ≤ _lim_.
1. Return CreateArrayFromList(_substrings_).
```
or even (with a new _maxLength_ parameter for CreateArrayFromList)
```
1. Let _substrings_ be StringSplitToList(_S_, _R_, _lim_).
1. Return CreateArrayFromList(_substrings_, _lim_).
```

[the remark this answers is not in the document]
justingrant: OK thanks for the explanation, I'll remove the limit parameter. Note that the assertion that _S_ isn't empty means that some callsites like `IsWellFormedUnitIdentifier` need to add steps to deal with the empty string case. Is that OK?

spec/locales-currencies-tz.html
justingrant: OK, I added an editorial note in the latest commit. Let me know what you think.

spec/locales-currencies-tz.html
justingrant: Thanks for catching the bug!  I updated the spec text to take some of this suggestion but not all of it. Two things I didn't move over:
1. The `_fields_[0] is *"Link"*` assertion was redundant with the earlier StringIndexOf call (or would have been if the StringIndexOf call included the terminating space separator. So I started the StringSplitToList after `"Link "`.
2. I liked the clarity of `_backzoneAndLink_` as the local variable name instead of the more generic `_fields_`. This algorithm is pretty complex so I think any additional documentation is helpful here. 

This is in the latest commit. Let me know if this is OK!

justingrant:
> If you want to update StringSplitToList to support empty-string _S_ (with result in alignment with String.prototype.split, e.g. « **""** »), I would not be opposed. But dealing with it at call sites for now seems fine to me.

Nah, I think it's fine as-is, and I actually think it's kinda good bad to force callsites to decide how they want to handle the empty-string case because it's not always intuitive how to handle that case. 

Anyway, I adopted both of your last suggestions, rebased, and squashed. I think this PR should be ready to merge.

gibson042 (approved):
On second thought, let's save the IsWellFormedUnitIdentifier changes for a followup. This PR ought to be narrowly focused on time zones, and IsWellFormedUnitIdentifier is not the only preexisting operation that will benefit from StringSplitToList.

justingrant:
> On second thought, let's save the IsWellFormedUnitIdentifier changes for a followup.

OK, I reverted that part. I think this PR should be ready to present at Helsinki, unless there's other review feedback. Thanks @gibson042 for your help whipping this into shape!

sffc:
@justingrant Is this ready to merge, or should we wait for CLDR-17111?

justingrant:
AFAIK this is ready to merge. It lacks Test262 tests, but I assume these can be added later?

sffc:
@ben-allen If this is ready to merge, please merge it :)

FrankYFTang:
@justingrant I got the following bug https://issues.chromium.org/issues/457835186 Any comment about that?

PR #878, opened 2024-04-04
Editorial: Fix missing dot operator in Intl.Segmenter-related objects

ben-allen:
The intrinsics `%Intl.SegmentIteratorPrototype%` and `%Intl.SegmentsPrototype%` had been written without the member access operator, i.e. `%IntlSegmentIteratorPrototype%` and `%IntlSegmentsPrototype%`. Also regularized wording of `<emu-clause>` `id` attributes

### DIFF
--- a/spec/conventions.html
+++ b/spec/conventions.html
@@ -94,12 +94,12 @@
         <tr>
           <td>%Intl.SegmentIteratorPrototype%</td>
           <td></td>
-          <td>The prototype of Segment Iterator objects (<emu-xref href="#sec-intlsegmentiteratorprototype-object"></emu-xref>)</td>
+          <td>The prototype of Segment Iterator objects (<emu-xref href="#sec-intl-segmentiteratorprototype-object"></emu-xref>)</td>
         </tr>
         <tr>
           <td>%Intl.SegmentsPrototype%</td>
           <td></td>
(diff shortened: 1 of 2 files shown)


spec/conventions.html
```diff
@@ -92,12 +92,12 @@ <h1>Well-Known Intrinsic Objects</h1>
           <td>The `Intl.Segmenter` constructor (<emu-xref href="#sec-intl-segmenter-constructor"></emu-xref>)</td>
         </tr>
         <tr>
-          <td>%IntlSegmentIteratorPrototype%</td>
+          <td>%Intl.SegmentIteratorPrototype%</td>
```
anba, line 95: Per [Well-Known Intrinsic Objects](https://tc39.es/ecma262/#sec-well-known-intrinsic-objects), the syntax `%Intl.SegmentIteratorPrototype%` means that the `"SegmentIteratorPrototype"` property of the intrinsic object `%Intl%` is accessed. This wrong because there's no `Intl.SegmentIteratorPrototype` property and therefore the intrinsic name `%IntlSegmentIteratorPrototype%` has to be kept.

anba (changes requested):

ben-allen:
Closing as wrong / a bad idea

PR #869, opened 2024-02-29
Editorial: Consolidate PartitionRelativeTimePattern algorithm steps

gibson042:

PR #882, opened 2024-04-25
Editorial: Replace irregular "is equal to" language

ben-allen:
Editorial: Inspired by https://github.com/tc39/ecma402/pull/869/, regularize comparison language by replacing "is equal to" and "is not equal to" with "is" and "is not" as appropriate.


ryzokuken (approved):
Overall LGTM, would love to hear @gibson042's thoughts on this.

gibson042 (approved):
Nice!

PR #888, opened 2024-05-07
Editorial: Use "sorted according to lexicographic code unit order"

ptomato:
See tc39/ecma262#3299. This is the language preferred by the editors.


ben-allen:
LGTM, also thank you!

ben-allen (approved):

Issue #726, opened 2022-11-02
Specify canonicalization algorithms for Intl enumeration


ptomato:
Once Intl.supportedValuesOf becomes part of the specification, apply the changes proposed in https://github.com/tc39/proposal-intl-enumeration/pull/49

sffc:
@ben-allen to sync with @ptomato and coordinate on a PR.

justingrant:
Note that the time zones (unlike the other things that `Intl.supportedValuesOf` enumerates) also live in 262, because non-UTC IANA time zones (and hence the need to canonicalize them) can exist in non-402 implementations of ECMAScript. 

@gibson042 and I have been working on an editorial PR for 262 (https://github.com/tc39/ecma262/pull/3035) to specify how time zone canonicalization works there. (With editorial PRs for 402 and Temporal to follow if the 262 PR is accepted.) The summary of the PR is:
* Add an implementation-defined AO `AvailableTimeZoneIdentifiers` which returns a list of `{Identifier, CanonicalIdentifier}` records.
* Add a non-implementation-defined AO `GetAvailableTimeZoneIdentifier(id)` which returns the record where `Identifier` ASCII-case-insensitively matches `id`, or `~empty~`.
* Replace `CanonicalizeTimeZoneIdentifier` with calls to `GetAvailableTimeZoneIdentifier(id).CanonicalIdentifier`.
* Replace `IsAvailableTimeZoneIdentifier` with calls to `GetAvailableTimeZoneIdentifier(id) is not ~empty~`
* Rename `DefaultTimeZone` to `SystemTimeZoneIdentifier` to match naming of other related AOs. (And because there's no "default" time zone in Temporal.)

I don't know if this approach is relevant to other things that `Intl.supportedValuesOf` enumerates, but Richard and I would be happy to coordinate with @ben-allen and @ptomato if similar idioms could be used for those other enumerations too.

BTW, here's a few reasons for the set of AOs above: 
* Simplify and consolidate AOs and reduce the number of implementation-defined AOs required (instead of one for enumeration, another for canonicalization, another for case normalization, etc.)
* Simplify 402, Temporal, and (if accepted) [proposal-canonical-tz](https://github.com/tc39/proposal-canonical-tz) spec text.
[...]

PR #889, opened 2024-05-08
Editorial: Replace AvailableCanonicalCalendars with AvailableCalendars

ptomato:
AvailableCalendars should return all possible aliases, so that other places in the spec (e.g. in the future, validating a string calendar ID in Temporal) can use them to determine whether a given input value is valid. This input value can subsequently be canonicalized by another abstract operation, CanonicalizeUValue, which we can use in several other places.

In Intl.supportedValuesOf(), on the other hand, we should not return all possible aliases, so we filter them out using CanonicalizeUValue before returning the list of AvailableCalendars codes as an array to the caller.

See https://github.com/tc39/proposal-intl-enumeration/pull/49. This is the part of that PR that I consider relevant for the future integration of Temporal. The time zone parts were already done as part of #876. If desired, I could implement the rest of that PR, adding CanonicalizeCollation, CanonicalizeCurrency, CanonicalizeNumberingSystem, and CanonicalizeUnit as well.

Closes: #726

### DIFF
--- a/spec/intl.html
+++ b/spec/intl.html
@@ -128,7 +128,11 @@
       <emu-alg>
         1. Let _key_ be ? ToString(_key_).
         1. If _key_ is *"calendar"*, then
-          1. Let _list_ be AvailableCanonicalCalendars( ).
+          1. Let _list_ be a new empty List.
+          1. For each element _identifier_ of AvailableCalendars(), do
+            1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
+            1. If _identifier_ is _canonical_, then
+              1. Append _identifier_ to _list_.
         1. Else if _key_ is *"collation"*, then
           1. Let _list_ be AvailableCanonicalCollations( ).
         1. Else if _key_ is *"currency"*, then
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -452,14 +452,14 @@
       This specification identifies calendars using a <dfn variants="calendar types">calendar type</dfn> as defined by <a href="https://unicode.org/reports/tr35/tr35-dates.html#Calendar_Elements">Unicode Technical Standard #35 Part 4 Dates, Section 2 Calendar Elements</a>. Their canonical form is a string containing only Unicode Basic Latin lowercase letters (U+0061 LATIN SMALL LETTER A through U+007A LATIN SMALL LETTER Z) with zero or more medial hyphens (U+002D HYPHEN-MINUS).
     </p>
 
-    <emu-clause id="sec-availablecanonicalcalendars" type="implementation-defined abstract operation">
+    <emu-clause id="sec-availablecalendars" oldids="sec-availablecanonicalcalendars" type="implementation-defined abstract operation">
       <h1>
-        AvailableCanonicalCalendars (
+        AvailableCalendars (
         ): a List of Strings
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>The returned List is sorted according to lexicographic code unit order, and contains unique canonical calendar types identifying the calendars for which the implementation provides the functionality of Intl.DateTimeFormat objects. The List must include *"iso8601"*.</dd>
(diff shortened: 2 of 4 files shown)


spec/intl.html
```diff
@@ -128,7 +128,11 @@ <h1>Intl.supportedValuesOf ( _key_ )</h1>
       <emu-alg>
         1. Let _key_ be ? ToString(_key_).
         1. If _key_ is *"calendar"*, then
-          1. Let _list_ be AvailableCanonicalCalendars( ).
+          1. Let _list_ be a new empty List.
+          1. For each element _identifier_ of AvailableCalendars(), do
+            1. Let _canonical_ be CanonicalizeCalendar(_identifier_).
+            1. If _identifier_ is _canonical_, then
+              1. Append _canonical_ to _list_.
```
gibson042, line 135: Rather than introducing a calendar-specific operation, I'd like one that covers all "-u-…" keywords and is also used in [ResolveLocale](https://tc39.es/ecma402/#sec-resolvelocale) step 12.g.iii.
```suggestion
            1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
            1. If _identifier_ is _canonical_, then
              1. Append _identifier_ to _list_.
```

```
CanonicalizeUValue (
  _ukey_: a Unicode locale extension sequence key defined in <a href="https://unicode.org/reports/tr35/#Key_And_Type_Definitions_">Unicode Technical Standard #35 Part 1 Core, Section 3.6.1 Key and Type Definitions</a>,
  _uvalue_: a String,
): a String
```
```
1. Let _lowerValue_ be the ASCII-lowercase of _uvalue_.
1. Let _canonicalized_ be the String value resulting from canonicalizing _lowerValue_ as a value of key _ukey_ per <a href="https://unicode.org/reports/tr35/#processing-localeids">Unicode Technical Standard #35 Part 1 Core, Annex C LocaleId Canonicalization Section 5 Canonicalizing Syntax, Processing LocaleIds</a>.
1. NOTE: It is recommended that implementations use the 'u' extension data in <code>common/bcp47</code> provided by the Common Locale Data Repository (available at <a href="https://cldr.unicode.org/">https://cldr.unicode.org/</a>).
1. Return _canonicalized_.
```

spec/locales-currencies-tz.html
```diff
@@ -452,15 +452,33 @@ <h1>Calendar Types</h1>
...
+    <emu-clause id="sec-availablecalendars" oldids="sec-availablecanonicalcalendars" type="implementation-defined abstract operation">
       <h1>
-        AvailableCanonicalCalendars (
+        AvailableCalendars (
         ): a List of Strings
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>The returned List is sorted according to lexicographic code unit order, and contains unique canonical calendar types identifying the calendars for which the implementation provides the functionality of Intl.DateTimeFormat objects. The list must include *"iso8601"*.</dd>
+        <dd>The returned List is sorted according to lexicographic code unit order, and contains unique calendar types identifying the calendars for which the implementation provides the functionality of Intl.DateTimeFormat objects. The list must include *"iso8601"*.</dd>
```
gibson042, line 462:
```suggestion
        <dd>The returned List is sorted according to lexicographic code unit order, and contains unique calendar types in canonical form (<emu-xref href="#sec-calendar-types"></emu-xref>) identifying the calendars for which the implementation provides the functionality of Intl.DateTimeFormat objects, including their aliases (e.g., either both or neither of *"islamicc"* and *"islamic-civil"*). The List must include *"iso8601"*.</dd>
```

spec/locales-currencies-tz.html
```diff
@@ -452,15 +452,33 @@ <h1>Calendar Types</h1>
...
+        <dt>description</dt>
+        <dd>
+          The returned String is the canonical and case-regularized form of _id_.
+        </dd>
+      </dl>
+      <emu-alg>
+        1. Return the string _id_ after performing the algorithm steps to replace Unicode extension values with their canonical form per <a href="https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers">Unicode Technical Standard #35 LDML § 3.2.1 Canonical Unicode Locale Identifiers</a>, treating _id_ as an `uvalue` production.
+        1. NOTE: For example, if _id_ is *"ISLAMICC"*, return *"islamic-civil"*.
+      </emu-alg>
+    </emu-clause>
```
gibson042, line 482:
```suggestion
```

gibson042 (changes requested):

gibson042 (approved):
Great!

PR #901, opened 2024-06-21
Editorial: Fix various rebase errors

### LINKED ISSUE #899: `CreateDateTimeFormat` no longer handles `dateStyle` or `timeStyle`
It looks like commit 3deaa73ceb2cf6bf94bce2a4793a77491644ccb8 removed handling of these options, and `DateTimeStyleFormat` is never called. So with either of these options present, we do not have any formats chosen.

anba:
Fixes a couple of rebase errors. Also reverts attempts to address review feedback ("pattern" instead of "format" for some records), because the changes were largely unfinished and actually using "pattern" would result in equally confusing code like `pattern.[[pattern]]` which doesn't seem like an improvement when compared to `format.[[pattern]]`. For example see  [FormatDateTimePattern](https://tc39.es/ecma402/#sec-formatdatetimepattern), which has `format` and `pattern` arguments.

Fixes #899.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -60,7 +60,7 @@
       </dl>
 
       <emu-alg>
-        1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%DateTimeFormat.prototype%"*, &laquo; [[InitializedDateTimeFormat]], [[Locale]], [[Calendar]], [[NumberingSystem]], [[TimeZone]], [[HourCycle]], [[DateStyle]], [[TimeStyle]], [[DateTimeFormat]], [[BoundFormat]] &raquo;).
+        1. Let _dateTimeFormat_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%Intl.DateTimeFormat.prototype%"*, &laquo; [[InitializedDateTimeFormat]], [[Locale]], [[Calendar]], [[NumberingSystem]], [[TimeZone]], [[HourCycle]], [[DateStyle]], [[TimeStyle]], [[DateTimeFormat]], [[BoundFormat]] &raquo;).
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Set _options_ to ? CoerceOptionsToObject(_options_).
         1. Let _opt_ be a new Record.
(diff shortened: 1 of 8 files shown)


anba:
(<https://github.com/tc39/proposal-temporal/pull/2904> depends on these change. This is mentioned in the commit messages for the other PR, but not in the description, so explicitly mention it here so that GH can create cross-reference links.)

ben-allen (approved):
Thanks for this! Barring objections, I'll merge this midday tomorrow @gibson042

spec/datetimeformat.html
```diff
@@ -1403,31 +1412,31 @@ <h1>
         1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, *"minimumIntegerDigits"*, *2*<sub>𝔽</sub>).
         1. Perform ! CreateDataPropertyOrThrow(_nf2Options_, *"useGrouping"*, *false*).
         1. Let _nf2_ be ! Construct(%Intl.NumberFormat%, &laquo; _locale_, _nf2Options_ &raquo;).
-        1. Let _fractionalSecondDigits_ be _dateTimeFormat_.[[FractionalSecondDigits]].
-        1. If _fractionalSecondDigits_ is *undefined*, set _fractionalSecondDigits_ to 1.
```
ben-allen, line 1407: I was looking back over this one while reviewing #903 and I realized I had some questions about the removal of this line. The message for the commit where it was added  (731d597dff84102b7753bbd05d3cd20e49a1fedb) observes that `_fractionalSecondDigits_` being `*undefined*` here is a theoretical case / doesn't affect implementations; is it worthwhile to either keep this line in or else leave a note about it?

anba: The issue described in 731d597dff84102b7753bbd05d3cd20e49a1fedb no longer applies, because the number of fractional second digits are now directly read from the format record instead of the `Intl.DateTimeFormat` object.  

Before 731d597dff84102b7753bbd05d3cd20e49a1fedb it was possible that [PartitionDateTimeRangePattern](https://tc39.es/ecma402/2023/#sec-partitiondatetimerangepattern) calls [FormatDateTimePattern](https://tc39.es/ecma402/2023/#sec-formatdatetimepattern) with a `patternPart` whose `[[Type]]` is `"fractionalSecondDigits"`. Step 15.c.iii then performs:
> Let fv be FormatNumeric(nf3, v).

But `nf3` is only defined when `dateTimeFormat.[[FractionalSecondDigits]]` is not `undefiend`, cf. steps 11-12. 

This case can no longer apply in the current [FormatDateTimePattern](https://tc39.es/ecma402/#sec-formatdatetimepattern) operation, because if the pattern string `pattern` contains the substring `"{fractionalSecondDigits}"`, then `format` has a field `[[fractionalSecondDigits]]`, cf. steps 9 and 13.c.i. And the requirements in [Table 5](https://tc39.es/ecma402/#datetimeformat-format-record):
> Optional field. Present if [[pattern]] contains the substring "{fractionalSecondDigits}".

And [Table 7](https://tc39.es/ecma402/#datetimeformat-range-pattern-format-record):
> Optional field. Present if a Pattern String in [[PatternParts]] contains the substring "{fractionalSecondDigits}".

PR #907, opened 2024-07-18
Editorial: replaced &ge; &infin; &laquo; &ldquo; &le; &mdash; &ne; &r…

ben-allen:
https://github.com/tc39/ecma402/issues/634 observed that ECMA-262 had standardized on using HTML character entities rather than non-ASCII characters such as ≤ and ≥. However, in the time since that issue was raised 262 re-standardized on directly using the Unicode characters rather than using HTML entities. 

This PR replaces the relevant HTML character entities with the Unicode characters they represent.

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -15,7 +15,7 @@
       </p>
       <emu-alg>
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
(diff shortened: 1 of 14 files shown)


ben-allen:
Could I get a quick review of this one? It's been sitting for a minute. @gibson042 @ryzokuken

ben-allen:
@gibson042 quick review? Mostly I want to make sure that my interpretation of 262's standard for html entities is correct.

ptomato (approved):
I did a similar change in the Temporal proposal recently, this looks correct.

Issue #691, opened 2022-06-13
Numeric range validation is inconsistent


gibson042:
[DefaultNumberOption](https://tc39.es/ecma402/#sec-defaultnumberoption) (used by [GetNumberOption](https://tc39.es/ecma402/#sec-getnumberoption) and [SetNumberFormatDigitOptions](https://tc39.es/ecma402/#sec-setnfdigitoptions) for processing options like "fractionalSecondDigits", "minimumIntegerDigits", and "{minimum,maximum}{Fraction,Significant}Digits") converts its primary argument to a Number value, validates the result against the range bounds, and then returns its truncation (implemented with floor but equivalent because no existing use specifies a negative lower bound).

As a result, implementations are required to exhibit strange behavior like `minimumFractionDigits` 19.9 being synonymous with 19 but 20.9 _not_ being synonymous with 20. It is also inconsistent with ECMA-262, in which e.g. `x.toFixed(-0.9)` is synonymous with `x.toFixed(0)` and `x.toFixed(100.9)`is synonymous with `x.toFixed(100)`.

This is coming up in https://github.com/tc39/proposal-temporal/issues/2296 , which affects new ECMA-262 behavior inspired by ECMA-402. I think the divergence between ECMA-262 and ECMA-402 here should be reduced, allowing Temporal to comfortably adopt the latter without being too jarring in the context of the former. Specifically, ECMA-402 DefaultNumberOption should perform range validation after truncation (converting some current RangeError cases into valid input) but maintain rejection of non-numeric input:
```patch
         1. If _value_ is *undefined*, return _fallback_.
         1. Set _value_ to ? ToNumber(_value_).
-        1. If _value_ is *NaN* or less than _minimum_ or greater than _maximum_, throw a *RangeError* exception.
-        1. Return floor(_value_).
+        1. If _value_ is *NaN*, throw a *RangeError* exception.
+        1. Let _value_ be truncate(ℝ(_value_)).
+        1. If _value_ &lt; _minimum_ or _value_ &gt; _maximum_, throw a *RangeError* exception.
+        1. Return _value_.
```
(note: truncate will be added to ECMA-262 by https://github.com/tc39/ecma262/pull/2781 ; for nonnegative input it is equivalent to floor)

sffc:
2024-03-28 TG2 notes: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-03-28.md#numeric-range-validation-is-inconsistent-691

[...]

PR #908, opened 2024-07-18
Normative: Make DefaultNumberOption truncate before validating range

ben-allen:
Currently the behaviour of ECMA-402 differs from ECMA-262 in that 402's [`DefaultNumberOption`](https://tc39.es/ecma402/#sec-defaultnumberoption) validates that the value passed in is within the specified range before truncating that value, whereas 262's [`toFixed`](https://tc39.es/ecma262/#sec-number.prototype.tofixed) truncates before range validation. This PR changes 402 to align with 262. See https://github.com/tc39/ecma402/issues/691, also https://github.com/tc39/proposal-temporal/issues/2296

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -426,7 +426,7 @@
       <emu-alg>
         1. If _value_ is *undefined*, return _fallback_.
         1. Set _value_ to ? ToNumber(_value_).
-        1. If _value_ is not finite or _value_ is *NaN*, throw a *RangeError* exception.
+        1. If _value_ is not finite, throw a *RangeError* exception.
         1. Set _value_ to truncate(ℝ(_value_)).
         1. If _value_ &lt; _minimum_ or _value_ &gt; _maximum_, throw a *RangeError* exception.
         1. Return _value_.
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -426,8 +426,10 @@
       <emu-alg>
         1. If _value_ is *undefined*, return _fallback_.
         1. Set _value_ to ? ToNumber(_value_).
-        1. If _value_ is not finite or ℝ(_value_) &lt; _minimum_ or ℝ(_value_) &gt; _maximum_, throw a *RangeError* exception.
-        1. Return floor(ℝ(_value_)).
+        1. If _value_ is not finite or _value_ is *NaN*, throw a *RangeError* exception.
+        1. Set _value_ to truncate(ℝ(_value_)).
+        1. If _value_ &lt; _minimum_ or _value_ &gt; _maximum_, throw a *RangeError* exception.
+        1. Return _value_.
       </emu-alg>
     </emu-clause>
 


spec/negotiation.html
```diff
@@ -426,8 +426,10 @@ <h1>
       <emu-alg>
         1. If _value_ is *undefined*, return _fallback_.
         1. Set _value_ to ? ToNumber(_value_).
-        1. If _value_ is not finite or ℝ(_value_) &lt; _minimum_ or ℝ(_value_) &gt; _maximum_, throw a *RangeError* exception.
-        1. Return floor(ℝ(_value_)).
+        1. If _value_ is not finite or _value_ is *NaN*, throw a *RangeError* exception.
```
michaelficarra, line 429:
```suggestion
        1. If _value_ is not finite, throw a *RangeError* exception.
```

NaN is not finite. See https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#finite.

michaelficarra:
From a normative perspective, this seems  to go against our normative convention for rejecting non-integral arguments where we expect integers: https://github.com/tc39/how-we-work/blob/main/normative-conventions.md#integral-number-taking-inputs-should-reject-non-integral-arguments.

Why isn't this just throwing a RangeError for these invalid inputs?

anba:
> From a normative perspective, this seems to go against our normative convention for rejecting non-integral arguments where we expect integers: https://github.com/tc39/how-we-work/blob/main/normative-conventions.md#integral-number-taking-inputs-should-reject-non-integral-arguments.

Does that mean Temporal should be changed to avoid [ToIntegerWithTruncation](https://tc39.es/proposal-temporal/#sec-tointegerwithtruncation)?

> Why isn't this just throwing a RangeError for these invalid inputs?

We can't really change `DefaultNumberOption` to throw for non-integer inputs, for example `new Intl.NumberFormat("en", {minimumFractionDigits: 2.5}).format(1)` currently works and returns `"1.00"`. 

This PR only allows two more possible inputs:
- When zero is a valid input, `x ∈ ]-1, 0[` is now additionally valid, too. Example: `new Intl.NumberFormat("en", {minimumFractionDigits: -0.5}).format(1)`
- When `max` is the maximum valid integer, `x ∈ ]max, max + 1[` is now valid, too. Example:  `new Intl.NumberFormat("en", {minimumFractionDigits: 100.5}).format(1)`

ben-allen:
I'll put in a link to the plenary's conclusions on this PR once the notes are up on GitHub, but in the meantime here's a summary:

* We shouldn't change 402 to have the same behaviour as 262 in this case, since checking for this user error is more than valuable than consistency would be.
* It's definitely not worth the effort to change the behaviour of 262 to match 402's current behaviour.
* We'll investigate whether changing 402's behaviour to reject non-integers is worth the effort (I'm thinking no; see anba's comment above).
* Future proposals should reject non-integers in similar cases.

PR #914, opened 2024-08-05
Editorial: Rename [[RelevantExtensionKeys]] to [[LocaleExtensionKeys]] for %Intl.Locale%

anba:
Rename the internal slot name to avoid confusion with the `[[RelevantExtensionKeys]]` for Intl service constructors, because `Intl.Locale` isn't an Intl service constructor.

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -17,11 +17,11 @@
 
       <emu-alg>
         1. If NewTarget is *undefined*, throw a *TypeError* exception.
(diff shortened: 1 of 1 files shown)


ben-allen:
oh, thank you! this clarifies things a great deal &mdash; I've been caught out before on differences between the behaviour of `Intl.Locale` and the other Intl objects.

ben-allen (approved):

PR #915, opened 2024-08-05
Editorial: Add GetLocaleBaseName abstract operation

anba:
Add `GetLocaleBaseName` to avoid repeating the same steps to extract the `unicode_language_id` parts of a `unicode_locale_id`. The abstract operation is named to `GetLocaleBaseName` to match the accessor property `Intl.Locale.prototype.baseName`.

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -82,21 +82,22 @@
         <dd>It updates the <code>unicode_language_id</code> subtags in _tag_ from the corresponding properties of _options_ and returns the <emu-xref href="#sec-isstructurallyvalidlanguagetag">structurally valid</emu-xref> but non-canonicalized result.</dd>
       </dl>
       <emu-alg>
-        1. Let _languageId_ be the longest prefix of _tag_ matched by the <code>unicode_language_id</code> Unicode locale nonterminal.
-        1. Let _language_ be ? GetOption(_options_, *"language"*, ~string~, ~empty~, GetLocaleLanguage(_languageId_)).
+        1. Let _baseName_ be GetLocaleBaseName(_tag_).
+        1. Let _language_ be ? GetOption(_options_, *"language"*, ~string~, ~empty~, GetLocaleLanguage(_baseName_)).
         1. If _language_ cannot be matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _script_ be ? GetOption(_options_, *"script"*, ~string~, ~empty~, GetLocaleScript(_languageId_)).
+        1. Let _script_ be ? GetOption(_options_, *"script"*, ~string~, ~empty~, GetLocaleScript(_baseName_)).
         1. If _script_ is not *undefined*, then
           1. If _script_ cannot be matched by the <code>unicode_script_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _region_ be ? GetOption(_options_, *"region"*, ~string~, ~empty~, GetLocaleRegion(_languageId_)).
+        1. Let _region_ be ? GetOption(_options_, *"region"*, ~string~, ~empty~, GetLocaleRegion(_baseName_)).
         1. If _region_ is not *undefined*, then
           1. If _region_ cannot be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _variants_ be GetLocaleVariants(_languageId_).
-        1. Let _newLanguageId_ be _language_.
-        1. If _script_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _script_.
-        1. If _region_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _region_.
-        1. If _variants_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _variants_.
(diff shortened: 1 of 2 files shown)


spec/locale.html
```diff
@@ -82,21 +82,22 @@ <h1>
...
         1. If _region_ is not *undefined*, then
           1. If _region_ cannot be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _variants_ be GetLocaleVariants(_languageId_).
-        1. Let _newLanguageId_ be _language_.
-        1. If _script_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _script_.
-        1. If _region_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _region_.
-        1. If _variants_ is not *undefined*, set _newLanguageId_ to the string-concatenation of _newLanguageId_, *"-"*, and _variants_.
-        1. Let _newTag_ be _tag_ with the <emu-not-ref>substring</emu-not-ref> matched by the <code>unicode_language_id</code> Unicode locale nonterminal replaced by the string _newLanguageId_.
+        1. Let _variants_ be GetLocaleVariants(_baseName_).
+        1. Let _allExtensions_ be the suffix of _tag_ following _baseName_.
```
anba, line 95: Taking the suffix string matches the approach taken in `IsStructurallyValidLanguageTag`. This also let's us avoid having to reference `unicode_language_id` below and we can instead use simple string concatenation.

spec/locale.html
```diff
@@ -377,10 +391,9 @@ <h1>
       <dl class="header">
       </dl>
       <emu-alg>
-        1. Assert: _locale_ can be matched by the <code>unicode_locale_id</code> Unicode locale nonterminal.
-        1. Let _languageId_ be the longest prefix of _locale_ matched by the <code>unicode_language_id</code> Unicode locale nonterminal.
-        1. Assert: The first subtag of _languageId_ can be matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal.
-        1. Return the first subtag of _languageId_.
+        1. Let _baseName_ be GetLocaleBaseName(_locale_).
```
ben-allen, line 394: Is there a reason why the calls to `GetLocaleBaseName` in `GetLocaleLanguage`, `GetLocaleRegion`, etc. can't occur in the associated getters rather than in the AOs themselves, in the interest of not calling `GetLocaleBaseName` twice when calling these AOs from `UpdateLanguageId`?

ben-allen, line 394: Is there a reason why the calls to `GetLocaleBaseName` can't occur in the associated getters rather than in `GetLocaleLanguage`, `GetLocaleScript`, etc., in the interest of not calling it twice when calling these AOs from `UpdateLanguageId`?

spec/locale.html
```diff
@@ -368,6 +368,20 @@ <h1>Properties of Intl.Locale Instances</h1>
...
+    <emu-clause id="sec-getlocalebasename" type="abstract operation">
+      <h1>
+        GetLocaleBaseName (
+          _locale_: a String,
+        ): a String
+      </h1>
+      <dl class="header">
+      </dl>
+      <emu-alg>
+        1. Assert: _locale_ can be matched by the <code>unicode_locale_id</code> Unicode locale nonterminal.
```
ben-allen, line 380: Can this Assert be moved into the header?

spec/locale.html
ben-allen, line 394: Is there a reason why the calls to GetLocaleBaseName can't occur in the associated getters rather than in GetLocaleLanguage, GetLocaleScript, etc., in the interest of not calling it twice when calling these AOs from UpdateLanguageId?

ben-allen (commented):
This is essentially an approve &mdash; I have a couple of questions, but I suspect they're based on me missing subtleties about spec language rather than actual editorial problems

spec/locale.html
anba: The call to `GetLocaleBaseName` is spec-only and won't actually happen in implementations. Therefore it seems nicer to allow any `unicode_locale_id` input for `GetLocale{Language,Script,Region}` instead of restricting the input to `unicode_language_id`. (Also I'd prefer if we have fewer references to `unicode_language_id`, because it's too easy to confuse with `unicode_language_subtag`.) 

I'd even prefer if `UpdateLanguageId` calls `GetLocale{Language,Script,Region}` with `tag` instead of `baseName`/`languageId`, because first extracting the base name parts seems like a spec-only optimisation which doesn't affect implementations anyway. For example compare:
- https://github.com/WebKit/WebKit/blob/95b06f4446ccccd849b94e81168963f7cb7ceb2e/Source/JavaScriptCore/runtime/IntlLocale.cpp#L266-L288
- https://github.com/v8/v8/blob/94e627d588d941d5e454adb4296219056a7c4fc8/src/objects/js-locale.cc#L288-L340
- https://github.com/mozilla/gecko-dev/blob/e942f7bc56cd103bba86b396ddeba5b1ab04f1a4/js/src/builtin/intl/Locale.cpp#L314-L390

No implementation actually calls `GetLocale{Language,Script,Region}`, because that requires additional string allocations when compared to only updating the relevant subtags when an option is not `undefined`.

spec/locale.html
anba: Is there any precedent for this kind of type declaration? Do you have something like this in mind?

```
GetLocaleBaseName (
  _locale_: a String matching `unicode_locale_id`,
): a String matching `unicode_language_id`
```

Unless we're already using something similar elsewhere, I'd prefer to not introduce new kinds of type declarations at this point of time.

ben-allen (approved):

PR #916, opened 2024-08-06
Editorial: Remove stray emu-clause close tag.

anba:
Drive-by change:
- Fix indentation for `StringSplitToList`.


anba:
The `</emu-clause>` tag was incorrectly added in 602aaf0b408d3cb427f1598fe079ae3d54ed8b8c.

Issue #819, opened 2023-08-04
Numbering system option value not correctly passed to internal Intl.NumberFormat instances


anba:
`new Intl.RelativeTimeFormat("en", {numberingSystem:"hanidec"}).format(2, "days")` is currently spec'ed to return the string `"in 2 days"` instead of `"in 二 days"`. 

`new Intl.RelativeTimeFormat("en-u-nu-hanidec").format(2, "days")` is correctly spec'ed to return `"in 二 days"`, though.

[InitializeRelativeTimeFormat](https://tc39.es/ecma402/#sec-InitializeRelativeTimeFormat)
> 19. Let relativeTimeFormat.[[NumberFormat]] be ! Construct(%NumberFormat%, « locale »).

Should instead be:

> 19. Let nfOptions be OrdinaryObjectCreate(null).
> 20. Perform ! CreateDataPropertyOrThrow(nfOptions, "numberingSystem", relativeTimeFormat.[[NumberingSystem]]).
> 21. Let relativeTimeFormat.[[NumberFormat]] be ! Construct(%NumberFormat%, « locale, nfOptions »).

Also applies to the `Intl.NumberFormat` instances in [FormatDateTimePattern](https://tc39.es/ecma402/#sec-formatdatetimepattern).

PR #919, opened 2024-08-21
Normative: Fix bug with RelativeTimeFormat and DateTimeFormat not correctly passing "numberingSystem" option to NumberFormat

ben-allen:
Fix https://github.com/tc39/ecma402/issues/819

This fixes a spec bug; as far as I can tell, all actually existing implementations do this the right way rather than the specified way. Will change title to start with "Normative:" if this category of fix requires TG1 consensus.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -1404,16 +1404,19 @@
       <emu-alg>
         1. Let _locale_ be _dateTimeFormat_.[[Locale]].
         1. Let _nfOptions_ be OrdinaryObjectCreate(*null*).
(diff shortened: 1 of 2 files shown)


sffc:
TG2 agreement: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-08-22.md#fix-bug-with-relativetimeformat-not-correctly-passing-numberingsystem-option

foxrider598 (approved):

FrankYFTang:
I believe this is presented to Oct TC39 meeting and still waiting for Mozilla to express their opinion.

ben-allen:
I believe they changed their conditional approval to approval at the start of the third day of plenary, but may have misunderstood. @codehag

FrankYFTang:
Should we also add test to test262 to test this PR got implement correctly?

ben-allen:
yes, I'll be writing them soon

Issue #835, opened 2023-09-18
Support cash rounding


sffc:
See https://github.com/tc39/ecma402/issues/134

CLDR has both normal (financial) rounding and cash rounding for currencies. We should consider adding an option to Intl.NumberFormat to support cash rounding.

sffc:
Sample API:

```javascript
new Intl.NumberFormat(undefined, { style: "currency", currencyRounding: "cash" })
```

Should the option be named:

- currencyRounding
- currencyPrecision
- currencyUse

ben-allen:
I've put up PR #839  for this as an item for discussion. The PR uses currencyPrecision for the name of the option. currencyRounding seems bad (even though it's the most straightforward seeming) since it implies that only the rounding increment changes. currencyUse appears as valid as currencyPrecision, though.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/master/meetings/notes-2024-01-18.md#support-cash-rounding-835

Conclusion: Add Test262 and docs for the current spec behavior.

ryzokuken:
I believe this was closed by https://github.com/tc39/ecma402/pull/922.

sffc:
We never added cash rounding. However, the discussion above makes it sound like we couldn't identify a use case for cash rounding, and are actively skeptical that adding such an option would benefit the platform. So I think we can close this and re-open in the future if someone can come with clear use cases where they need this.

Issue #134, opened 2017-03-10
Permit CLDR rather than ISO 4217 for CurrencyDigits?


littledan:
[CurrencyDigits](https://tc39.github.io/ecma402/#sec-currencydigits) is an abstract operation which normatively references ISO 4217 to get the default number of digits to display for a currency. That data can be found [on ISO 4217's website in this XML file](https://www.currency-iso.org/dam/downloads/lists/list_one.xml).

CLDR has its own notion of how many digits should be used for currency display by default. [Here](http://www.unicode.org/cldr/charts/30/supplemental/detailed_territory_currency_information.html#format_info) is the data for CLDR 30.

These charts actually differ in a number of places. For example, afghani (AFN) is listed as having 2 decimal digits in ISO 4217, but 0 in CLDR 30. [According to Wikipedia](https://en.wikipedia.org/wiki/Afghan_afghani), the new afghani (AFN) has never had subdivisions issued, though it is nominally divided into 100 smaller units.

I'm not sure what accounts for this difference, whether ISO 4217 is going for more legalistic definitions and CLDR is going for what's linguistically relevant, or whether one has fresher data than the other, or something else. I'd be interested if anyone has more context here.

V8 has been using CLDR data for CurrencyDigits, whereas SpiderMonkey (and now JSC) has been using ISO 4217 data. I wrote patches to switch SpiderMonkey and JSC to switch to CLDR, not realizing, until it was caught in SpiderMonkey's test automation (thanks @syg!), that this was actually a semantic change, and introduced a spec violation caught by [a test262 test](https://github.com/tc39/test262/blob/master/test/intl402/NumberFormat/11.1.1_20_c.js).

Apparently, either source is web-compatible, but I don't know which one is better. Does anyone have more background here?

cc @jungshik

NorbertLindenberg:
The ISO 4217 definition of the decimal digits for currencies is indeed legalistic and sometimes detached from reality. I contacted the maintenance agency in 2013 to ask about the Indonesian rupiah, which was listed as having two decimal digits. In reality, the smallest coin in Indonesia is 50 rupiah, which is worth about half a US cent, and a 1000:1 redenomination of the rupiah has been considered for years. The maintenance agency contacted the Central Bank of Indonesia, which replied that the minor unit was still used in accounting and therefore should remain. And so it remained.

littledan:
[...]

PR #921, opened 2024-08-27
Editorial: Note on use of CLDR data for rounding currency values

ben-allen:
For some currencies, CLDR gives two values for determining the number of minor digits displayed when formatting those currencies. This note explains that the value of the "cashDigits" attribute, and likewise the value of the "cashRounding" attribute, should not be used when formatting values denominated in those currencies.

fix https://github.com/tc39/ecma402/issues/835 (or at least, close it). 

address https://github.com/tc39/ecma402/issues/134

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -697,6 +697,11 @@
         1. Assert: _currency_ is the ASCII-uppercase of _currency_.
         1. If the ISO 4217 currency and funds code list contains _currency_ as an alphabetic code, return the minor unit value corresponding to the _currency_ from the list; otherwise, return 2.
       </emu-alg>
+
+      <emu-note>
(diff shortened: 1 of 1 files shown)


anba:
The current spec text doesn't even allow to use CLDR in the first place, because it explicitly references the "ISO 4217 currency and funds code list". I think we first need to a normative change to either allow in CLDR in addition to ISO 4217 or alternatively use CLDR instead of ISO 4217. 

IIRC it depends on the specific application if ISO 4217 or CLDR is a better choice. And for some applications neither is actually the best choice.

ben-allen:
Replaces https://github.com/tc39/ecma402/pull/839

sffc:
Does CLDR endeavor to match ISO-4217? Or are the CLDR currency digit settings just different from the ISO ones?

ben-allen:
Closed because replaced by https://github.com/tc39/ecma402/pull/922, which makes the currency list used purely implementation-defined.

PR #922, opened 2024-09-10
Normative: Allow use of non-ISO 4217 data in `CurrencyDigits` AO

### LINKED ISSUE #134: Permit CLDR rather than ISO 4217 for CurrencyDigits?

ben-allen:
This PR allows for locale data other than ISO 4217 for determining the number of fractional digits to be used when formatting currency values.

In some applications ISO 4217 is most appropriate, in some CLDR data, and in some neither is best. See https://github.com/tc39/ecma402/pull/921#issuecomment-2313173054

Resolves #134

This PR makes the `CurrencyDigits` AO implementation-defined.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -697,7 +697,6 @@
 
       <emu-alg>
         1. Assert: IsWellFormedCurrencyCode(_currency_) is *true*.
-        1. Assert: _currency_ is the ASCII-uppercase of _currency_.
         1. Return a non-negative integer indicating the number of fractional digits used when formatting quantities of the currency corresponding to _currency_. If there is no available information on the number of digits to be used, return 2.
       </emu-alg>
     </emu-clause>
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -684,18 +684,21 @@
   <emu-clause id="sec-numberformat-abstracts">
     <h1>Abstract Operations for NumberFormat Objects</h1>
 
-    <emu-clause id="sec-currencydigits" type="abstract operation">
+    <emu-clause id="sec-currencydigits" type="implementation-defined abstract operation">
       <h1>
         CurrencyDigits (
           _currency_: a String,
         ): a non-negative integer
       </h1>
       <dl class="header">
+        <dt>description</dt>
+        <dd></dd>
       </dl>
+
       <emu-alg>
         1. Assert: IsWellFormedCurrencyCode(_currency_) is *true*.
         1. Assert: _currency_ is the ASCII-uppercase of _currency_.
-        1. If the ISO 4217 currency and funds code list contains _currency_ as an alphabetic code, return the minor unit value corresponding to the _currency_ from the list; otherwise, return 2.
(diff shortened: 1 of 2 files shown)


spec/numberformat.html
```diff
@@ -684,18 +684,21 @@ <h1>Properties of Intl.NumberFormat Instances</h1>
...
         ): a non-negative integer
       </h1>
       <dl class="header">
+        <dt>description</dt>
+        <dd></dd>
       </dl>
+
       <emu-alg>
         1. Assert: IsWellFormedCurrencyCode(_currency_) is *true*.
         1. Assert: _currency_ is the ASCII-uppercase of _currency_.
```
gibson042, line 700: I don't think we need this assertion if we're not going directly against ISO 4217.
```suggestion
```

gibson042 (approved):

spec/numberformat.html
dmrazzy: Change

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-09-26.md#normative-allow-use-of-non-iso-4217-data-in-currencydigits-ao-922

sffc:
CLDR issue: https://unicode-org.atlassian.net/browse/CLDR-18008

FrankYFTang:
I believe this PR is discussed in TC39 2024-10-09 meeting and reach consensus

ben-allen:
Received consensus at October 2024 plenary, with guidance to make future editorial changes giving recommended data source

Issue #405, opened 2020-01-31
ECMA-402 contents should be arranged more consistently


gibson042:
Suggested template for Foo functionality:
* Foo Objects
  1. The Intl.Foo Constructor
     1. Intl.Foo ( [ _locales_ [ , _options_ ] ] )
  1. Properties of the Intl.Foo Constructor
     1. Intl.Foo.prototype
     1. Intl.Foo.supportedLocalesOf ( _locales_ [, _options_ ] )
     1. Internal slots
  1. Properties of the Intl.Foo Prototype Object
     1. Intl.Foo.prototype.constructor
     1. Intl.Foo.prototype.resolvedOptions ()
     1. ..._Foo-specific properties in alphabetic order_...
     1. Intl.Foo.prototype [ %Symbol.toStringTag% ] (**_or maybe this could be absent from new extensions?_**)
  1. Properties of Intl.Foo Instances
  1. ..._other Foo-specific surface area_...
  1. Abstract Operations for Foo Objects

The biggest differences of this as compared to http://tc39.es/ecma402/ are in the ordering under "Properties of the Intl.Foo Prototype Object" (which currently all go like "constructor", "@@toStringTag", ..., "resolvedOptions") and in pushing abstract operations to be after the references to them (rather than before the Constructor definition as they are now).

--------

It also looks like a great deal of the constructor code could be abstracted:
1. [**_since 2018 edition PluralRules, cf. #384_**] If NewTarget is *undefined*, throw a *TypeError* exception.
1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
1. Let _obj_ be ? OrdinaryCreateFromConstructor(_newTarget_, _proto_, _internalSlotsList_).
1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
1. [**_since 2021 edition DisplayNames and ListFormat, cf. #493 and #538_**] Set _options_ to ? GetOptionsObject(_options_).
1. Set _options_ to ? CoerceOptionsToObject(_options_).
1. Let _matcher_ be ? GetOption(_options_, **"localeMatcher"**, <tt>STRING</tt>, « **"lookup"**, **"best fit"** », **"best fit"**).
1. Let _opt_ be the Record { [[localeMatcher]]: _matcher_ }.
1. For each row of $RelevantExtensionKeysTable, except the header row, in table order, do
   1. Let _key_ be the Extension Key value of the current row.
[...]

PR #923, opened 2024-09-12
Editorial: Arrange contents consistently

ben-allen:
Resolves the following aspects of #405:

Placed Prototype Object properties in the following order for all Intl Objects:

1. Intl._Foo_.prototype.constructor
2. Intl._Foo_.prototype.resolvedOptions ()
3. Intl._Foo_.prototype [ %Symbol.toStringTag% ]
4. ..._Foo_-specific properties in alphabetic order...

Gathered all Abstract Operations associated with each Object into the Abstract Operations section for that Object. Previously Intl.DateTimeFormat, Intl.Locale, and Intl.NumberFormat had several Abstract Operations appear inside the constructor clauses of those Objects.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -894,95 +894,6 @@
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-intl.datetimeformat.prototype-%symbol.tostringtag%" oldids="sec-intl.datetimeformat.prototype-@@tostringtag">
-      <h1>Intl.DateTimeFormat.prototype [ %Symbol.toStringTag% ]</h1>
-
-      <p>
-        The initial value of the %Symbol.toStringTag% property is the String value *"Intl.DateTimeFormat"*.
-      </p>
-
-      <p>
-        This property has the attributes { [[Writable]]: *false*, [[Enumerable]]: *false*, [[Configurable]]: *true* }.
-      </p>
-    </emu-clause>
-
-    <emu-clause id="sec-intl.datetimeformat.prototype.format">
-      <h1>get Intl.DateTimeFormat.prototype.format</h1>
-
-      <p>
-        Intl.DateTimeFormat.prototype.format is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:
-      </p>
-
-      <emu-alg>
-        1. Let _dtf_ be the *this* value.
-        1. If the implementation supports the normative optional constructor mode of <emu-xref href="#legacy-constructor"></emu-xref>, then
-          1. Set _dtf_ to ? UnwrapDateTimeFormat(_dtf_).
-        1. Perform ? RequireInternalSlot(_dtf_, [[InitializedDateTimeFormat]]).
-        1. If _dtf_.[[BoundFormat]] is *undefined*, then
-          1. Let _F_ be a new built-in function object as defined in DateTime Format Functions (<emu-xref href="#sec-datetime-format-functions"></emu-xref>).
-          1. Set _F_.[[DateTimeFormat]] to _dtf_.
-          1. Set _dtf_.[[BoundFormat]] to _F_.
-        1. Return _dtf_.[[BoundFormat]].
-      </emu-alg>
-
-      <emu-note>
-        The returned function is bound to _dtf_ so that it can be passed directly to `Array.prototype.map` or other functions.
-        This is considered a historical artefact, as part of a convention which is no longer followed for new features, but is preserved to maintain compatibility with existing programs.
-      </emu-note>
-    </emu-clause>
-
-    <emu-clause id="sec-Intl.DateTimeFormat.prototype.formatToParts">
-      <h1>Intl.DateTimeFormat.prototype.formatToParts ( _date_ )</h1>
-
-      <p>
-        When the `formatToParts` method is called with an argument _date_, the following steps are taken:
-      </p>
-
-      <emu-alg>
-        1. Let _dtf_ be the *this* value.
-        1. Perform ? RequireInternalSlot(_dtf_, [[InitializedDateTimeFormat]]).
-        1. If _date_ is *undefined*, then
-          1. Let _x_ be ! Call(%Date.now%, *undefined*).
-        1. Else,
-          1. Let _x_ be ? ToNumber(_date_).
-        1. Return ? FormatDateTimeToParts(_dtf_, _x_).
-      </emu-alg>
-    </emu-clause>
-
-    <emu-clause id="sec-intl.datetimeformat.prototype.formatRange">
-      <h1>Intl.DateTimeFormat.prototype.formatRange ( _startDate_, _endDate_ )</h1>
-
-      <p>
-        When the `formatRange` method is called with arguments _startDate_ and _endDate_, the following steps are taken:
-      </p>
-
-      <emu-alg>
-        1. Let _dtf_ be *this* value.
-        1. Perform ? RequireInternalSlot(_dtf_, [[InitializedDateTimeFormat]]).
-        1. If _startDate_ is *undefined* or _endDate_ is *undefined*, throw a *TypeError* exception.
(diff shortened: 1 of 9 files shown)


anba:
ECMA-262 sorts built-in properties alphabetically, with Symbol-keyed properties being sorted to the end. Does it make sense to follow in ECMA-402?

gibson042:
I'm ambivalent about that, especially dropping `resolvedOptions` at a varying position for each prototype—there's something to be said for a simple rule, but also for tweaking that rule to highlight internal consistency. On balance, I think I prefer the latter, ideally including documentation justifying it.

ben-allen:
My preference strongly matches @gibson042's -- `resolvedOptions` in predictable place, add text to the style guide explaining the choice. Should we discuss it at TG2?

spec/locale.html
```diff
@@ -203,38 +125,6 @@ <h1>Intl.Locale.prototype [ %Symbol.toStringTag% ]</h1>
       </p>
     </emu-clause>
 
-    <emu-clause id="sec-Intl.Locale.prototype.maximize">
```
gibson042, line 206: Intl.Locale is not a service constructor, so we should probably just follow ECMA-262 conventions—string-keyed properties alphabetically, then symbol-keyed properties alphabetically (i.e., Symbol.toStringTag last).

spec/datetimeformat.html
```diff
@@ -1245,6 +1083,168 @@ <h1>Abstract Operations for DateTimeFormat Objects</h1>
...
+            1. For each property name _prop_ of « *"year"*, *"month"*, *"day"* », do
+              1. Set _formatOptions_.[[&lt;_prop_&gt;]] to *"numeric"*.
+          1. If _needDefaults_ is *true* and _defaults_ is either ~time~ or ~all~, then
+            1. For each property name _prop_ of « *"hour"*, *"minute"*, *"second"* », do
+              1. Set _formatOptions_.[[&lt;_prop_&gt;]] to *"numeric"*.
+          1. Let _formats_ be _resolvedLocaleData_.[[formats]].[[&lt;_resolvedCalendar_&gt;]].
+          1. If _formatMatcher_ is *"basic"*, then
+            1. Let _bestFormat_ be BasicFormatMatcher(_formatOptions_, _formats_).
+          1. Else,
+            1. Let _bestFormat_ be BestFitFormatMatcher(_formatOptions_, _formats_).
```
gibson042, line 1219: Nit for future reference: we should really rename these to DTF-specific operations to indicate that detail (e.g., DateTimeFormat{Basic,BestFit}Matcher).

spec/datetimeformat.html
```diff
@@ -1245,6 +1083,168 @@ <h1>Abstract Operations for DateTimeFormat Objects</h1>
...
+        1. Let _absoluteMinutes_ be abs(_offsetMinutes_).
+        1. Let _hours_ be floor(_absoluteMinutes_ / 60).
+        1. Let _minutes_ be _absoluteMinutes_ modulo 60.
+        1. Return the string-concatenation of _sign_, ToZeroPaddedDecimalString(_hours_, 2), the code unit 0x003A (COLON), and ToZeroPaddedDecimalString(_minutes_, 2).
+      </emu-alg>
+    </emu-clause>
+
     <emu-clause id="sec-date-time-style-format" type="abstract operation">
       <h1>
         DateTimeStyleFormat (
```
gibson042, line 1250: Nit for a followup (if only to minimize review burden): we should also consider alphabetizing these AO lists, and moving/reorganizing where that would hinder comprehension.

spec/listformat.html
```diff
@@ -202,6 +161,47 @@ <h1>Intl.ListFormat.prototype.resolvedOptions ( )</h1>
         </table>
       </emu-table>
     </emu-clause>
+
+    <emu-clause id="sec-Intl.ListFormat.prototype-toStringTag">
```
gibson042, line 165: Noticed in passing:
```suggestion
    <emu-clause id="sec-intl.listformat.prototype-%symbol.tostringtag%" oldids="sec-Intl.ListFormat.prototype-toStringTag">
```

(nd likewise for RelativeTimeFormat (or we could save both for a bigger `id` cleanup that also covers non-symbol sections)

spec/numberformat.html
```diff
@@ -64,148 +64,6 @@ <h1>Intl.NumberFormat ( [ _locales_ [ , _options_ ] ] )</h1>
           1. Return ? ChainNumberFormat(_numberFormat_, NewTarget, _this_).
         1. Return _numberFormat_.
       </emu-alg>
-
-      <emu-normative-optional>
-      <emu-clause id="sec-chainnumberformat" type="abstract operation">
-        <h1>
-          ChainNumberFormat (
```
gibson042, line 71: Same comment here regarding constructor-specific operations as for DTF.

gibson042 (approved):
Looking at the rendered spec, I think I do have a slight preference for Symbol.toStringTag properties being after string-keyed properties—especially since they might not even be universal in the future.

spec/datetimeformat.html
```diff
@@ -23,168 +23,6 @@ <h1>Intl.DateTimeFormat ( [ _locales_ [ , _options_ ] ] )</h1>
           1. Return ? ChainDateTimeFormat(_dateTimeFormat_, NewTarget, _this_).
         1. Return _dateTimeFormat_.
       </emu-alg>
-
-      <emu-normative-optional>
-      <emu-clause id="sec-chaindatetimeformat" type="abstract operation">
-        <h1>
-          ChainDateTimeFormat (
```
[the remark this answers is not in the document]
ben-allen: Thanks for this one! I was torn between leaving them where they were or putting them down with the rest of the AOs. The least-bad argument I have for moving them down is the possibility of those AOs getting called from elsewhere in the spec -- though upon brief perusal the only place that seems to happen is the call to `SetNumberFormatDigitOptions` from the PluralRules constructor. I'll move them back.

spec/datetimeformat.html
ben-allen: I'm provisionally skeptical about alphabetizing the AO lists, largely because the decision on when to break from alphabetical order for higher comprehension is difficult / might introduce inconsistencies that are themselves a source of confusion. I'd almost rather just have the rule be alphabetical order, full stop, no exceptions. Open to arguments otherwise, though -- this isn't a strong preference, and it's likely a somewhat uninformed one.

ben-allen:
pushed a commit addressing @gibson042 concerns, largely through dropping the commit that moved constructor-specific AOs down to the section with the rest of the AOs.

spec/datetimeformat.html
gibson042: To clarify, that's in line with my suggestion... the kind of moving/reorganization that I suggest, where necessary, relates to document structure (e.g., [FilterLocales](https://tc39.es/ecma402/#sec-filterlocales) and [ResolveLocale](https://tc39.es/ecma402/#sec-resolvelocale) could be alphabetically sorted as the only children of a "Locale Matching" subsection).

ben-allen: ah, that makes a lot of sense! I've put it on my todo list

spec/locale.html
gibson042: > Intl.Locale is not a service constructor, so we should probably just follow ECMA-262 conventions—string-keyed properties alphabetically, then symbol-keyed properties alphabetically (i.e., Symbol.toStringTag last).

This one still stands; please move at least `Intl.Locale.prototype [ %Symbol.toStringTag% ]` to the end of "Properties of the Intl.Locale Prototype Object".

gibson042 (approved):
I'd still [prefer Symbol.toStringTag properties to follow string-keyed properties](https://github.com/tc39/ecma402/pull/923#pullrequestreview-2363930156), but can live with this.

spec/datetimeformat.html
gibson042: #961

ryzokuken (approved):

PR #924, opened 2024-09-17
Editorial: Define ILD and ILND in exactly one place (4.4)

ben-allen:
Previously ILD acronym had been defined in several different places, likewise ILND. New version defines these exactly once, in 4.4. Implementation Dependencies


spec/overview.html
```diff
@@ -73,6 +73,10 @@ <h1>Implementation Dependencies</h1>
       </li>
     </ul>
 
+    <p>
+      Throughout this specification, behaviour that is dependent on implementation and locale is referred to as ILD. Behaviour that is dependent on implementation, locale, and numbering system is referred to as ILND.
```
gibson042, line 77: These initialisms should be auto-linked, and ideally also continue to convey their source phrasing.
```suggestion
      Throughout this specification, implementation- and locale-dependent behaviour is referred to as <dfn>ILD</dfn>, and implementation-, locale-, and numbering system-dependent behaviour is referred to as <dfn>ILND</dfn>.
```

gibson042 (commented):
There are also some phrases to replace in spec/datetimeformat.html and spec/numberformat.html:
* https://github.com/tc39/ecma402/blob/60eda81014ed237c7cb0fe1fc2620928f0fa8054/spec/datetimeformat.html#L1470-L1480
* https://github.com/tc39/ecma402/blob/60eda81014ed237c7cb0fe1fc2620928f0fa8054/spec/numberformat.html#L564

ben-allen:
@gibson042 updated based on your suggestions!

PR #925, opened 2024-09-17
Normative: When formatting currency values, only use data on the number of minor units used to display that currency when using standard notation.

### LINKED ISSUE #912: Currency-derived fractional digit count defaults are inappropriately applied with scientific/engineering/etc. `notation`
_(originally posted by @gibson042 in https://github.com/tc39/ecma402/pull/839#discussion_r1370737507 )_

We arguably have a bug where currency-derived fractional digit count defaults are applied even when the decimal point is moved away from its absolute position, as in `new Intl.NumberFormat("en", { style: "currency", currency, currencyDisplay: "code", notation: "engineering" }).format(12345).replace(/^.*\s/, "")` resulting in "12E3" for JPY but "12.35E3" for USD).

Put another way, I think I expect currency-derived precision and corresponding rounding to apply only when `notation` is "standard" (i.e., not even when the decimal point _happens_ to be in the right place with scientific/engineering/compact notation).

ben-allen:
Previously this data was inappropriately used to set the number of fractional digits displayed when formatting currency values in scientific, engineering, and compact notations. Fix #912.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -35,6 +35,9 @@
         1. Let _style_ be _numberFormat_.[[Style]].
         1. If _style_ is *"currency"*, then
           1. Let _currency_ be _numberFormat_.[[Currency]].
+        1. Let _notation_ be ? GetOption(_options_, *"notation"*, ~string~, « *"standard"*, *"scientific"*, *"engineering"*, *"compact"* », *"standard"*).
+        1. Set _numberFormat_.[[Notation]] to _notation_.
+        1. If _style_ is *"currency"* and *"notation"* is *"standard"*, then
           1. Let _cDigits_ be CurrencyDigits(_currency_).
           1. Let _mnfdDefault_ be _cDigits_.
           1. Let _mxfdDefault_ be _cDigits_.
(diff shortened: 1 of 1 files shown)


sffc (approved):

gibson042 (approved):
I love it when the fix is so simple!

gibson042:
P.S. We should also try to cover this in test262.

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-09-26.md#normative-when-formatting-currency-values-only-use-data-on-the-number-of-minor-units-used-to-display-that-currency-when-using-standard-notation-925

sffc:
CLDR spec change: https://github.com/unicode-org/cldr/pull/4112

FrankYFTang:
I believe this is presented to Oct TC39 meeting and still waiting for Mozilla to express their opinion.

spec/numberformat.html
```diff
@@ -35,6 +35,9 @@ <h1>Intl.NumberFormat ( [ _locales_ [ , _options_ ] ] )</h1>
         1. Let _style_ be _numberFormat_.[[Style]].
         1. If _style_ is *"currency"*, then
           1. Let _currency_ be _numberFormat_.[[Currency]].
```
FrankYFTang, line 37: Why is the line " Let _currency_ be _numberFormat_.[[Currency]]." not inside the same if block of 
```
          1. Let _cDigits_ be CurrencyDigits(_currency_).
          1. Let _mnfdDefault_ be _cDigits_.
          1. Let _mxfdDefault_ be _cDigits_.
```
Nowhere else need the value of _currency_ in this AO, right?

FrankYFTang:
v8 tracking bug https://issues.chromium.org/issues/372731679
v8 cl https://chromium-review.googlesource.com/c/v8/v8/+/5925080

ben-allen:
Achieved consensus at October 24 TC39 plenary

FrankYFTang:
https://github.com/tc39/test262/issues/4259

PR #928, opened 2024-09-25
Editorial: Clarify and simplify GetLocaleVariants and IsStructurallyValidLanguageTag

gibson042:
Replace complex substring descriptions with better use of subtag references and existing operations.


ben-allen (approved):
Much clearer, especially the changes to locales-currencies-tz. The note in locale.html initially caused me more confusion than it relieved, but I genuinely cannot think of a better wording.

PR #929, opened 2024-09-26
Normative: Add 8 new numbering systems for Unicode 16

FrankYFTang:
Unicode 16 added 8 new numbering systems for the newly added 5,185 new characters, including 3,995 additional Egyptian Hieroglyph characters plus seven new scripts, seven new emoji characters, and over 700 symbols from legacy computing environments. see https://blog.unicode.org/2024/09/announcing-unicode-standard-version-160.html

CLDR add that in https://github.com/unicode-org/cldr/pull/3658 which will be available in ICU 76

This PR update the ECMA402 spec to include that new numbering systems. 

We should consider merging it after ICU 76.1 is wildely available to implementer. We probably should discuss this in TG2 around Nov. or Dec. 2024.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -1006,6 +1006,10 @@
             <td>U+FF10 to U+FF19</td>
           </tr>
           <tr>
+            <td>gara</td>
+            <td>U+10D40 to U+10D49</td>
+          </tr>
+          <tr>
             <td>gong</td>
             <td>U+11DA0 to U+11DA9</td>
           </tr>
@@ -1016,6 +1020,10 @@
           <tr>
             <td>gujr</td>
             <td>U+0AE6 to U+0AEF</td>
+          </tr>
+          <tr>
+            <td>gukh</td>
+            <td>U+16130 to U+16139</td>
           </tr>
           <tr>
             <td>guru</td>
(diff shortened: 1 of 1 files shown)


FrankYFTang:
@sffc @ben-allen we should talk about this around Nov/Dec 2024 in TG2 but not earlier since ICU 76.1 is not release yet.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-11-25.md#normative-add-8-new-numbering-systems-for-unicode-16-929

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-12-19.md#normative-add-8-new-numbering-systems-for-unicode-16-929

gibson042 (approved):

ryzokuken:
We should add test262 tests in order to validate that these new numbering systems are supported.

ryzokuken:
We asked TG1 for consensus, right? Could someone link the notes to the wiki?

dminor:
> We asked TG1 for consensus, right? Could someone link the notes to the wiki?

I don't remember this coming to TG1 and was unable to find anything by grepping the notes.

ryzokuken:
I'll recheck the notes again but I have a feeling I presented it in one of our ECMA-402 updates back when we asked for consensus on normative PRs in that section. If not, let's get consensus on this change from TG1 at the next plenary.

Issue #635, opened 2022-01-06
Inline GetOperands into PluralRuleSelect


gibson042:
Following up on #594, I'd like to inline [GetOperands](https://tc39.es/ecma402/#sec-getoperands) into [PluralRuleSelect](https://tc39.es/ecma402/#sec-pluralruleselect), replacing the Record _operands_ parameter of the latter with the String that is currently input of the former ([[FormattedString]] output from [FormatNumericToString](https://tc39.es/ecma402/#sec-formatnumberstring)). The implementation-defined behavior of PluralRuleSelect would still have access to what is now the input of GetOperands and could deconstruct it in that way if/when appropriate.
```patch
         1. Let _res_ be ! FormatNumericToString(_pluralRules_, _n_).
-        1. Let _s_ be _res_.[[FormattedString]].
-        1. Let _operands_ be ! GetOperands(_s_).
-        1. Return ! PluralRuleSelect(_locale_, _type_, _n_, _operands_).
+        1. Return ! PluralRuleSelect(_locale_, _type_, _n_, _res_.[[FormattedString]]).
```

We might also be able to remove the Number _n_ parameter of PluralRuleSelect, but only if implementations are required to select the same plural category for distinct numbers that format identically, such as `"one"` for `1.01` when rounding options format it as `"1"` (which I *think* is the case but would need confirmation):
```
        1. Return ! PluralRuleSelect(_locale_, _type_, _res_.[[FormattedString]]).
```

ben-allen:
I'm reading through these older issues -- do you think this one is still worth pursuing?

@gibson042

gibson042:
Yes. It's weird that ResolvePlural takes in a number, formats it to a string, calls out to a dedicated GetOperands operation (_i.e., called from nowhere else_) for deconstructing that string and reconstructing a number value for it, and then passes the resulting Record to an opaque implementation-defined PluralRuleSelect operation. Because PluralRuleSelect is opaque, implementations already have the freedom to perform the steps of GetOperands within it as warranted.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2024-08-22.md#inline-getoperands-into-pluralruleselect

PR #930, opened 2024-09-27
Editorial: Inlined Intl.PluralRules's GetOperands AO.

ben-allen:
`GetOperands` was only called from the `ResolvePlural` AO, which immediately passed the Record returned by `GetOperands` to the implementation-defined PluralRuleSelect AO. The revised version of `ResolvePlural` directly passes `PluralRuleSelect` the decimal String which was previously parsed by `GetOperands`. Implementations of `PluralRuleSelect` may (or may not) use something like `GetOperands` to determine the correct plural category.

fixes https://github.com/tc39/ecma402/issues/635

### DIFF
--- a/spec/pluralrules.html
+++ b/spec/pluralrules.html
@@ -312,7 +312,9 @@
           1. Return the Record { [[PluralCategory]]: *"other"*, [[FormattedString]]: _s_ }.
         1. Let _res_ be FormatNumericToString(_pluralRules_, ℝ(_n_)).
         1. Let _s_ be _res_.[[FormattedString]].
-        1. Let _p_ be PluralRuleSelect(_pluralRules_, _s_).
+        1. Let _locale_ be _pluralRules_.[[Locale]].
+        1. Let _type_ be _pluralRules_.[[Type]].
+        1. Let _p_ be PluralRuleSelect(_locale_, _type_, _s_).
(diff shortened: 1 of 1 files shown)


spec/pluralrules.html
```diff
@@ -281,106 +281,16 @@ <h1>Properties of Intl.PluralRules Instances</h1>
...
-
     <emu-clause id="sec-pluralruleselect" type="implementation-defined abstract operation">
       <h1>
         PluralRuleSelect (
-          _locale_: a String,
-          _type_: *"cardinal"* or *"ordinal"*,
-          _n_: a finite Number,
-          _operands_: a Plural Rules Operands Record derived from formatting _n_,
+          _pluralRules_: an Intl.PluralRules,
+          _s_: a decimal String,
```
gibson042, line 288: I think this is going too far; PluralRuleSelect is implementation-defined but can still only act on its input (which should therefore be appropriately scoped and similar to PluralRuleSelectRange).
```suggestion
          _locale_: a language tag,
          _type_: *"cardinal"* or *"ordinal"*,
          _s_: a decimal String,
```

gibson042 (approved):
Thanks! I have a suggested tweak, but I'm so happy to see this simplification!

ben-allen:
Look good? @gibson042

gibson042 (approved):

PR #931, opened 2024-10-11
Editorial: Simplify PR925

FrankYFTang:
There are no reason we need to put the "Let _currency_ be _numberFormat_.[[Currency]]." line in a separate if block. Move that down


ben-allen (approved):
ah, good catch.

PR #938, opened 2024-11-11
Editorial: Handle case when backzone doesn't contain link

anba:
A couple of links aren't in `backzone`. Change the assertion to an if-step to correctly handle this case.

List of time zones which weren't correctly handled:
```
Africa/Kinshasa
Africa/Lubumbashi
America/Atikokan
America/Blanc-Sablon
America/Creston
Antarctica/DumontDUrville
Antarctica/McMurdo
Antarctica/Syowa
Asia/Kuala_Lumpur
Europe/Busingen
Pacific/Chuuk
Pacific/Majuro
Pacific/Midway
Pacific/Pohnpei
Pacific/Wake
```

For example "Africa/Kinshasa", which is a zone in `backzone`:
- "Africa/Kinshasa" is a link to "Africa/Lagos" per `backward`.
- The region code for "Africa/Lagos" is "NG", the region code for "Africa/Kinshasa" is "CD". That means the condition in step 5.b.iii.3 evaluates to false.
- There are two entries for the region "CD" in `zone.tab`: "Africa/Kinshasa" and "Africa/Lubumbashi". That means the condition in step 5.b.iii.4.b also evaluates to false.
- "Africa/Kinshasa" is a zone in `backzone`, so the loop in step 5.b.iii.4.c.iii won't find any entries.

All other cases are also zones in `backzone`, except for "Europe/Busingen", which is always a link. "Europe/Busingen" links to "Europe/Zurich" per `backward`, but is still a primary identifier in ECMA-402, because Büsingen is politically a part of Germany, so it shouldn't be merged with the time zone for Zürich (Switzerland).

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -282,8 +282,8 @@
                     1. If _backzoneAndLink_[1] is _identifier_, then
                       1. Assert: _backzone_ is *undefined*.
                       1. Set _backzone_ to _backzoneAndLink_[0].
-                  1. Assert: _backzone_ is not *undefined*.
-                  1. Set _primary_ to _backzone_.
+                  1. If _backzone_ is not *undefined*, then
+                    1. Set _primary_ to _backzone_.
           1. If _primary_ is one of *"Etc/UTC"*, *"Etc/GMT"*, or *"GMT"*, set _primary_ to *"UTC"*.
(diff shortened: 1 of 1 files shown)


anba:
Whoops, seems like I've skipped over the "[...] identifier is not present in the “TZ” column of `zone.tab`" part of the condition. Sorry about the false alarm!

PR #939, opened 2024-11-11
Editorial: Correct markup for 'notation' variable

anba:


ben-allen:
Thank you for finding this one!

ben-allen (approved):

PR #943, opened 2024-11-22
Normative: add Intl.DurationFormat

ben-allen:
https://github.com/tc39/proposal-intl-duration-format

### DIFF
--- a/spec/durationformat.html
+++ b/spec/durationformat.html
@@ -154,7 +154,7 @@ contributors: Ujjwal Sharma, Younies Mahmoud
       </h1>
       <dl class="header">
         <dt>description</dt>
-        <dd>It returns 1 if the most significant non-zero element in its arguments is positive, and -1 if the most significant non-zero element is negative. If all of its arguments are zero, it returns 0.</dd>
+        <dd>It returns 1 if the most significant non-zero field in the _duration_ argument is positive, and -1 if the most significant non-zero field is negative. If all of _duration_'s fields are zero, it returns 0.</dd>
       </dl>
       <emu-alg>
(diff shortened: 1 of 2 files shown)


ben-allen:
fix https://github.com/tc39/proposal-intl-duration-format/issues/209

spec/durationformat.html
```diff
@@ -0,0 +1,1093 @@
...
+        1. NOTE: The above step cannot be implemented directly using floating-point arithmetic. Multiplying by 10<sup>-3</sup>, 10<sup>-6</sup>, and 10<sup>-9</sup> respectively may be imprecise when _milliseconds_, _microseconds_, or _nanoseconds_ is an unsafe integer. This multiplication can be implemented in C++ with an implementation of `std::remquo()` with sufficient bits in the quotient. String manipulation will also give an exact result, since the multiplication is by a power of 10.
+        1. If abs(_normalizedSeconds_) ≥ 2<sup>53</sup>, return *false*.
+        1. Return *true*.
+      </emu-alg>
+      <emu-note>The abstract operations ToIntegerIfIntegral, DurationSign, and IsValidDuration take the same parameters as the abstract operations of the same name in the Temporal proposal's specification, aside from the use of a Duration Record instead of a Temporal.Duration instance. They will be removed in favor of the equivalent Temporal abstract operations if the Temporal proposal becomes part of ECMAScript before this proposal does.</emu-note>
```
ptomato, line 204: I'd delete this note, it's irrelevant once this PR is merged.

ryzokuken: hmm, I was under the impression that this note would need to be changed and that the idea would be that once Temporal lands the AOs would be removed from 402 and we'd use AOs from 262 instead.

ptomato: Yes, that's still the idea, but I don't think the note is relevant. Temporal moves several more AOs from 402 into 262, not just these.

ryzokuken:
Added a fixup commit with the new changes in the proposal so we could review those in isolation.

Please let me know if something is amiss, otherwise I'll squash the commits together and merge.

gibson042:
@ryzokuken This merged just a little bit before I got to it; see #963 for followup editorial work.

PR #948, opened 2024-12-10
Editorial: Add two more required formats

anba:
This makes format selection in BasicFormatMatcher unambiguous when requesting the following two formats:

1. Standalone "month":

There's currently a tie between "year, month" and "month, day". Adding standalone "month" as a required format resolves this tie.

2. "dayPeriod, hour, minute, second, fractionalSecondDigits":

There's a tie between "hour, minute, second, fractionalSecondDigits" and "dayPeriod, hour, minute, second". Requiring a format which has both "dayPeriod" and "fractionalSecondDigits" resolves this tie.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -264,12 +264,14 @@
             <li>year, month, day</li>
             <li>year, month</li>
             <li>month, day</li>
+            <li>month</li>
             <li>hour, minute, second, fractionalSecondDigits</li>
             <li>hour, minute, second</li>
             <li>hour, minute</li>
-            <li>dayPeriod, hour</li>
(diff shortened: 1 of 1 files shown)


sffc (approved):
LGTM but is it editorial?

anba:
If we consider these ties as spec bugs, I'd say this is editorial. Otherwise it's normative, because it adds new requirements for implementations. Except no (non-experimental) implementation actually implements `BasicFormatMatcher`, so implementers can likely just ignore this PR. 

(Even if there's no actual implementation for `BasicFormatMatcher`, it's still useful to keep it in the spec, because it defines the minimum requirements for `BestFitFormatMatcher`.)

gibson042 (approved):

ryzokuken (approved):

PR #953, opened 2025-01-16
Editorial: Distinguish 'available time zone identifier'

ptomato:
Temporal will introduce separate definitions for 'time zone identifier' (string consisting of characters that are legal in time zone identifiers) and 'available time zone identifier' (time zone identifier that is known to the implementation.)

For some usages it's probably more appropriate to speak of available time zone identifiers than time zone identifiers. For others, it's probably more appropriate to use 'available named time zone identifier' which is a definition given in ECMA-262.

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -1159,7 +1159,7 @@
       <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
       <li>[[Calendar]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier">Unicode Calendar Identifier</a> used for formatting.</li>
       <li>[[NumberingSystem]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeNumberSystemIdentifier">Unicode Number System Identifier</a> used for formatting.</li>
-      <li>[[TimeZone]] is a String value used for formatting that is either a time zone identifier from the IANA Time Zone Database or a UTC offset in ISO 8601 extended format.</li>
+      <li>[[TimeZone]] is a String value used for formatting that is an available time zone identifier.</li>
       <li>[[HourCycle]] is a String value indicating whether the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*) should be used. *"h11"* and *"h23"* start with hour 0 and go up to 11 and 23 respectively. *"h12"* and *"h24"* start with hour 1 and go up to 12 and 24. [[HourCycle]] is only used when [[DateTimeFormat]] has an [[hour]] field.</li>
       <li>[[DateStyle]], [[TimeStyle]] are each either *undefined*, or a String value with values *"full"*, *"long"*, *"medium"*, or *"short"*.</li>
       <li>[[DateTimeFormat]] is a DateTime Format Record.</li>
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -170,7 +170,7 @@
     <h1>Use of the IANA Time Zone Database</h1>
 
     <p>
-      Implementations that adopt this specification must be time zone aware: they must use the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a> to supply time zone identifiers and data used in ECMAScript calculations and formatting.
+      Implementations that adopt this specification must be time zone aware: they must use the IANA Time Zone Database <a href="https://www.iana.org/time-zones/">https://www.iana.org/time-zones/</a> to supply available named time zone identifiers and data used in ECMAScript calculations and formatting.
       This section defines how the IANA Time Zone Database should be used by time zone aware implementations.
       No String may be an available named time zone identifier unless it is a Zone name or a Link name in the IANA Time Zone Database.
       Available named time zone identifiers returned by ECMAScript built-in objects must use the casing found in the IANA Time Zone Database.
@@ -196,7 +196,7 @@
 
     <emu-note>
       <p>
-        The IANA Time Zone Database offers build options that affect which time zone identifiers are primary.
+        The IANA Time Zone Database offers build options that affect which available named time zone identifiers are primary.
         The default build options merge different countries' time zones, for example *"Atlantic/Reykjavik"* is built as a Link to the Zone *"Africa/Abidjan"*.
         Geographically and politically distinct locations are likely to introduce divergent time zone rules in a future version of the IANA Time Zone Database.
         The exceptions above serve to mitigate these future-compatibility issues.


justingrant (approved):
Good changes, thanks. 

Dumb question: if "time zone identifier" is just a string that satisfies a particular syntax, then why not just use the syntax-directed operation (not sure if I'm using that term correctly) instead of a custom-defined term?  This would make it obvious to readers that you're just talking about a string with valid syntax, rather than any actual meaningful identifier.

sffc:
Will this show up as a definition link like "mathematical value" for example?

ptomato:
> Will this show up as a definition link like "mathematical value" for example?

It will eventually. We'd need to add the definition "An available time zone identifier is either an available named time zone identifier or an offset time zone identifier." to https://tc39.es/ecma262/#sec-time-zone-identifiers. This addition is part of Temporal, but we could upstream that sentence into ECMA-262 earlier if need be.

> Dumb question: if "time zone identifier" is just a string that satisfies a particular syntax, then why not just use the syntax-directed operation (not sure if I'm using that term correctly) instead of a custom-defined term? This would make it obvious to readers that you're just talking about a string with valid syntax, rather than any actual meaningful identifier.

You can't tell just from the syntax whether a time zone identifier is available or not, you need the TZDB. We could write a syntax-directed operation that says to consult the TZDB, but I suspect implementations would object to that because it'd complicate the dependencies of their parsers.

justingrant:
> > Dumb question: if "time zone identifier" is just a string that satisfies a particular syntax, then why not just use the syntax-directed operation (not sure if I'm using that term correctly) instead of a custom-defined term? This would make it obvious to readers that you're just talking about a string with valid syntax, rather than any actual meaningful identifier.
> 
> You can't tell just from the syntax whether a time zone identifier is available or not, you need the TZDB. We could write a syntax-directed operation that says to consult the TZDB, but I suspect implementations would object to that because it'd complicate the dependencies of their parsers.

Sorry, I was talking about "time zone identifier" not "available time zone identifier". The former is just syntax, right? So can we call it something like "a _TimeZoneIdentifier_ Parse Node" in the spec so it's clear that it's just syntax? Like this?

image

I'm pushing on this because I think "time zone identifier" vs. "available time zone identifier" feels confusingly similar to each other, and by more clearly calling out that one is just syntax, then it removes that potential source of confusion.

gibson042 (approved):
The changes as-is look good to me, although I would not necessarily object to more iteration. @justingrant's point about similarity is valid, and elsewhere in ECMA-402 we do have the concept of a "[structurally valid language tag](https://tc39.es/ecma402/#sec-language-tags)", which is broader than the list of _supported_ locale identifiers that appear in an [Available Locales List](https://tc39.es/ecma402/#available-locales-list).

ptomato:
I quite like the idea of "structurally valid time zone identifier", does that make it clear enough that it's parseable but not necessarily available?

ptomato:
We'd probably have to define that in ECMA-262, though: https://tc39.es/ecma262/#sec-time-zone-identifiers The advantage of keeping "time zone identifier" for now, is that we wouldn't need to block this PR.

If we agree here that "structurally valid time zone identifier" is good, my preference would be to introduce that in ECMA-262 first, and when that makes it through the editing process I'll write PRs for ECMA-402 and Temporal using the new terminology. In the meantime, we could merge this PR.

sffc:
+1 for qualifying all instances of "time zone identifier".

Sounds like we will have 3? "structurally valid", "available", and "primary"?

justingrant:
> +1 for qualifying all instances of "time zone identifier".
> 
> Sounds like we will have 3? "structurally valid", "available", and "primary"?

There's also "non-primary". 

> If we agree here that "structurally valid time zone identifier" is good, my preference would be to introduce that in ECMA-262 first, and when that makes it through the editing process I'll write PRs for ECMA-402 and Temporal using the new terminology. In the meantime, we could merge this PR.

I admit that I'm not convinced that there's a need for a new term. This space is already jumbled with different terms with names that are very similar to each other.

How many uses of this term will exist?  If only a few, can we just say something like this?  
> a string that when passed to `ParseText` returns a Parse Node for the [TimeZoneIdentifier](https://tc39.es/proposal-temporal/#prod-TimeZoneIdentifier) goal symbol"

I also don't prefer "available time zone identifier" because it's so similar to "available named time zone identifier".  Seems fine to just say "available named time zone identifier or an offset time zone identifier" if we only need it in a few places.

gibson042:
In terms of sets and relationships (and using maximally precise names without committing to them), we have
* available named time zone identifier = primary available named time zone identifier ∪ non-primary available named time zone identifier ∪ offset time zone identifier
* primary available named time zone identifier ∩ non-primary available named time zone identifier = ∅
* primary available named time zone identifier ⊆ structurally valid named time zone identifier
* non-primary available named time zone identifier ⊊ structurally valid named time zone identifier (_this is a strict subset because "UTC" is required to be primary_)
* offset time zone identifier ∩ structurally valid named time zone identifier = ∅

ptomato:
If we want to make more changes to this terminology beyond what I proposed in this PR, I'll have to coordinate three PRs between ECMA-262, ECMA-402, and Temporal. Before starting that work, I'd like to make sure that there's a shared understanding among everyone of where we intend to land.

I would also prefer to land this PR in the meantime, to make things as consistent as possible with the terminology that we are currently using.

spec/datetimeformat.html
```diff
@@ -1159,7 +1159,7 @@ <h1>Properties of Intl.DateTimeFormat Instances</h1>
       <li>[[Locale]] is a String value with the language tag of the locale whose localization is used for formatting.</li>
       <li>[[Calendar]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier">Unicode Calendar Identifier</a> used for formatting.</li>
       <li>[[NumberingSystem]] is a String value representing the <a href="https://unicode.org/reports/tr35/#UnicodeNumberSystemIdentifier">Unicode Number System Identifier</a> used for formatting.</li>
-      <li>[[TimeZone]] is a String value used for formatting that is either a time zone identifier from the IANA Time Zone Database or a UTC offset in ISO 8601 extended format.</li>
+      <li>[[TimeZone]] is a String value used for formatting that is an available time zone identifier.</li>
```
justingrant, line 1162: I'm skeptical about using the term "available time zone identifier". I'd prefer to avoid it here so we can more easily remove it later from other specs if that's what we decide. 

```suggestion
      <li>[[TimeZone]] is a String value used for formatting that is either an available named time zone identifier or an offset time zone identifier.</li>
```

justingrant:
I added one comment with suggested alternative text that doesn't use the term "available time zone identifier" which is a term I don't think should exist. If you're OK with that suggestion then I'm 👍  to merge this PR. 

Sorry, I forgot to bring up this topic in the last Temporal Champions meeting!  We can probably resolve it in a sidebar next week in Seattle.

sffc (approved):

justingrant:
Temporal Meeting 2025-03-20: 👍  from @sffc, @gibson042 , @nekevss, @justingrant. Let's merge this!

PR #956, opened 2025-01-30
Normative: Update String toLocale{Lower,Upper}Case to ResolveLocale with best-fit matching

### LINKED ISSUE #896: toLocaleLowerCase/toLocaleUpperCase should better align with the rest of ECMA-402
String.prototype.{toLocaleLowerCase,toLocaleUpperCase} were introduced in 9ee894044e558cac13ede075a3bdc26037075189 and first appeared in the [second edition of ECMA-402](https://402.ecma-international.org/2.0/index.html). But toLocaleLowerCase and toLocaleUpperCase have some gratuitous differences from the preexisting String.prototype.localeCompare and Intl.Collator (the former implemented on top of the latter)—specifically,
1. They resolve a locale using only the first element of a _locales_ array, even if it is not supported but later elements are (e.g., on my machine in en-US, `"I".toLocaleLowerCase(["tr"])` is the dotless "ı" but `"I".toLocaleLowerCase(["zzz", "tr"])` is "i", while `new Intl.Collator(["zzz", "tr"], { sensitivity: "variant" }).compare("i", "ı")` has the same non-default behavior as `new Intl.Collator(["tr"], { sensitivity: "variant" }).compare("i", "ı")`).
2. The resolved locale is selected by prefix rather than best fit (although this might not be observable).
3. They don't have an _options_ parameter, which if it did exist would support a "localeMatcher" property for configuring the previous.

The second two points probably don't matter, at least with current locale data. But the first difference would surprise anyone sufficiently familiar with Intl APIs, and results in a worse end-user experience. I'd like to fix it by aligning the String methods with other Intl APIs, specifically fixing at least the first two points by replacing [TransformCase](https://tc39.es/ecma402/#sec-transform-case)'s direct use of `LookupMatchingLocaleByPrefix(_availableLocales_, _noExtensionsLocale_)` (which is accidentally invalid as of f16f3ad5f57a72f2a479253f78a396b76d6d67c5 and should be something like `LookupMatchingLocaleByPrefix(_availableLocales_, « _noExtensionsLocale_ »)`) with e.g. `ResolveLocale(_availableLocales_, _requestedLocales_, _resolutionOptions_, « », _emptyLocaleData_).[[Locale]]` (which would also be a change to using the default locale rather than "und" when the first requested locale is not supported) or `ResolveLocale(_availableLocales_, _requestedLocales_, _resolutionOptions_, « », _emptyLocaleData_, *"und"*).[[Locale]]` (extending ResolveLocale to support a configurable default in support of preserving the current behavior).

gibson042:
Fixes #896

* Do not ignore locales after the first in the list returned by CanonicalizeLocaleList(_locales_) (observable via e.g. `"I".toLocaleLowerCase(["zzz", "tr"]) === "ı"`).
* Match against the items of that list by best-fit rather than prefix, aligning with the rest of ECMA-402 (although the difference is not necessarily observable).
* When the list is not empty but no matching locale is found, default to DefaultLocale() rather than "und", aligning with empty-list behavior and with the rest of ECMA-402 (observable via e.g. `"I".toLocaleLowerCase("zzz") === "I".toLocaleLowerCase()` regardless of default locale, just like `new Intl.Collator("zzz", { sensitivity: "variant" }).compare("i", "ı") === new Intl.Collator(undefined, { sensitivity: "variant" }).compare("i", "ı")`).

### DIFF
--- a/spec/locale-sensitive-functions.html
+++ b/spec/locale-sensitive-functions.html
@@ -79,13 +79,11 @@
         </dl>
         <emu-alg>
           1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
-          1. If _requestedLocales_ is not an empty List, then
-            1. Let _requestedLocale_ be _requestedLocales_[0].
-          1. Else,
-            1. Let _requestedLocale_ be DefaultLocale().
           1. Let _availableLocales_ be an Available Locales List which includes the language tags for which the Unicode Character Database contains language-sensitive case mappings. If the implementation supports additional locale-sensitive case mappings, _availableLocales_ should also include their corresponding language tags.
-          1. Let _match_ be LookupMatchingLocaleByPrefix(_availableLocales_, « _requestedLocale_ »).
-          1. If _match_ is not *undefined*, let _locale_ be _match_.[[locale]]; else let _locale_ be *"und"*.
+          1. Let _opt_ be the Record { [[localeMatcher]]: *"best fit"* }.
+          1. Let _relevantExtensionKeys_ be a new empty List.
+          1. Let _r_ be ResolveLocale(_availableLocales_, _requestedLocales_, _opt_, _relevantExtensionKeys_).
+          1. Let _locale_ be _r_.[[Locale]].
           1. Let _codePoints_ be StringToCodePoints(_S_).
           1. If _targetCase_ is ~lower~, then
             1. Let _newCodePoints_ be a List whose elements are the result of a lowercase transformation of _codePoints_ according to an implementation-derived algorithm using _locale_ or the Unicode Default Case Conversion algorithm.
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -227,7 +227,7 @@
           _requestedLocales_: a Language Priority List,
           _options_: a Record,
           _relevantExtensionKeys_: a List of Strings,
-          _localeData_: a Record,
+          optional _localeData_: a Record,
         ): a Record
       </h1>
       <dl class="header">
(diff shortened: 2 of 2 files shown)


anba:
> * Do not ignore locales after the first in the list returned by CanonicalizeLocaleList(_locales_) (observable via e.g. `"I".toLocaleLowerCase(["zzz", "tr"]) === "ı"`).

But also `"I".toLocaleLowerCase(["en", "tr"]) === "ı"`, because "en" will generally don't have locale-sensitive case mappings, which means the next locale in the list gets selected.

> * Match against the items of that list by best-fit rather than prefix, aligning with the rest of ECMA-402 (although the difference is not necessarily observable).

"best fit" matching isn't supported in browsers. (V8 has "--harmony-intl-best-fit-matcher", but that's not available by default.)

> * When the list is not empty but no matching locale is found, default to DefaultLocale() rather than "und", aligning with empty-list behavior and with the rest of ECMA-402 [...]

That means `"I".toLocaleLowerCase("und")` can now return either `"i"` or `"ı"`, depending on the user-locale.

anba:
> Ah yeah, I guess the Available Locales List needs to include more than just locale identifiers with language-sensitive case mappings. Any thoughts on what it should be? Mayble %Intl.Collator%.[[SortLocaleData]]?

If I had to guess, I'd say one of the reasons locale case conversion works differently from the other APIs, is that it's difficult to find an appropriate Available Locales list. I'm not sure if Intl.Collator is a good fit. 


> That's irrelevant, because [LookupMatchingLocaleByBestFit](https://tc39.es/ecma402/#sec-lookupmatchinglocalebybestfit) is defined to produce results "at least as good as those produced by the LookupMatchingLocaleByPrefix algorithm" (and therefore any implementation is free to just reuse LookupMatchingLocaleByPrefix).

I had assumed "the difference is not necessarily observable" was in reference to actual browser behaviour. If we assume an implementation that supports "best fit", which most likely uses the data from <https://github.com/unicode-org/cldr/blob/main/common/supplemental/languageInfo.xml>, then it's possible to have observable differences. There are three relevant entries:
```xml
<languageMatch desired="ku"	supported="tr"	distance="30"	oneway="true"/>
<languageMatch desired="azb" supported="az" distance="10" oneway="true"/>
<languageMatch desired="az"	supported="ru"	distance="30"	oneway="true"/>
```

That means `"I".toLocaleLowerCase("ku")` may fallback to `"I".toLocaleLowerCase("tr")`, because there's the fallback "ku" &rarr; "tr". For example V8 doesn't ship locale data for Kurdish (`Intl.Collator.supportedLocalesOf("ku")` returns the empty array), so if V8 started to officially support the "best fit" matcher, but string conversion is tied to the Intl.Collator Availables Locales, then `"I".toLocaleLowerCase("ku")` could start to return the dot-less i (U+0131).  

> I believe that would also be addressed via the Available Locales List provided to ResolveLocale, but regardless should align with other Intl services in general and Intl.Collator in particular. Probably, "und" should just always be considered available, but definitely should not be given special treatment exclusively in TransformCase.

I gave "und" as a special case, because at least for programmers with a Java background, using "und" shouldn't be too uncommon. (Java's String case conversion methods use `java.util.Locale.getDefault()` by default, which can result in bugs when the default locale is Turkish/Azeri. Instead it's necessary to use `str.toLowerCase(Locale.ROOT)`.)

gibson042:
Moving to draft because the necessary list of "case-mappable languages" is not conveniently available in implementations.

sffc:
TG2 discussion in May 2025: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-05-08.md#normative-update-string-tolocaleloweruppercase-to-resolvelocale-with-best-fit-matching-956

sffc:
There is a pre-existing CLDR issue to add data for which languages should use Turkic case folding:

https://unicode-org.atlassian.net/browse/CLDR-16202

Issue #461, opened 2020-06-18
Cannot output only era or timeZoneName


FrankYFTang:
Currently, we can use the option in Date.toLocale*String or Intl.DateTimeFormat to output most of the property in "Table 6: Components of date and time formats"
https://ecma-international.org/ecma-402/#datetimeformat-objects

```
d8> (new Date()).toLocaleString("en", {weekday: "long"})
"Thursday"
d8> (new Date()).toLocaleString("en", {year: "numeric"})
"2020"
d8> (new Date()).toLocaleString("en", {month: "long"})
"June"
d8> (new Date()).toLocaleString("en", {day: "numeric"})
"18"
d8> (new Date()).toLocaleString("en", {hour: "numeric"})
"12 AM"
d8> (new Date()).toLocaleString("en", {hour: "numeric"})
"12 AM"
d8> (new Date()).toLocaleString("en", {minute: "numeric"})
"3"
d8> (new Date()).toLocaleString("en", {second: "numeric"})
"51"
```

But if we put era or timeZoneName into the option, it will output more than just era / timeZoneName due to the ToDateTimeOptions
```
d8> (new Date()).toLocaleString("en", {era: "long"})
"6 18, 2020 Anno Domini, 12:01:54 AM"
d8> (new Date()).toLocaleString("en", {timeZoneName: "long"})
"6/18/2020, 12:04:05 AM Pacific Daylight Time"
```

Should we change ToDateTimeOptions to consider also era and timeZoneName so we can only output era or timeZoneName as

d8> (new Date()).toLocaleString("en", {era: "long"})
"Anno Domini"
d8> (new Date()).toLocaleString("en", {timeZoneName: "long"})
"Pacific Daylight Time"

FrankYFTang:
@sffc @anba @zbraniecki @littledan @rkirsling @gibson042

anba:
A workaround to get the time zone resp. era name is to invoke `formatToParts`:
```js
js> new Intl.DateTimeFormat("en", {timeZoneName:"long"}).formatToParts().filter(({type}) => type === "timeZoneName")[0].value               
"Pacific Daylight Time"
js> new Intl.DateTimeFormat("en", {era:"long"}).formatToParts().filter(({type}) => type === "era")[0].value                   
"Anno Domini"
```

I wonder if there are any meeting notes or es-discuss mails why "era" and "timeZoneName" aren't allowed as standalone elements in ToDateTimeOptions.

[...]

PR #957, opened 2025-02-06
Normative: Don't add default formatting to lone era or timeZoneName

ptomato:
Previously, if `era` or `timeZoneName` were given by themselves in the formatting options, the default formatting for the other date/time components would be added. This is inconsistent with what we do for `dayPeriod` where a lone `dayPeriod` produces formats such as "in the afternoon".

The use case of obtaining the localized era name has also been mentioned several times.

This change makes a lone `era` or `timeZoneName` be passed to the format matcher algorithm without added date/time components.

Closes: #461


anba:
Does this require any changes in <https://tc39.es/ecma402/#sec-intl.datetimeformat-internal-slots>? Right now neither `era` nor `timeZoneName` are required to be present in the minimum available formats. That means for example when requesting `{era: "long"}`, it's valid for [BasicFormatMatcher](https://tc39.es/ecma402/#sec-basicformatmatcher) to return the format for "hour, minute" (assuming only the minimum required formats are present in the implementation.)

When given the request `{era: "long"}`, BasicFormatMatcher will compute the best score for:
- year, month
- month, day
- hour, minute
- dayPeriod, hour

There's a removal penalty because `era` is not present, and twice an addition penalty, because other fields are present. So the best score is -120 - 2*20 = -160. The formats list is unordered, so any of these four formats can be returned.

ptomato:
I assumed no, because we also permit `dayPeriod` by itself but dayPeriod by itself is not part of the minimum required formats. (Although, maybe this is slightly different because dayPeriod _is_ present with other fields.)

anba:
> (Although, maybe this is slightly different because dayPeriod _is_ present with other fields.)

Yes, this is the important difference. Standalone `dayPeriod` is not required to be an available format, but at least it's guaranteed that a format is returned which includes `dayPeriod` when requesting standalone `dayPeriod`. Standalone `era` or standalone `timeZoneName` can return formats which contain neither `era` nor `timeZoneName`.

spec/datetimeformat.html
```diff
@@ -141,11 +141,11 @@ <h1>
         1. Else,
           1. Let _needDefaults_ be *true*.
           1. If _required_ is ~date~ or ~any~, then
-            1. For each property name _prop_ of « *"weekday"*, *"year"*, *"month"*, *"day"* », do
+            1. For each property name _prop_ of « *"weekday"*, *"era"*, *"year"*, *"month"*, *"day"* », do
               1. Let _value_ be _formatOptions_.[[&lt;_prop_&gt;]].
               1. If _value_ is not *undefined*, set _needDefaults_ to *false*.
           1. If _required_ is ~time~ or ~any~, then
-            1. For each property name _prop_ of « *"dayPeriod"*, *"hour"*, *"minute"*, *"second"*, *"fractionalSecondDigits"* », do
+            1. For each property name _prop_ of « *"dayPeriod"*, *"hour"*, *"minute"*, *"second"*, *"fractionalSecondDigits"*, *"timeZone"* », do
```
ptomato, line 148:
```suggestion
            1. For each property name _prop_ of « *"dayPeriod"*, *"hour"*, *"minute"*, *"second"*, *"fractionalSecondDigits"*, *"timeZoneName"* », do
```

ptomato:
Conclusion in the TG2 meeting today was:
- This should be split up into two PRs and `era` added separately from `timeZoneName`
- In both cases research should be done to check if it is likely to be web compatible
- If not web compatible, consider other options like `{ era: 'long', standalone: true }`

ptomato:
I've updated this to modify only `era`. I will open a separate PR for `timeZoneName`.

ryzokuken (approved):

sffc:
TG2 discussion on 2025-02-06: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-02-06.md#normative-dont-add-default-formatting-to-lone-era-or-timezonename-957

sffc:
Although I would love to see this, I propose that we close this PR given that the following code has worked since Intl 1.0 and seems like a reasonable thing to write:

```javascript
new Intl.DateTimeFormat("en", { era: "long" }).format(new Date())
// '6/9/2025 Anno Domini'
```

I think we should consider a different shape instead of redefining the current shape.

sffc:
Era was an option in Intl 1.0:

https://402.ecma-international.org/1.0/index.html#sec-12.1.1.1

But even at that time, it was not not included in the list in question.

sffc:
As in https://github.com/tc39/ecma402/pull/958#issuecomment-3085694374, I'm seeking agreement on the following proposed conclusion:

> We feel that the proposed behavior is better, but we are skeptical that making this change would not break the web. If we see evidence to the contrary, we can revisit. But, we are confident enough in our skepticism that we intend to close this pull request and explore other avenues for supporting standalone time zone and era formatting, which was the original motivation for the pull request.

nciric:
I am pretty sure we didn't make a deliberate decision around the era: long that would produce the stated result.
This could be a bug in the implementation, where if none of year, month, day values are not specified we revert to some default object.

The only discussion relevant to dates was with Microsoft, which at the time couldn't support more than 5 specific skeletons. Thus the spec was left open for "implementer discretion on whether more than 5 skeletons are provided".

PR #958, opened 2025-02-07
Normative: Don't add default formatting to lone timeZoneName

ptomato:
Previously, if `timeZoneName` was given by itself in the formatting options, the default formatting for the other date/time components would be added. This is inconsistent with what we do for `dayPeriod` where a lone `dayPeriod` produces formats such as "in the afternoon".

The use case of obtaining the localized time zone name in all its possible variants such as specific/generic, has been mentioned several times.

This change makes a lone `timeZoneName` be passed to the format matcher algorithm without added date/time components, and requires implementations to have at least a year-month-day-hour-minute-second-timeZoneName format available.

See: #461

### DIFF
--- a/spec/datetimeformat.html
+++ b/spec/datetimeformat.html
@@ -145,7 +145,7 @@
               1. Let _value_ be _formatOptions_.[[&lt;_prop_&gt;]].
               1. If _value_ is not *undefined*, set _needDefaults_ to *false*.
           1. If _required_ is ~time~ or ~any~, then
-            1. For each property name _prop_ of « *"dayPeriod"*, *"hour"*, *"minute"*, *"second"*, *"fractionalSecondDigits"* », do
+            1. For each property name _prop_ of « *"dayPeriod"*, *"hour"*, *"minute"*, *"second"*, *"fractionalSecondDigits"*, *"timeZoneName"* », do
               1. Let _value_ be _formatOptions_.[[&lt;_prop_&gt;]].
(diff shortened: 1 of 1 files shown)


ptomato:
See #957 for a similar change for `era`.

ryzokuken (approved):

sffc:
I believe that the following code has worked since Intl 1.0, and it seems like a reasonable thing to write:

```javascript
new Date().toLocaleTimeString("en", { timeZoneName: "long" })
'12:56:33 PM Pacific Daylight Time'
```

sffc:
Yes, I agree that I think this was a questionable design choice. It's only that the bar for changing a feature that has been working for 10 years is high, and "it seems like a reasonable thing to write" is unlikely to meet that bar. It would have to be "this seems extremely unlikely to have happened and here's some data to back that up." 😔

ptomato:
It's somewhere on my to-do list to investigate if there is such data, but it's not a high priority right now.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-07-17.md#normative-dont-add-default-formatting-to-lone-timezonename-958

I'm seeking agreement on the following proposed conclusion:

> We feel that the proposed behavior is better, but we are skeptical that making this change would not break the web. If we see evidence to the contrary, we can revisit. But, we are confident enough in our skepticism that we intend to close this pull request and explore other avenues for supporting standalone time zone and era formatting, which was the original motivation for the pull request.

PR #960, opened 2025-02-14
Normative: Add Intl.Locale.prototype.variants

### LINKED ISSUE #900: Why is there no `Intl.Locale.prototype.variants`?
Why is there no `Intl.Locale.prototype.variants`? There are getters for `language`, `region` and `script` but I saw no information about the reason `variants` is missing or shouldn't be there as well.

gibson042:
Fixes #900

test262 PR: https://github.com/tc39/test262/pull/4474

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -217,20 +217,6 @@
       </emu-alg>
     </emu-clause>
 
-    <emu-clause id="sec-Intl.Locale.prototype.getVariants">
-      <h1>Intl.Locale.prototype.getVariants ( )</h1>
-
-      <emu-alg>
-        1. Let _loc_ be the *this* value.
-        1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
-        1. Let _variantsString_ be GetLocaleVariants(_loc_.[[Locale]]).
-        1. If _variantsString_ is *undefined*, then
-          1. Return ! ArrayCreate(0).
-        1. Let _variantsList_ be StringSplitToList(_variantsString_, *"-"*).
-        1. Return CreateArrayFromList(_variantsList_).
-      </emu-alg>
-    </emu-clause>
-
     <emu-clause id="sec-Intl.Locale.prototype.hourCycle">
       <h1>get Intl.Locale.prototype.hourCycle</h1>
       <p>`Intl.Locale.prototype.hourCycle` is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:</p>
@@ -321,6 +307,16 @@
         1. Let _loc_ be the *this* value.
         1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
         1. Return _loc_.[[Locale]].
+      </emu-alg>
+    </emu-clause>
+
+    <emu-clause id="sec-Intl.Locale.prototype.variants">
+      <h1>get Intl.Locale.prototype.variants</h1>
+      <p>`Intl.Locale.prototype.variants` is an accessor property whose set accessor function is *undefined*. Its get accessor function performs the following steps:</p>
+      <emu-alg>
+        1. Let _loc_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
+        1. Return GetLocaleVariants(_loc_.[[Locale]]).
       </emu-alg>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


spec/locale.html
```diff
@@ -277,6 +277,18 @@ <h1>get Intl.Locale.prototype.collation</h1>
...
 
+    <emu-clause id="sec-Intl.Locale.prototype.getVariants">
+      <h1>Intl.Locale.prototype.getVariants ( )</h1>
+
+      <emu-alg>
+        1. Let _loc_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
+        1. Let _variantsString_ be GetLocaleVariants(_loc_.[[Locale]]).
+        1. Let _variantsList_ be StringSplitToList(_variantsString_, *"-"*).
+        1. Return CreateArrayFromList(_variantsList_).
```
FrankYFTang, line 288: + @anba Should this also state the order need to be in alphabetical order ?
See the order discussion in 
https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers

FrankYFTang:
If we do this, then how about setting variants in option bag them?

spec/locale.html
gibson042: It might benefit from a note, but nothing more—alphabetization is already required by use of [CanonicalizeUnicodeLocaleId](https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid) in the [Intl.Locale constructor](https://tc39.es/ecma402/#sec-Intl.Locale).

gibson042:
> If we do this, then how about setting variants in option bag them?

Good question, and I think the answer is "yes". This also interacts with the design question about representing variants as an array of strings or as a single dash-separated string, and I'm increasingly inclined towards the latter.

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-03-06.md#normative-add-localeprototypegetvariants-960

sffc:
TG2 discussion, focused on the motivation: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-04-03.md#normative-add-localeprototypegetvariants-960

spec/locale.html
```diff
@@ -277,6 +277,18 @@ <h1>get Intl.Locale.prototype.collation</h1>
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-Intl.Locale.prototype.getVariants">
+      <h1>Intl.Locale.prototype.getVariants ( )</h1>
+
+      <emu-alg>
+        1. Let _loc_ be the *this* value.
+        1. Perform ? RequireInternalSlot(_loc_, [[InitializedLocale]]).
+        1. Let _variantsString_ be GetLocaleVariants(_loc_.[[Locale]]).
```
eemeli, line 286: This can be `undefined`, which is not supported by `StringSplitToList()`.
```suggestion
        1. Let _variantsString_ be GetLocaleVariants(_loc_.[[Locale]]).
        1. If _variantsString_ is *undefined*, then
          1. Return ! ArrayCreate(0).
```

eemeli (changes requested):

eemeli (approved):

gibson042:
In accord with https://github.com/tc39/ecma402/pull/960#issuecomment-2704732439 and TG2 discussion, this will be exposed as a string-returning `variants` getter to align with the other subtag-based properties (`language`, `baseName`, `script`, `region`, `calendar`, etc.).

ryzokuken (approved):

spec/locale.html
```diff
@@ -87,7 +87,9 @@ <h1>
         1. Let _region_ be ? GetOption(_options_, *"region"*, ~string~, ~empty~, GetLocaleRegion(_baseName_)).
         1. If _region_ is not *undefined*, then
           1. If _region_ cannot be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
-        1. Let _variants_ be GetLocaleVariants(_baseName_).
+        1. Let _variants_ be ? GetOption(_options_, *"variants"*, ~string~, ~empty~, GetLocaleVariants(_baseName_)).
+        1. If _variants_ is not *undefined*, then
+          1. If _variants_ cannot be matched by the <code>unicode_variant_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
```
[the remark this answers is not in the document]
sffc: #1006

sffc:
TG1 notes from 2025 May: https://github.com/tc39/notes/blob/main/meetings/2025-05/may-28.md#normative-add-intllocaleprototypevariants-ecma402960

PR #966, opened 2025-03-02
Editorial: Document DurationFormat in API Overview, Well-Known Intrinsic Objects, and Implementation Dependent Behaviour

gibson042:
Ref #943


ryzokuken:
Sorry for not realizing this was missing in the original PR and thanks!

PR #963, opened 2025-03-01
Editorial: Clean up DurationFormat

gibson042:
* `emu-format`
* align section order with #923
* replace outdated use of `@@toStringTag` and "abrupt completion"
* improve documentation of internal slots
* simplify `resolvedOptions` and `GetDurationUnitOptions`

PR #969, opened 2025-03-03
Editorial: Fix spec link for and remove unused parameter from new ValidateDurationUnitStyle AO

trflynn89:


spec/durationformat.html
```diff
@@ -648,10 +648,9 @@ <h1>
           }.
       </emu-alg>
 
-      <emu-clause id="sec-stringpaddingbuiltinsimpl" type="abstract operation">
+      <emu-clause id="sec-validatedurationunitstyle" type="abstract operation">
```
gibson042, line 651: Thanks @trflynn89; I've merged this fix as #970.

spec/durationformat.html
```diff
@@ -648,10 +648,9 @@ <h1>
...
         <h1>
           ValidateDurationUnitStyle (
-            _unit_: a String,
```
gibson042, line 654: Inclusion of a seemingly-irrelevant _unit_ parameter was intentional; I expect its contents to appear in the message of any thrown exception just as it would have before being factored out of GetDurationUnitOptions in https://github.com/tc39/ecma402/pull/963/commits/20daf2630a9e187c51373e8ebdeb7ef1fcab77f9. But this is obviously too subtle... perhaps, rather than dropping the parameter, we should have a note explaining how it might be used?

trflynn89: Ah gotcha, yeah a note here would be helpful for clarity. In Ladybird's implementation of these changes, I've actually included a couple of extra parameters to this AO to create useful exception messages, so I'm not sure that `unit` being treated specially by the spec is all that necessary for implementers.

gibson042: #975

trflynn89: Thanks! I'll go ahead and close this PR

PR #977, opened 2025-03-05
Editorial: Remove unused unit parameter from ValidateDurationUnitStyle

linusg:
This feels like something ecmarkup should lint for.


ptomato:
See https://github.com/tc39/ecma402/pull/975

linusg:
Ah, fair enough :)

PR #978, opened 2025-03-05
Editorial: Simplify NumberFormat

gibson042:
* Fix `id` values to align with operation names.
* Refactor spec aliases and parameter types for readability.
* Collapse rounding mode negation data into one table.
* Fold GetUnsignedRoundingMode into ApplyUnsignedRoundingMode

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -36,6 +36,14 @@ emu-normative-optional:before {
 #ecma-logo {
   width: 500px;
 }
+#spec-container {
+  container: spec-container / inline-size;
+}
+@container spec-container (width >= 768px) {
+  .c-md-width {
+    width: var(--c-md-width);
+  }
+}
 </style>
 <pre class="metadata">
   title: ECMAScript&reg; 2025 Internationalization API Specification
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -643,7 +643,6 @@
         </tr>
       </table>
       <emu-note>
-        The Negation column is used by GetUnsignedRoundingMode.
         When the rounding mode in the Identifier column produces outputs of the same magnitude for all inputs of the same magnitude, it is its own negation and the Negation column is blank (e.g., abs(truncate(-1.5)) = abs(truncate(1.5)) = 1).
         Otherwise, the Negation column contains the rounding mode that should be applied to the absolute value of a negative number to produce a result of the same magnitude as application of the rounding mode in the Identifier column to its signed value (e.g., abs(ceiling(-1.5)) ≠ abs(floor(1.5)), but abs(ceiling(-1.5)) = abs(floor(1.5)) = 1).
       </emu-note>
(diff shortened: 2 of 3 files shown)


ptomato:
Would you have time to update the usages of GetUnsignedRoundingMode and ApplyUnsignedRoundingMode in Temporal while this is still fresh?

sffc (changes requested):
Hmm. I quite like having a clean separation between RoundingMode (what users specify) and UnsignedRoundingMode (what implementations need). I'm not convinced that essentially replacing the internal enum `UnsignedRoundingMode` with an internal tuple that pairs RoundingMode with `~positive~` or `~negative~` is an improvement.

gibson042:
> Hmm. I quite like having a clean separation between RoundingMode (what users specify) and UnsignedRoundingMode (what implementations need).

I might agree with you if we didn't make comparisons against the former all over the place, including in many steps in or downstream of [NumberFormat PartitionNumberPattern](https://tc39.es/ecma402/#sec-partitionnumberpattern) (_numberFormat_.[[Style]] vs. **"percent"**/**"unit"**/**"currency"**, _intlObject_.[[TrailingZeroDisplay]] vs. **"stripIfInteger"**, _numberFormat_.[[SignDisplay]] vs. **"never"**/**"auto"**/**"always"**/**"exceptZero"**/**"negative"**, _numberFormat_.[[Notation]] vs. **"scientific"**/**"engineering"**/**"compact"**, etc.).

> I'm not convinced that essentially replacing the internal enum `UnsignedRoundingMode` with an internal tuple that pairs RoundingMode with `~positive~` or `~negative~` is an improvement.

Well, that's exactly what [GetUnsignedRoundingMode](https://tc39.es/ecma402/#sec-getunsignedroundingmode) does right now. The benefits of instead directly propagating that pair into ToRaw{Precision,Fixed} and ApplyUnsignedRoundingMode are
* Make clear in the definition of rounding modes that only the ceil/floor family need this adjustment when working with absolute values, defining details about the rounding modes in one place.
* Avoid essentially duplicating that table a huge distance away.
* Avoid needless introduction of new entities that basically duplicate the rounding modes anyway.

gibson042:
> https://unicode-org.github.io/icu4x/rustdoc/fixed_decimal/enum.UnsignedRoundingMode.html

I don't have the context to evaluate how (or if) that might be different if the spec had already looked like I'm proposing in this PR, but the "Comparative table of all the rounding modes" at that link includes all 9 user-specifiable ECMA-402 rounding modes and not the current spec enum values produced by GetUnsignedRoundingMode (which makes sense, because the latter are redundant with the former—<tt>INFINITY</tt>/<tt>HALF-INFINITY</tt> is just "{ceil,expand}"/"half{Ceil,Expand}", <tt>ZERO</tt>/<tt>HALF-ZERO</tt> is just "{floor,trunc}"/"half{Floor,Trunc}", and <tt>HALF-EVEN</tt> is exactly "halfEven"—due to the respective equivalence of "ceil" vs. "expand" and "floor" vs. "trunc" when limited to non-negative input).

sffc (commented):
There are two parts of this PR. The first 4 commits look fine to me. I would approve those as a separate PR.

The last 4 commits are the ones that change `UnsignedRoundingMode` to essentially a tuple of `RoundingMode` and "ordinal sign". I need to think more about this.

sffc:
> > https://unicode-org.github.io/icu4x/rustdoc/fixed_decimal/enum.UnsignedRoundingMode.html
> 
> I don't have the context to evaluate how (or if) that might be different if the spec had already looked like I'm proposing in this PR, but the "Comparative table of all the rounding modes" at that link includes all 9 user-specifiable ECMA-402 rounding modes and not the current spec enum values produced by GetUnsignedRoundingMode

The spec table is in terms of the signed rounding mode because that is the table in the spec. It should probably be on the signed rounding mode enum in the docs. The docs themselves acknowledge this and link back and forth between the rounding mode enums.

---

My qualm is that, based on implementer experience, both recently and less recently, floating point and decimal numbers (the types NumberFormat deals with) are stored unsigned, with a sign bit on the side. So as far as we are concerned, there are only 5 rounding modes, the unsigned ones.

sffc:
We've agreed on a path forward that improves this PR. @gibson042 intends to push a fix.

PR #984, opened 2025-03-30
Editorial: Add an operation abstracting service constructor locale resolution

### LINKED ISSUE #405: ECMA-402 contents should be arranged more consistently

gibson042:
Fixes #405

### DIFF
--- a/spec/collator.html
+++ b/spec/collator.html
@@ -13,6 +13,7 @@
         1. If NewTarget is *undefined*, let _newTarget_ be the active function object, else let _newTarget_ be NewTarget.
         1. Let _internalSlotsList_ be « [[InitializedCollator]], [[Locale]], [[Usage]], [[Collation]], [[Numeric]], [[CaseFirst]], [[Sensitivity]], [[IgnorePunctuation]], [[BoundCompare]] ».
         1. Let _collator_ be ? OrdinaryCreateFromConstructor(_newTarget_, *"%Intl.Collator.prototype%"*, _internalSlotsList_).
+        1. NOTE: The source of locale data for ResolveOptions depends upon the *"usage"* property of _options_, but the following two steps must observably precede that lookup (and must not observably repeat inside ResolveOptions).
         1. Let _requestedLocales_ be ? CanonicalizeLocaleList(_locales_).
         1. Set _options_ to ? CoerceOptionsToObject(_options_).
         1. Let _usage_ be ? GetOption(_options_, *"usage"*, ~string~, « *"sort"*, *"search"* », *"sort"*).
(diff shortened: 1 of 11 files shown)


spec/datetimeformat.html
```diff
@@ -58,25 +58,14 @@ <h1>
...
-        1. Let _hour12_ be ? GetOption(_options_, *"hour12"*, ~boolean~, ~empty~, *undefined*).
-        1. Let _hourCycle_ be ? GetOption(_options_, *"hourCycle"*, ~string~, « *"h11"*, *"h12"*, *"h23"*, *"h24"* », *undefined*).
-        1. If _hour12_ is not *undefined*, then
-          1. Set _hourCycle_ to *null*.
-        1. Set _opt_.[[hc]] to _hourCycle_.
-        1. Let _r_ be ResolveLocale(%Intl.DateTimeFormat%.[[AvailableLocales]], _requestedLocales_, _opt_, %Intl.DateTimeFormat%.[[RelevantExtensionKeys]], %Intl.DateTimeFormat%.[[LocaleData]]).
+        1. Let _hour12_ be *undefined*.
+        1. Let _modifyResolutionOptions_ be a new Abstract Closure with parameters (_options_) that captures _hour12_ and performs the following steps when called:
+          1. Set _hour12_ to _options_.[[hour12]].
+          1. Remove field [[hour12]] from _options_.
```
gibson042, line 64: I'm _this close_ to finally replacing [_record_.[[\<_var_>]]](https://tc39.es/ecma402/#conventions) with a Mapping specification type and corresponding Has/Get/Set/Remove operations.

spec/collator.html
```diff
@@ -25,40 +22,18 @@ <h1>Intl.Collator ( [ _locales_ [ , _options_ ] ] )</h1>
...
-        1. Let _sensitivity_ be ? GetOption(_options_, *"sensitivity"*, ~string~, « *"base"*, *"accent"*, *"case"*, *"variant"* », *undefined*).
-        1. If _sensitivity_ is *undefined*, then
-          1. If _usage_ is *"sort"*, then
-            1. Set _sensitivity_ to *"variant"*.
-          1. Else,
-            1. Set _sensitivity_ to _resolvedLocaleData_.[[sensitivity]].
-        1. Set _collator_.[[Sensitivity]] to _sensitivity_.
+        1. If _usage_ is *"sort"*, let _defaultSensitivity_ be *"variant"*. Otherwise, let _defaultSensitivity_ be _resolvedLocaleData_.[[sensitivity]].
+        1. Set _collator_.[[Sensitivity]] to ? GetOption(_options_, *"sensitivity"*, ~string~, « *"base"*, *"accent"*, *"case"*, *"variant"* », _defaultSensitivity_).
         1. Let _defaultIgnorePunctuation_ be _resolvedLocaleData_.[[ignorePunctuation]].
```
ryzokuken, line 35: Can this line be collapsed into the next one as well?

ryzokuken (approved):
Looks great, thanks! One odd nit inline.

spec/collator.html
gibson042: Yes, but I think the _defaultIgnorePunctuation_ name is worth keeping (especially as an analog to _defaultSensitivity_).

ryzokuken: Yeah, that's fair enough. Alright then I'll take a quick skim yet again today given the size of this patch to make sure everything looks alright once again and hit merge.

PR #986, opened 2025-04-02
Meta: fix watch script

ryzokuken:
At some point, ecmarkup stopped supporting --watch and --strict at the same time. This commit updates the watch script to not use --strict to make it work again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "npm run build-only -- --lint-spec --strict",
     "build-for-pdf": "npm run build -- --old-toc",
     "test": "npm run build-to -- --lint-spec /dev/null",
-    "watch": "npm run build -- --watch"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "npm run build-only -- --lint-spec --strict",
     "build-for-pdf": "npm run build -- --old-toc",
     "test": "npm run build-to -- --lint-spec /dev/null",
```
gibson042, line 13: Might as well fix `test` while we're at it.
```suggestion
    "test": "npm run build-to -- /dev/null --lint-spec --strict",
```

gibson042 (approved):
:+1:

package.json
ryzokuken: Both versions work for me but yeah seems like a change I'm more than happy to take!

PR #992, opened 2025-04-07
Editorial: Pass `%Intl.Segmenter%` to its invocation of `ResolveOptions`

trflynn89:
We do not want to use Intl.NumberFormat's internal slots here.


ryzokuken (approved):
This looks like a spec bug that made it, should've caught it in the review. Thanks a lot @trflynn89.

Issue #536, opened 2021-02-11
API naming conventions should be documented


gibson042:
We should explicitly document naming conventions, and maintain them by evaluating new proposals against them.

* Property names and values should avoid abbreviations.
  * Good examples: "fractionalSeconds", "minimumSignificantDigits"
* Property names and values that are compound words should combine them using camelCase.
  * Good examples: "signDisplay", "exceptZero"
  * Exception: for values only, nonzero numbers should be expressed as digits (as in "h23")
  * Historical exceptions: "2-digit", "best fit"
* [open question about "[...]style" vs. "[...]display" for naming high-level configuration properties]

sffc:
Yes.  We have a start in https://github.com/tc39/ecma402/blob/master/docs/style-guide.md, and we should expand that document to cover additional conventions.

sffc:
@ben-allen to make a PR adding more of the content from the OP in time to discuss at the October TG2 meeting.

PR #996, opened 2025-04-22
Meta: Add naming conventions section in style guide

ryzokuken:
Fixes: #536 

Took a stab at it, let me know what you think about the phrasing.


docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+
+This section provides guidelines for naming conventions in ECMA-402, focusing on clarity, consistency, and adherence to established practices.
+
+### Avoiding Abbreviations in Property Names and Values
+
+:star2: *Property names and values in ECMA-402 should avoid abbreviations to ensure clarity and readability. Use descriptive names that clearly convey their purpose.* :star2:
+
+#### Guidelines
+
+- Use full, descriptive names for properties and values, even if they are longer.
```
gibson042, line 245:
```suggestion
- Use full, descriptive names for properties and values, even if they are longer (good examples: `fractionalSecondDigits`, `minimumSignificantDigits`).
```

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+- Abbreviations may be acceptable if they are widely recognized and standardized (e.g., `UTC`, `ISO`).
+- When abbreviations are used, they should be documented thoroughly to avoid confusion.
+
+### Compound Words in Property Names and Values
+
+:star2: *Combine compound words in property names and values using camelCase.* :star2:
+
+#### Guidelines
+
+- Use camelCase for property names and values that are compound words.
```
gibson042, line 266:
```suggestion
- Use camelCase for property names and values that are compound words (good examples: `signDisplay`, `"exceptZero"`).
```

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+- Avoid separating words with underscores, hyphens, or other delimiters.
```
gibson042, line 267:
```suggestion
- Avoid separating words with underscores, hyphens, or other delimiters (historical exceptions: "2-digit", "best fit").
```

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+- For values only, nonzero numbers should be expressed as digits (as in `"h23"`).
+
+#### Rationale
+
+- camelCase is a widely accepted convention in JavaScript and aligns with existing ECMA-402 practices.
+- Using digits for nonzero numbers in values ensures brevity and consistency with historical conventions.
+
+### Open Question: "[...]style" vs. "[...]display" for High-Level Configuration Properties
+
+There is an open question regarding the naming of high-level configuration properties: should we prefer the suffix `[...]style` or `[...]display`? They're both used relatively interchangably. We must analyze this problem space in TG2 and come up with a convention.
```
gibson042, line 277:
```suggestion
### Open Question: `Style` vs. `Display` for High-Level Configuration Properties

There is an open question regarding the naming of high-level configuration properties: should we prefer to name them like `…Style` or `…Display`? They're both used relatively interchangably (e.g., `style`/`dateStyle`/`timeStyle` vs. `compactDisplay`/ `signDisplay`/`unitDisplay`). We must analyze this problem space in TG2 and come up with a convention.
```

gibson042 (approved):
LGTM, but I'd also like @sffc to weigh in.

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+#### Guidelines
+
+- Use full, descriptive names for properties and values, even if they are longer.
+- Avoid cryptic abbreviations or shortened forms that may confuse readers or developers unfamiliar with the context.
+- Prioritize clarity over brevity, especially for properties that are frequently used or have significant impact.
+
+#### Rationale
+
+- Descriptive names improve code readability and maintainability.
+- They reduce the cognitive load for developers, especially those new to the specification.
```
sffc, line 252: Also, spelled-out names are more accessible to non-native English speakers, because they are more googlable, etc. We are building an i18n library, after all.

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+
+### Compound Words in Property Names and Values
+
+:star2: *Combine compound words in property names and values using camelCase.* :star2:
+
+#### Guidelines
+
+- Use camelCase for property names and values that are compound words.
+- Avoid separating words with underscores, hyphens, or other delimiters.
+- For values only, nonzero numbers should be expressed as digits (as in `"h23"`).
```
sffc, line 268: A better example is "min2". Note that "h23" is a term that comes from UTS 35.

docs/style-guide.md
```diff
@@ -221,3 +231,47 @@ For an example of when *not* to use lexicographic order, consider an array holdi
...
+
+#### Rationale
+
+- camelCase is a widely accepted convention in JavaScript and aligns with existing ECMA-402 practices.
```
sffc, line 272: Please dig up the thread in NumberFormat V3 where this was discussed, but we had a number of other reasons. One was that sometimes these strings end up as identifiers, and identifiers in JavaScript are camel case.

docs/style-guide.md
ryzokuken: What about the immediately following point? I believe it comes close so at the very least I should add something along these lines to it (or after it).

docs/style-guide.md
ryzokuken: Is this it? https://github.com/tc39/proposal-intl-numberformat-v3/issues/9

sffc: Yeah, I thought that was the original issue but I guess it is https://github.com/tc39/proposal-intl-displaynames/issues/32

Actually, the style guide already says to use camel case here. There is a big section explaining the rationale. I think just refer to that previous section.

PR #997, opened 2025-04-23
Editorial: mark DefaultLocale as a fingerprinting vector

### LINKED ISSUE #110: Consider marking DefaultLocale as a fingerprinting vector
HTML marks fingerprinting vectors with [a big fingerprint icon](https://html.spec.whatwg.org/#fingerprinting-vector). Should we do the same with DefaultLocale in ECMA 402? This applies to #106 even more.

(Side note: How does DefaultLocale relate to navigator.locale? Should they be specified somewhere to be equal? Are they in current browsers?)

cc @domenic @foolip @annevk

ryzokuken:
Add a normative note to the DefaultLocale AO to indicate that it can be used as a fingerprinting vector in browser environments.

Fixes #110

### DIFF
--- a/spec/locales-currencies-tz.html
+++ b/spec/locales-currencies-tz.html
@@ -107,6 +107,9 @@
         <dt>description</dt>
         <dd>The returned String value represents the structurally valid (<emu-xref href="#sec-isstructurallyvalidlanguagetag"></emu-xref>) and canonicalized (<emu-xref href="#sec-canonicalizeunicodelocaleid"></emu-xref>) language tag for the host environment's current locale. It must not contain a Unicode locale extension sequence.</dd>
       </dl>
+      <emu-note>
+        The returned value is is a potential fingerprinting vector. In browser environments, it should match <a href="https://html.spec.whatwg.org/#language-preferences"><code>navigator.language</code></a> to avoid providing any additional distinguishing information.
+      </emu-note>
     </emu-clause>
   </emu-clause>
 


ryzokuken:
@sffc I tried taking a (rather minimal) stab at your comment at https://github.com/tc39/ecma402/issues/110#issuecomment-2091896616. Let me know how the phrasing feels.

sffc (commented):
This conveys the intent, I think. Mainly I want the spec to require that DefaultLocale derives from navigator.languages, such that DefaultLocale does not introduce any _new_ entropy.

spec/locales-currencies-tz.html
```diff
@@ -107,6 +107,9 @@ <h1>DefaultLocale ( ): a Unicode canonicalized locale identifier</h1>
         <dt>description</dt>
         <dd>The returned String value represents the structurally valid (<emu-xref href="#sec-isstructurallyvalidlanguagetag"></emu-xref>) and canonicalized (<emu-xref href="#sec-canonicalizeunicodelocaleid"></emu-xref>) language tag for the host environment's current locale. It must not contain a Unicode locale extension sequence.</dd>
       </dl>
+      <emu-note>
+        In browser environments, this value is derived from <code>navigator.languages</code>, which could potentially be used as a fingerprinting vector. In non-browser environments, it is generally determined by the host environment's locale settings.
```
gibson042, line 111: "derived from" doesn't seem right to me, rather `navigator.language` and DefaultLocale should agree with each other and **both** derive from the same root source.
```suggestion
        The returned value is is a potential fingerprinting vector. In browser environments, it should match <a href="https://html.spec.whatwg.org/#language-preferences"><code>navigator.language</code></a> to avoid providing any additional distinguishing information.
```

gibson042 (approved):

PR #1007, opened 2025-05-30
Editorial: Fix incorrect application of `unicode_variant_subtag` against a subtag sequence

gibson042:
Fixes #960


sffc (approved):
Thanks

gibson042:
Thank you for the thorough review, @anba.

spec/locale.html
```diff
@@ -89,7 +89,10 @@ <h1>
           1. If _region_ cannot be matched by the <code>unicode_region_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
         1. Let _variants_ be ? GetOption(_options_, *"variants"*, ~string~, ~empty~, GetLocaleVariants(_baseName_)).
         1. If _variants_ is not *undefined*, then
-          1. If _variants_ cannot be matched by the <code>unicode_variant_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
+          1. If _variants_ is the empty String, throw a *RangeError* exception.
+          1. For each element _variant_ of StringSplitToList(_variants_, *"-"*), do
+            1. If _variant_ cannot be matched by the <code>unicode_variant_subtag</code> Unicode locale nonterminal, throw a *RangeError* exception.
+          1. If _variants_ contains any duplicate subtags, throw a *RangeError* exception.
```
anba, line 95: Do we trust that readers know that subtags are to be compared case-insensitively? `IsStructurallyValidLanguageTag` explicitly converts to lower-case to ensure case-insensitive comparison.

gibson042: Hmm, good point. Clarified in #1009.

PR #1009, opened 2025-05-31
Editorial: Clarify that duplicate-variants checks are case-insensitive

gibson042:
Ref https://github.com/tc39/ecma402/pull/1007#pullrequestreview-2883658270


sffc:
Can you explain why this is editorial?

gibson042:
UpdateLanguageId is documented to return a structurally valid language tag, which prohibits duplicate variant subtags (case-insensitive). But @anba pointed out in https://github.com/tc39/ecma402/pull/1007#pullrequestreview-2883658270 that case-insensitivity was not obvious in context, so this PR made it explicit.

PR #1012, opened 2025-06-13
Optional merge: Edits to ready this edition for print(able) publication

gesa:
These are the changes that went into publishing the 2025 edition of ECMA-402. I suggest merging them into this branch, but that's editors' prerogative and regardless they will be here for posterity.


gesa:
Draft PDF https://hosted.akiro.se/2025/ecma/DRAFT-ECMA-402_12th_edition_June_2025.pdf

PR #1024, opened 2025-08-19
Typo fix: Intl.NumberFormat's SetNumberFormatDigitOptions AO

### LINKED ISSUE #1023: minor typo in the SetNumberFormatDigitOptions function.
In the SetNumberFormatDigitOptions function(https://402.ecma-international.org/#sec-setnumberformatdigitoptions)
...
18. If mnfd is undefined and mxsd is undefined, let hasFd be false. Otherwise, let hasFd be true.
...
IMO We need to check mxfd variable instead of mxsd variable here

implementations also check mxfd instead of mxsd
https://github.com/v8/v8/blob/10df7d05ba574bc8073f2445f3eaacd9bf434e8c/src/objects/intl-objects.cc#L1692
https://github.com/WebKit/WebKit/blob/8a194767935ccd544f80cc447e29cd5c1d28c035/Source/JavaScriptCore/runtime/IntlNumberFormatInlines.h#L108

ben-allen:
As the result of a typo in #796 (cf. 053a606ce2d1355be419891796fce31d7571acf9), _mxsd_ was checked rather than _mxfd_ when determining value of _hasFd_. This fixes that.

fix #1023

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -111,7 +111,7 @@
         1. Set _intlObj_.[[RoundingMode]] to _roundingMode_.
         1. Set _intlObj_.[[TrailingZeroDisplay]] to _trailingZeroDisplay_.
         1. If _mnsd_ is *undefined* and _mxsd_ is *undefined*, let _hasSd_ be *false*. Otherwise, let _hasSd_ be *true*.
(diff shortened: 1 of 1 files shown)


gibson042 (approved):

sffc:
I think this is a correct fix. There are many things in that function which are unintuitive though so we should be careful when editing it.

PR #1035, opened 2025-11-07
Normative: Add 1 new numbering system "tols" for Unicode 17

FrankYFTang:
There is one new set of decimal digits added in Unicode 17.0, for the newly encoded Tolong Siki script. Implementations of numeric values and numeric formatting should take this new set into account.

https://www.unicode.org/versions/Unicode17.0.0/

CLDR add that into CLDR 48 which will be available in ICU 78

This PR update the ECMA402 spec to include that new numbering system.

### DIFF
--- a/spec/numberformat.html
+++ b/spec/numberformat.html
@@ -1209,6 +1209,10 @@
             <td>U+16AC0 to U+16AC9</td>
           </tr>
           <tr>
+            <td>tols</td>
+            <td>U+11DE0 to U+11DE9</td>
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-01-08.md#normative-add-1-new-numbering-system-tols-for-unicode-17-1035

PR #1037, opened 2025-12-04
Editorial: "languageDisplay" can be undefined in Intl.DisplayNames.prototype.resolvedOptions

anba:
The [[LanguageDisplay]] slot can be `undefined`, so the assertion needs to be changed to an If-step.


sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2025-12-04.md#languagedisplay-can-be-undefined-in-intldisplaynamesprototyperesolvedoptions-1037

PR #1050, opened 2026-04-02
Editorial: Refactor SetNumberFormatDigitOptions for comprehensibility

gibson042:
Prompted by recent reviews of [proposal-amount](https://github.com/tc39/proposal-amount) PRs.


sffc (commented):
I agree that this is a pure refactor, no logic change.

I have a general preference for keeping the variable names the same; I don't see a reason to change them.

The main change here is to switch some if statements from

let variable be A
if B, then set variable to C

to

if B, let variable be C
else, let variable be A

This seems fine and probably beneficial.

PR #1057, opened 2026-04-15
Editorial: Check for undefined [[FirstDayOfWeek]] field

anba:
`[[FirstDayOfWeek]]` can be `undefined`, but `WeekdayUValueToNumber` expects String values. Perform the check for `undefined` in `WeekInfoOfLocale` to match other operations which allow Unicode extension overrides (`CalendarsOfLocale`, `CollationsOfLocale`, `HourCyclesOfLocale`, and `NumberingSystemsOfLocale`.)


sffc:
I think this is from Intl Locale Info. CC @FrankYFTang

FrankYFTang (approved):

ptomato (approved):

ryzokuken (approved):

PR #1059, opened 2026-04-15
Normative: Align RegionPreference and CanonicalUnicodeSubdivision with UTS 35

### LINKED ISSUE #1058: RegionPreference: Align with UTS 35?
[RegionPreference](https://tc39.es/ecma402/#sec-regionpreference) doesn't seem to match UTS 35, see also <https://unicode.org/reports/tr35/tr35.html#RegionOverride>. 

Issues:

1. [CanonicalUnicodeSubdivision](https://tc39.es/ecma402/#sec-canonicalunicodesubdivision) calls [CanonicalizeUnicodeLocaleId](https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid) to replace region aliases. It's unclear if this is valid per UTS 35. For example in [Processing LocaleIds](https://unicode.org/reports/tr35/tr35.html#processing-localeids), `unicode_subdivision_id` replaces subdivisions using `<subdivisionAlias>` without first canonicalizing the subdivision's `unicode_region_subtag` part. 

Example:
```js
js> new Intl.Locale("en-u-rg-826zzzz").getHourCycles()[0]
"h12"
js> new Intl.Locale("en-u-rg-gbzzzz").getHourCycles()[0]
"h23"
```

Per the current spec text, both calls should return `"h23"`. (`826` is the numeric region for `GB`.) 

2. [Region Override](https://unicode.org/reports/tr35/#RegionOverride) defines that only "regular" and "unknown" regions are allowed and any other value should be ignored. Validity is checked against <https://github.com/unicode-org/cldr/blob/main/common/validity/region.xml>. I'm not sure if the "unknown" region should be accepted, though, because at least ICU4C seems to ignore it. 

Examples:
```js
js> new Intl.Locale("en-JP-u-rg-001zzzz").getCalendars()
["gregory", "japanese"]
```

The macro-region `001` has [`<calendarPreference>` data](<https://github.com/unicode-org/cldr/blob/5ea4a87cfe26e7140e4b37aac0ab4a36691d9170/common/supplemental/supplementalData.xml#L4459>), but the region override is ignored nonetheless. 

```js
js> new Intl.Locale("en-JP-u-rg-zzzzzz").getCalendars()
["gregory", "japanese"]
js> new Intl.Locale("ja-u-rg-zzzzzz-sd-thzzzz").getCalendars()
["buddhist", "gregory"]
```

Region override uses the "unknown" region `ZZ`, but it's ignored and instead the calendar preference from `JP` resp. `TH` is used.

```js
js> new Intl.Locale("ja-u-sd-euzzzz").getCalendars()
["gregory", "japanese"]
js> new Intl.Locale("ja-u-sd-thzzzz").getCalendars()
["buddhist", "gregory"]
```

The subdivision using the macro-region `EU` is ignored and instead the calendar preference from `JP` (added through likely subtags) is used. When using a regular region (`TH`), the `sd` Unicode extension value is used. 

---

Existing (browser) implementations directly call into ICU4C and don't implement any of the ECMA-402 specific operations, so changing the spec won't cause any web-compatibility issues. And ICU4C seems to ignore both region aliases and non-regular regions.

---

ICU4C is only buggy when using an empty Unicode extension value:
```js
d8> new Intl.Locale("ja-u-rg").getCalendars()
["gregory", "islamic", "islamic-civil", "islamic-tbla"]
```

This probably happens because the empty string is treated as a shorthand for `yes` (either directly or by first replacing with the value `true`), and `yes` starts with the region code for `YE` (Yemen).

anba:
Changes in `CanonicalUnicodeSubdivision`:
- Don't canonicalize region subtags to replace aliases.
- Only allow valid regular region codes using validity data defined in UTS 35.

Changes in `RegionPreference`:
- Return the first region which can be matched, using the following preference order:
    1. Region override Unicode extension key `rg`.
    2. Region subtag in locale.
    3. Subdivision Unicode extenion key `sd`.
    4. Region subtag added through likely subtags.
    5. Fallback to region code `001`.

Fixes #1058

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -575,15 +575,17 @@
         ): a String or *undefined*
       </h1>
       <dl class="header">
+        <dt>description</dt>
+        <dd>The following algorithm refers to Region validity data specified in <a href="https://unicode.org/reports/tr35/#unicode_region_subtag_validity">Unicode Technical Standard #35 Part 1 Core, Language Identifier Field Definitions</a>.</dd>
       </dl>
       <emu-alg>
         1. Let _subdivision_ be UnicodeExtensionValue(_locale_, _key_).
         1. If _subdivision_ is ~empty~, return *undefined*.
         1. If _subdivision_ cannot be matched by the `unicode_subdivision_id` Unicode locale nonterminal, return *undefined*.
         1. Let _region_ be the longest prefix of _subdivision_ matched by the `unicode_region_subtag` Unicode locale nonterminal.
-        1. Let _regionLocale_ be the string-concatenation of *"und-"* and _region_.
-        1. Set _regionLocale_ to CanonicalizeUnicodeLocaleId(_regionLocale_).
-        1. Return GetLocaleRegion(_regionLocale_).
+        1. Set _region_ to the ASCII-uppercase of _region_.
+        1. If _region_ is not a valid regular region code, return *undefined*.
+        1. Return _region_.
       </emu-alg>
     </emu-clause>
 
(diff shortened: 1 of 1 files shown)


sffc:
TG2 would like @FrankYFTang and the engine implementers to take a look (cc @Constellation @dminor).

https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-05-19.md#anbas-normative-prs

FrankYFTang:
Could we put down some test cases to show the difference of before/after behavior?

sffc:
TG2 discussion: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-06-11.md#normative-synchronize-regionpreference-and-canonicalunicodesubdivision-with-uts-35-standards

The group thought that this problem is best solved in a proposal that fixes it not only here but throughout ECMA-402

anba:
JSC/V8 (directly via ICU4C) and SM already implement this PR.

PR #1062, opened 2026-05-11
Editorial: Simplify IsWellFormedUnitIdentifier

gibson042:


eemeli (approved):

sffc (approved):
Looks like you pulled from https://tc39.es/proposal-intl-sequence-units/ and simplified it further. LGTM

PR #1067, opened 2026-05-13
Return only available and canonical calendars in Locale.p.getCalendars

### LINKED ISSUE #1054: CalendarsOfLocale: Require returned calendars are canonicalized and a subset of AvailableCalendars
1. [CalendarsOfLocale](https://tc39.es/ecma402/#sec-calendarsoflocale) currently only requires "canonical form", but to match [Intl.supportedValuesOf](https://tc39.es/ecma402/#sec-intl.supportedvaluesof), the returned calendars should also be canonicalized through [CanonicalizeUValue](https://tc39.es/ecma402/#sec-canonicalizeuvalue).
2. And the returned calendars should be a subset of [AvailableCalendars](https://tc39.es/ecma402/#sec-availablecalendars).

ptomato:
Ensures returned calendars are actually supported by the implementation and in canonical form according to CanonicalizeUValue, not just the prose description of canonical form.

Closes: #1054

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -630,7 +630,13 @@
           1. Let _lookupRegion_ be _regionOverride_.
         1. Else,
           1. Let _lookupRegion_ be _region_.
-        1. Let _list_ be a List of unique calendar types in canonical form (<emu-xref href="#sec-calendar-types"></emu-xref>), sorted in descending preference of those in common use for date and time formatting in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
+        1. Let _list_ be a new empty List.
+        1. Let _availableCalendars_ be AvailableCalendars().
+        1. Let _calendarsInUse_ be a List of unique calendar types in common use for date and time formatting in _lookupRegion_, sorted by descending preference of use in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
+        1. For each element _identifier_ of _calendarsInUse_, do
+          1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
+          1. If _identifier_ is _canonical_ and _availableCalendars_ contains _identifier_, then
+            1. Append _identifier_ to _list_.
         1. If _list_ is empty, set _list_ to « *"gregory"* ».
         1. Return CreateArrayFromList(_list_).
       </emu-alg>


spec/locale.html
```diff
@@ -630,7 +630,14 @@ <h1>
...
           1. Let _lookupRegion_ be _region_.
-        1. Let _list_ be a List of unique calendar types in canonical form (<emu-xref href="#sec-calendar-types"></emu-xref>), sorted in descending preference of those in common use for date and time formatting in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
+        1. Let _list_ be a new empty List.
+        1. Let _availableCalendars_ be AvailableCalendars().
+        1. Let _calendarsInUse_ be a List of unique calendar types in common use for date and time formatting in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
+        1. For each element _identifier_ of _calendarsInUse_, do
+          1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
+          1. If _identifier_ is _canonical_ and _availableCalendars_ contains _identifier_, then
+            1. Append _identifier_ to _list_.
+        1. Sort _list_ by descending preference of use in _lookupRegion_.
```
gibson042: I have a mild preference for defining the iterated List to already be sorted, as in AvailableCalendars.
```suggestion
        1. Let _calendarsInUse_ be a List of unique calendar types in common use for date and time formatting in _lookupRegion_, sorted by descending preference of use in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
        1. For each element _identifier_ of _calendarsInUse_, do
          1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
          1. If _identifier_ is _canonical_ and _availableCalendars_ contains _identifier_, then
            1. Append _identifier_ to _list_.
```

gibson042 (approved):
Thanks!

spec/locale.html
ptomato: Ah, you are right. I thought of doing it this way at first, but figured it needed to be sorted after the canonicalization step. But then I rewrote the canonicalization step to be only a filter and not a map, so we can indeed sort beforehand.

gibson042 (approved):

PR #1072, opened 2026-05-29
Normative: Don't fall back to default locale in Locale.p.getCollations()

### LINKED ISSUE #1053: CollationsOfLocale: Reconsider calling DefaultLocale
[CollationsOfLocale](https://tc39.es/ecma402/#sec-collationsoflocale) calls `DefaultLocale`, because that's what ICU4C returns when calling `ucol_getKeywordValuesForLocale("collation", ...)`. This was noted in:
- https://github.com/tc39/proposal-intl-locale-info/issues/76#issuecomment-1686746537
- https://github.com/tc39/proposal-intl-locale-info/pull/92

For example in current implementations:

V8:
```sh
$ LANG=en ./d8 -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor
$ LANG=de ./d8 -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor,phonebk
$ LANG=sv ./d8 -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor,trad
$ LANG=zh ./d8 -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor,pinyin,stroke,zhuyin
$ LANG=ko ./d8 -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor,searchjl,unihan
```

JSC:
```sh
~$ LANG=en ./jsc -e 'print(new Intl.Locale("abcdefgh").getCollations())'
emoji,eor
~$ LANG=de ./jsc -e 'print(new Intl.Locale("abcdefgh").getCollations())'
phonebk,emoji,eor
~$ LANG=sv ./jsc -e 'print(new Intl.Locale("abcdefgh").getCollations())'
trad,emoji,eor
~$ LANG=zh ./jsc -e 'print(new Intl.Locale("abcdefgh").getCollations())'
pinyin,stroke,unihan,zhuyin,emoji,eor
~$ LANG=ko ./jsc -e 'print(new Intl.Locale("abcdefgh").getCollations())'
searchjl,unihan,emoji,eor
```

While it's possible to mirror the ICU4C behaviour when implementing this in ICU4X, maybe the spec should be changed to not call `DefaultLocale`? 

It's easy for ICU4C-based implementations to detect when the fallback to the default locale was used (`ucol_getKeywordValuesForLocale` returns the status code `U_USING_DEFAULT_WARNING`), so no large code changes will be needed. 

Proposed spec:
> 1. If loc.[[Collation]] is not undefined, then
>     1. Return CreateArrayFromList(« loc.[[Collation]] »).
> 1. Let language be GetLocaleLanguage(loc.[[Locale]]).
> 1. Let r be LookupMatchingLocaleByPrefix(%Intl.Collator%.[[AvailableLocales]], « loc.[[Locale]] »).
> 1. If r is not undefined, then
>     1. Let foundLocale be r.[[locale]].
>     1. Let foundLocaleData be %Intl.Collator%.[[SortLocaleData]].[[<foundLocale>]].
>     1. Let list be a copy of foundLocaleData.[[co]].
>     1. Assert: list[0] is null.
>     1. Remove the first element from list.
> 1. Else,
>     1. Let list be « "emoji", "eor" ».
> 1. Let sorted be a copy of list, sorted according to lexicographic code unit order.
> 1. Return CreateArrayFromList(sorted).

ptomato:
This is Anba's proposed change from the issue. It would change the handling of locales with undefined languages such as `und-u-co-emoji` and also remove the fallback to the default locale (so `LANG=` environment variables would have no effect.)

Closes: #1053

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -662,10 +662,9 @@
       <emu-alg>
         1. If _loc_.[[Collation]] is not *undefined*, then
           1. Return CreateArrayFromList(« _loc_.[[Collation]] »).
-        1. Let _language_ be GetLocaleLanguage(_loc_.[[Locale]]).
-        1. If _language_ is not *"und"*, then
-          1. Let _match_ be LookupMatchingLocaleByPrefix(%Intl.Collator%.[[AvailableLocales]], « _loc_.[[Locale]] »).
-          1. If _match_ is not *undefined*, let _foundLocale_ be _match_.[[locale]]; else let _foundLocale_ be DefaultLocale().
+        1. Let _match_ be LookupMatchingLocaleByPrefix(%Intl.Collator%.[[AvailableLocales]], « _loc_.[[Locale]] »).
+        1. If _match_ is not *undefined*, then
+          1. Let _foundLocale_ be _match_.[[locale]].
           1. Let _foundLocaleData_ be %Intl.Collator%.[[SortLocaleData]].[[&lt;_foundLocale_&gt;]].
           1. Let _list_ be a copy of _foundLocaleData_.[[co]].
           1. Assert: _list_[0] is *null*.


sffc (approved):
This makes it more consistent with other ops like the one for numbering system. Thanks!

FrankYFTang:
How this PR should be test. Could someone illustrate the method of testing by drafting a test262 PR?

gibson042 (approved):

anba:
> Could someone illustrate the method of testing by drafting a test262 PR?

This is not testable in test262, because it requires setting the default locale. See <https://github.com/tc39/ecma402/issues/1053#issue-4252013181>.

FrankYFTang:
> > Could someone illustrate the method of testing by drafting a test262 PR?
> 
> This is not testable in test262, because it requires setting the default locale. See [#1053 (comment)](https://github.com/tc39/ecma402/issues/1053#issue-4252013181).

if so, then how is this change meaningful?

sffc:
Suppose your browser locale is `zh`.

```javascript
new Intl.Locale("abcdefgh").getCollations()

// Old behavior: emoji,eor,pinyin,stroke,zhuyin
// New behavior: emoji,eor
```

The result should not depend on the browser locale.

ptomato:
I'd have to page the context back in as it's been a few weeks since I looked at this, but I believe there is a change to `und` locales such as `und-u-co-emoji` that would be testable in test262. I'll follow up here.

sffc:
TG2 agreement: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-06-11.md#normative-avoid-default-locale-fallback-within-localepgetcollations

ptomato:
Rebased.

ptomato:
Tests in https://github.com/tc39/test262/pull/5092.

Note that it is indeed not possible to test the change directly. I've added a test that may or may not fail if the change is not implemented, depending on the default locale, but should always pass after the change regardless of the default locale. Testing it locally on d8, for example, it fails with `LC_ALL=de` in the environment.

gibson042 (approved):

PR #1074, opened 2026-06-03
Normative: Allow Locale.p.getNumberingSystems to return >1 item

### LINKED ISSUE #1071: Intl.Locale.prototype.getNumberingSystems() does not return useful info
On Chrome (haven't checked other browsers yet):

```javascript
new Intl.Locale("hi").getNumberingSystems()
// ['latn']
new Intl.Locale("ar").getNumberingSystems()
// ['latn']
new Intl.Locale("bn").getNumberingSystems()
// ['beng']
new Intl.Locale("ar-EG").getNumberingSystems()
// ['arab']
```

This is just wrong. All of those should return multiple entries.

CLDR has this data:

```xml
<!-- hi.xml -->
		<defaultNumberingSystem>↑↑↑</defaultNumberingSystem>
		<otherNumberingSystems>
			<native>deva</native>
		</otherNumberingSystems>
```

https://www.unicode.org/reports/tr35/tr35-numbers.html#defaultNumberingSystem

Test262 issue: https://github.com/tc39/test262/issues/5062

CC @FrankYFTang @anba

ptomato:
This seems to be a regression from https://github.com/tc39/proposal-intl-locale-info/pull/92 that went unnoticed, accidentally removing the ability to return more than one numbering system from Intl.Locale.prototype.getNumberingSystems.

The sorting criterion added here matches the one that was removed in that change.

Closes: #1071

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -720,7 +720,7 @@
           1. Let _foundLocale_ be _match_.[[locale]].
           1. Let _foundLocaleData_ be %Intl.NumberFormat%.[[LocaleData]].[[&lt;_foundLocale_&gt;]].
           1. Let _numberingSystems_ be _foundLocaleData_.[[nu]].
-          1. Let _list_ be « _numberingSystems_[0] ».
+          1. Let _list_ be a copy of _numberingSystems_, sorted in descending preference of those in common use for formatting numeric values in _foundLocale_.
         1. Else,
           1. Let _list_ be « *"latn"* ».
         1. Return CreateArrayFromList(_list_).
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -713,7 +713,7 @@
           1. Let _foundLocale_ be _r_.[[locale]].
           1. Let _foundLocaleData_ be %Intl.NumberFormat%.[[LocaleData]].[[&lt;_foundLocale_&gt;]].
           1. Let _numberingSystems_ be _foundLocaleData_.[[nu]].
-          1. Let _list_ be « _numberingSystems_[0] ».
+          1. Let _list_ be a copy of _numberingSystems_, sorted in descending preference of those in common use for formatting numeric values in _foundLocale_.
         1. Else,
           1. Let _list_ be « *"latn"* ».
         1. Return CreateArrayFromList(_list_).
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -719,8 +719,13 @@
         1. If _match_ is not *undefined*, then
           1. Let _foundLocale_ be _match_.[[locale]].
           1. Let _foundLocaleData_ be %Intl.NumberFormat%.[[LocaleData]].[[&lt;_foundLocale_&gt;]].
-          1. Let _numberingSystems_ be _foundLocaleData_.[[nu]].
-          1. Let _list_ be a copy of _numberingSystems_, sorted in descending preference of those in common use for formatting numeric values in _foundLocale_.
+          1. Let _supportedNumberingSystems_ be _foundLocaleData_.[[nu]].
+          1. Let _numberingSystemsInUse_ be a List of unique numbering system identifiers in common use for formatting numeric values in locale _foundLocale_, sorted by descending preference of use in _foundLocale_.
+          1. Let _list_ be a new empty List.
+          1. For each element _identifier_ of _numberingSystemsInUse_, do
+            1. If _supportedNumberingSystems_ contains _identifier_, then
+              1. Append _identifier_ to _list_.
+          1. Assert: _list_[0] is _supportedNumberingSystems_[0].
         1. Else,
           1. Let _list_ be « *"latn"* ».
         1. Return CreateArrayFromList(_list_).
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -725,7 +725,7 @@
           1. For each element _identifier_ of _numberingSystemsInUse_, do
             1. If _supportedNumberingSystems_ contains _identifier_, then
               1. Append _identifier_ to _list_.
-          1. Assert: _list_[0] is _supportedNumberingSystems_[0].
+          1. Assert: _list_ is not empty and _list_[0] is _supportedNumberingSystems_[0].
         1. Else,
           1. Let _list_ be « *"latn"* ».
         1. Return CreateArrayFromList(_list_).


ptomato:
Tests in https://github.com/tc39/test262/pull/5067

FrankYFTang (approved):

sffc (approved):

gibson042 (approved):

anba:
`foundLocaleData.[[nu]]` returns the numbering systems supported for a specific locale, but not necessarily also in common use. I think the list should be restricted to those in common use, not just a copy sorted according to common use. 

For example `new Intl.Locale("en").getNumberingSystems().includes("mlym")` should return `false`, because English doesn't use Malayalam digits.

sffc:
Good point. Perhaps it should be allowed to return a subset of the full list.

https://tc39.es/ecma402/#sec-intl.numberformat-internal-slots

sffc:
TG2 agreement: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-06-11.md#normative-allow-localepgetnumberingsystems-to-return-1-item

ptomato:
@anba Where are you finding the data for numbering systems _not_ in common use? I don't see anywhere in CLDR that `"en"` supports Malayalam digits. CLDR only exposes "default", "native", and optional "traditional" and "finance" numbering systems for a language ID. For that matter the optional ones are a bit ambiguous how to sort them.

anba:
`NumberingSystemsOfLocale` uses `%Intl.NumberFormat%.[[LocaleData]]` because that allows to get the default numbering system for a locale. The first element of `foundLocaleData.[[nu]]` is guaranteed to contain the default numbering system per <https://tc39.es/ecma402/#sec-internal-slots>:
> [...] The value of each such locale-named field is a [Record](https://tc39.es/ecma262/#sec-list-and-record-specification-type) in which each element of [[RelevantExtensionKeys]] identifies the name of a field whose value is a non-empty [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type) of Strings representing the type values that are supported by the implementation in the relevant locale for the corresponding [Unicode locale extension sequence](https://tc39.es/ecma402/#unicode-locale-extension-sequence) key, **with the first element providing the default value for that key in the locale**.

If not just the default numbering system should be returned, then a similar wording as in [CalendarsOfLocale](https://tc39.es/ecma402/#sec-calendarsoflocale) can be used. "native", "traditio", or "finance" should be explicitly excluded, because these aren't allowed for `Intl.NumberFormat` per <https://tc39.es/ecma402/#sec-intl.numberformat-internal-slots>.

If the numbering systems should be returned which are in common use and additionally also supported by `Intl.NumberFormat`, then the intersection of `%Intl.NumberFormat%.[[LocaleData]].[[<foundLocale>]].[[nu]]` and the numbering systems in common use needs to be computed: 

> 1. Let numberingSystems be foundLocaleData.[[nu]].
> 1. Let numberingSystemsInUse be a List of unique numbering system identifiers in common use for number formatting in locale `foundLocale`, sorted by descending preference of use in `foundLocale`.
> 1. Let list be a new empty List.
> 1. For each element identifier of numberingSystemsInUse, do
>     1. If numberingSystems contains identifier, then
>         1. Append identifier to list.
> 1. Assert: list[0] is numberingSystems[0].

The assertion `list[0] is numberingSystems[0]` must hold because `%Intl.NumberFormat%.[[LocaleData]].[[<foundLocale>]].[[nu]]` contains the default numbering system, which is the numbering system most commonly used. 

Returning all elements of `%Intl.NumberFormat%.[[LocaleData]].[[<foundLocale>]].[[nu]]` is incorrect, because implementations (or at least ICU4C-based implementations) support formatting in numbering systems which aren't in common use. For example `new Intl.NumberFormat("en-u-nu-mlym").format(123)` works in all browsers and returns "൧൨൩", even though Malayalam digits aren't in common use for English. 

---

Also: ICU4C doesn't provide any public API to get all numbering systems in common use. There's only [unumsys_open](https://searchfox.org/firefox-main/rev/91d6cd2d1476bca635bb96f82ca34eda283b5460/intl/icu/source/i18n/unicode/unumsys.h#49-61) which returns the default numbering system. I haven't yet checked if it's possible to retrieve other numbering systems through the UResourceBundle API, similar to how Firefox [supports](https://searchfox.org/firefox-main/rev/91d6cd2d1476bca635bb96f82ca34eda283b5460/intl/components/src/DateTimeFormat.cpp#673-793) returning all hour-cycles for a locale, so that `new Intl.Locale("en-CA").getHourCycles()` returns `[ "h12", "h23" ]` instead of just `[ "h12" ]`.

ptomato:
I'm talking about data like https://github.com/unicode-org/cldr/blob/2ae632be7a630983fa454e4202ab8ca388e304f8/common/main/ar.xml#L5875-L5879

The structure of the data in CLDR is generally
```xml
<defaultNumberingSystem>XXXX</defaultNumberingSystem>
<otherNumberingSystems>
  <native>XXXX</native>
  <traditional>XXXX</traditional>
  <finance>XXXX</finance>
</otherNumberingSystems>
```
(specified in https://www.unicode.org/reports/tr35/tr35-numbers.html#Number_Elements)

So my question still stands, but let me ask it differently: is there anything in CLDR that specifically says `new Intl.NumberFormat("en-u-nu-mlym").format(123)` is supported? If so, where?

If yes, then we indeed have a problem with the text in this PR and we need to adjust it to distinguish between "supported" and "in common use" systems.

If no, then I'm assuming that all combinations of `new Intl.NumberFormat("LOCALE-u-nu-NUMBERINGSYSTEM").format(123)` will produce "123" in the given numbering system. In that case, I think this change is correct, but additionally it's probably not sensible to have the spec talk about "supported" numbering systems for %Intl.NumberFormat%.\[\[LocaleData]].\[\[...]].\[\[nu]], because evidently that field is always the list of all known numbering systems.

anba:
- `new Intl.Locale(...).getNumberingSystems()` must only return the numbering systems which are in common use (according to CLDR data).
- `Intl.NumberFormat` is allowed to support additional numbering systems, even if they aren't in common use. 

We don't want `new Intl.Locale(...).getNumberingSystems()` to return all numbering systems which are supported by `Intl.NumberFormat` for a specific locale. This a strict requirement. 

---

ICU4C-based implementations support all [numbering systems with simple digit mappings](https://tc39.es/ecma402/#table-numbering-system-digits) for any locale. IIRC no browser currently supports algorithmic numbering systems, cf. <https://github.com/unicode-org/cldr/blob/main/common/bcp47/number.xml>. But ECMA-402 allows that implementations support for example `new Intl.NumberFormat("en-u-nu-roman").format(123)` to return `"CXXIII"`.

ptomato:
I've updated the language based on Anba's suggested text, to treat "supported numbering systems" and "numbering systems in common use" as two different sets, of which potentially the only shared element is the default numbering system.

I've kept it in a separate commit for now so that it's clear what changed relative to the text that previously went through the consensus process. I'm not sure if this needs to go through the consensus process again.

spec/locale.html
```diff
@@ -719,8 +719,13 @@ <h1>
...
           1. Let _foundLocaleData_ be %Intl.NumberFormat%.[[LocaleData]].[[&lt;_foundLocale_&gt;]].
-          1. Let _numberingSystems_ be _foundLocaleData_.[[nu]].
-          1. Let _list_ be « _numberingSystems_[0] ».
+          1. Let _supportedNumberingSystems_ be _foundLocaleData_.[[nu]].
+          1. Let _numberingSystemsInUse_ be a List of unique numbering system identifiers in common use for formatting numeric values in locale _foundLocale_, sorted by descending preference of use in _foundLocale_.
+          1. Let _list_ be a new empty List.
+          1. For each element _identifier_ of _numberingSystemsInUse_, do
+            1. If _supportedNumberingSystems_ contains _identifier_, then
+              1. Append _identifier_ to _list_.
+          1. Assert: _list_[0] is _supportedNumberingSystems_[0].
```
gibson042:
```suggestion
          1. Assert: _list_ is not empty and _list_[0] is _supportedNumberingSystems_[0].
```

gibson042 (approved):

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-09-17.md#pr-1074-allow-localepgetnumberingsystems-to-return-1-item--consensus-on-updated-text-distinguishing-supported-systems-from-those-in-common-use

PR #1078, opened 2026-07-01
Editorial: Update Well-Known Intrinsic Objects table to match 262

linusg:
See:

- https://github.com/tc39/ecma262/pull/2056
- https://github.com/tc39/ecma262/pull/3881

This is what remains:

image

### DIFF
--- a/spec/conventions.html
+++ b/spec/conventions.html
@@ -35,58 +35,8 @@
         </thead>
         <tr>
           <td>%Intl%</td>
-          <td>`Intl`</td>
+          <td>*"Intl"*</td>
           <td>The `Intl` object (<emu-xref href="#intl-object"></emu-xref>)</td>
-        </tr>
-        <tr>
-          <td>%Intl.Collator%</td>
-          <td>`Intl.Collator`</td>
-          <td>The `Intl.Collator` constructor (<emu-xref href="#sec-intl-collator-constructor"></emu-xref>)</td>
-        </tr>
-        <tr>
-          <td>%Intl.DateTimeFormat%</td>
-          <td>`Intl.DateTimeFormat`</td>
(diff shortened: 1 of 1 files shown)


spec/conventions.html
```diff
@@ -35,58 +35,58 @@ <h1>Well-Known Intrinsic Objects</h1>
...
           <td>%Intl%</td>
-          <td>`Intl`</td>
+          <td>*"Intl"*</td>
           <td>The `Intl` object (<emu-xref href="#intl-object"></emu-xref>)</td>
         </tr>
         <tr>
           <td>%Intl.Collator%</td>
-          <td>`Intl.Collator`</td>
-          <td>The `Intl.Collator` constructor (<emu-xref href="#sec-intl-collator-constructor"></emu-xref>)</td>
+          <td>*"Intl.Collator"*</td>
```
gibson042: I don't think stings like "Intl.Collator" make sense in a “Global Name” column; the `name` property of an Intl constructor isn't prefixed with "Intl.". I think I prefer just removing the dotted-name rows (as was done for ECMA-262 in https://github.com/tc39/ecma262/pull/2056), but could alternatively just clear these cells. Any thoughts, @ryzokuken?

gibson042 (commented):
The “Global Name” column isn't right, but there are two approaches for fixing it and we need to pick one.

spec/conventions.html
linusg: Temporal only adds the namespace to the table (https://tc39.es/proposal-temporal/#table-well-known-intrinsic-objects) so removing WFM.

sffc:
TG2 discussion and approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-08-06.md#pr-1078-update-well-known-intrinsic-objects-table-to-match-262--choice-between-two-approaches-for-fixing-the-global-name-column

PR #1083, opened 2026-07-04
Editorial: Make the GetLocaleVariants algorithm more explicit

### LINKED ISSUE #1082: GetLocaleVariants has confusing wording about how to extract the string of variants
[GetLocaleVariants](https://tc39.es/ecma402/#sec-getlocalevariants) step 3:

> 3. Let _variants_ be the longest suffix of _baseName_ that starts with a **"-"** followed by a substring that is matched by the `unicode_variant_subtag` [Unicode locale nonterminal](https://tc39.es/ecma402/#sec-language-tags). If there is no such suffix, return **undefined**.

This sentence is grammatically ambiguous as to whether it means

- the longest suffix of _baseName_
  - that starts with **"-"**
  - (and is) followed by a substring matched by `unicode_variant_subtag`

(this interpretation makes no sense because a suffix can't be followed by anything, but it's how my brain parses the sentence)

or 

- the longest suffix of _baseName_
  - that starts with
    - **"-"** followed by a substring matched by `unicode_variant_subtag`

(the correct interpretation)

But even the correct interpretation is confusing to me because it sounds like it's referring to either the longest variant subtag (if you overlook the word "suffix") or the last variant subtag (if you overlook the word "longest"). Based on implementation behaviour I'm assuming what we want here is _all_ of the variant subtags, preceded and separated by `-`.

How about something like this?

3. Return the longest suffix of _baseName_ that consists entirely of substrings that are matched by the `unicode_variant_subtag` Unicode locale nonterminal, separated by **"-"**. If there is no such suffix, return **undefined**.

or even written out more algorithmically?

3. Let _subtags_ be StringSplitToList(_baseName_, **"-"**).
4. Repeat, until _subtags_[0] is matched by the `unicode_variant_subtag` Unicode locale nonterminal,
  - 1. Remove the first element of _subtags_.
5. If _subtags_ is empty, return **undefined**.
6. Let _variants_ be the empty String.
7. Let _k_ be 0.
8. Let _length_ be the number of elements in _subtags_.
9. Repeat, while _k_ < _length_,
  - 1. If _k_ > 0, set _variants_ to the string-concatenation of _variants_ and **"-"**.
    2. Set _variants_ to the string-concatenation of _variants_ and _subtags_[_k_].
    3. Set _k_ to _k_ + 1.
10. Return _variants_.

gibson042:
Fixes #1082

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -543,8 +543,15 @@
       <emu-alg>
         1. Let _baseName_ be GetLocaleBaseName(_locale_).
         1. NOTE: Each subtag in _baseName_ that is preceded by *"-"* is either a <code>unicode_script_subtag</code>, <code>unicode_region_subtag</code>, or <code>unicode_variant_subtag</code>, but any <emu-not-ref>substring</emu-not-ref> matched by <code>unicode_variant_subtag</code> is strictly longer than any prefix thereof which could also be matched by one of the other productions.
-        1. Let _variants_ be the longest suffix of _baseName_ that starts with a *"-"* followed by a <emu-not-ref>substring</emu-not-ref> that is matched by the <code>unicode_variant_subtag</code> Unicode locale nonterminal. If there is no such suffix, return *undefined*.
-        1. Return the substring of _variants_ from 1.
+        1. Let _subtags_ be StringSplitToList(_baseName_, *"-"*).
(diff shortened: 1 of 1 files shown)


ptomato (approved):
I like it!

PR #1085, opened 2026-07-06
Undo unintentional normative change in CalendarsOfLocale

### LINKED ISSUE #1084: Incorrect data returned from Intl.Locale.p.getCalendars when return value should include "gregory"
CLDR's calendar preference data uses the `gregorian` alias instead of the canonical `gregory` calendar ID:

https://github.com/unicode-org/cldr/blob/5709b4b308911736b8cdcb995aa3e258a492dc5f/common/supplemental/supplementalData.xml#L4492-L4508

Step 10 of [CalendarsOfLocale](https://tc39.es/ecma402/#sec-calendarsoflocale) says all non-canonical aliases should just be dropped, which is incorrect. This would lead to the following observable results:
```js
new Intl.Locale("ja").getCalendars()
// => ["japanese"]
```
whereas current web reality is to return `["gregory", "japanese"]`.

(This is actually my fault; I broke this in #1067. I think we can address this without a normative change, since the original change wasn't intended to be observable either, fixing this just puts the behaviour back the way it should have been.)

ptomato:
Regression from #1067. Intl.Locale.p.getCalendars should return only canonical and available calendar IDs, but should not drop non-canonical IDs.

Closes: #1084

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -635,8 +635,8 @@
         1. Let _calendarsInUse_ be a List of unique calendar types in common use for date and time formatting in _lookupRegion_, sorted by descending preference of use in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
         1. For each element _identifier_ of _calendarsInUse_, do
           1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
-          1. If _identifier_ is _canonical_ and _availableCalendars_ contains _identifier_, then
-            1. Append _identifier_ to _list_.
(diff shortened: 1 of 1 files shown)


gibson042 (approved):

ryzokuken (approved):
Do we need to treat this as a new normative change or given that it's a revert we can handle it differently?

sffc (approved):

sffc:
TG2 discussed this and agreed that you don't need consensus since you are changing back a previous unintentional change: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-07-09.md#pr-1085-undo-unintentional-normative-change-in-calendarsoflocale--consent-to-merge-approved-regression-fix

PR #1051, opened 2026-04-09
Normative: Call RegionPreference in WeekInfoOfLocale

anba:
`rg` and `sd` Unicode extension keys are valid for `WeekInfoOfLocale`.

This matches existing ICU4C-based implementations.

PR #1086, opened 2026-07-07
Normative: Take language subtag into account in locale hour cycle and calendar lookup

### LINKED ISSUE #1056: HourCyclesOfLocale: Add language subtag to lookup
[HourCyclesOfLocale](https://tc39.es/ecma402/#sec-hourcyclesoflocale) performs a lookup based only on the region subtag, but CLDR's [Time Data](https://unicode.org/reports/tr35/tr35-dates.html#Time_Data) accepts locale identifiers in addition to standalone region subtags.

For example `new Intl.Locale("en-CA").getHourCycles()[0]` should return `"h12"`, whereas `new Intl.Locale("fr-CA").getHourCycles()[0]` should return `"h23"`, but the current spec text doesn't allow this result.

ptomato:
Some of CLDR's hour cycle preference data is indexed by language and region, not just by region.

For example, Francophone Canada is listed as preferring a 24-hour clock while Anglophone Canada is listed as preferring a 12-hour clock. As of the current version of the spec text, it is not possible to return a value that takes these preferences into account, because only the region (Canada) is used for looking up the hour cycle preference.

This observably affects code such as

    new Intl.Locale("fr-CA").getHourCycles()

which, after this change, will return an array where element 0 is "h23", instead of "h12".

Closes: #1056

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -631,15 +631,17 @@
         1. If _loc_.[[Calendar]] is not *undefined*, then
           1. Return CreateArrayFromList(« _loc_.[[Calendar]] »).
         1. Let _preference_ be RegionPreference(_loc_.[[Locale]]).
-        1. Let _region_ be _preference_.[[Region]].
-        1. Let _regionOverride_ be _preference_.[[RegionOverride]].
-        1. If _regionOverride_ is not *undefined* and calendar preference data for _regionOverride_ are available, then
-          1. Let _lookupRegion_ be _regionOverride_.
-        1. Else,
-          1. Let _lookupRegion_ be _region_.
+        1. If _preference_.[[RegionOverride]] is not *undefined*, let _preferredRegions_ be « _preference_.[[RegionOverride]], _preference_.[[Region]] »; else let _preferredRegions_ be « _preference_.[[Region]] ».
+        1. Let _calendarsInUse_ be a new empty List.
+        1. Let _language_ be GetLocaleLanguage(_loc_.[[Locale]]).
+        1. For each String _region_ of _preferredRegions_, do
+          1. Let _locale_ be the string-concatenation of _language_, *"-"*, and _region_.
+          1. If _calendarsInUse_ is empty and calendar preference data for locale _locale_ are available, then
+            1. Set _calendarsInUse_ to a List of unique calendar types in common use for date and time formatting in locale _locale_, sorted by descending preference of use in _locale_.
+          1. If _calendarsInUse_ is empty and calendar preference data for region _region_ are available, then
+            1. Set _calendarsInUse_ to a List of unique calendar types in common use for date and time formatting in region _region_, sorted by descending preference of use in _region_.
         1. Let _list_ be a new empty List.
         1. Let _availableCalendars_ be AvailableCalendars().
-        1. Let _calendarsInUse_ be a List of unique calendar types in common use for date and time formatting in _lookupRegion_, sorted by descending preference of use in _lookupRegion_. The list is empty if no calendar preference data for _lookupRegion_ is available.
         1. For each element _identifier_ of _calendarsInUse_, do
           1. Let _canonical_ be CanonicalizeUValue(*"ca"*, _identifier_).
           1. If _availableCalendars_ contains _canonical_ and _list_ does not contain _canonical_, then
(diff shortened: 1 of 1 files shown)


ptomato:
Open question is whether we should rewrite the other preference lookup operations to work like this. As far as I can tell, hour cycles is the only data source where it matters. (Aside from `<weekOfPreference>` data, which we don't expose in JS.)

spec/locale.html
```diff
@@ -684,11 +684,17 @@ <h1>
...
+        1. Let _locale_ be the string-concatenation of _language_, *"-"*, and _region_.
+        1. Let _lookup_ be _region_.
+        1. If time data for _locale_ are available, set _lookup_ to _locale_.
+        1. If _regionOverride_ is not *undefined*, then
+          1. Let _localeOverride_ be the string-concatentation of _language_, *"-"*, and _regionOverride_.
+          1. If time data for _localeOverride_ are available, then
+            1. Set _lookup_ to _localeOverride_.
+          1. Else if time data for _regionOverride_ are available, then
+            1. Set _lookup_ to _regionOverride_.
+        1. Let _list_ be a List of unique hour cycle identifiers, which must be lower case String values indicating either the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*), sorted in descending preference of those in common use for date and time formatting in _lookup_. The list is empty if no time data for _lookup_ are available.
```
gibson042: I find this conflation of locale identifier and region for time data lookup to be a bit too handwavy. Suggestion:
```suggestion
        1. If _preference_.[[RegionOverride]] is not *undefined*, let _preferredRegions_ be « _preference_.[[RegionOverride]], _preference_.[[Region]] »; else let _preferredRegions_ be « _preference_.[[Region]] ».
        1. Let _hourCycles_ be a new empty List.
        1. Let _language_ be GetLocaleLanguage(_loc_.[[Locale]]).
        1. For each String _region_ of _preferredRegions_, do
          1. Let _locale_ be the string-concatenation of _language_, *"-"*, and _region_.
          1. If _hourCycles_ is empty and time data for locale _locale_ are available, then
            1. Set _hourCycles_ to a List of unique hour cycle identifiers, which must be lower case String values indicating either the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*), sorted in descending preference of those in common use for date and time formatting in locale _locale_.
          1. If _hourCycles_ is empty and time data for region _region_ are available, then
            1. Set _hourCycles_ to a List of unique hour cycle identifiers, which must be lower case String values indicating either the 12-hour format (*"h11"*, *"h12"*) or the 24-hour format (*"h23"*, *"h24"*), sorted in descending preference of those in common use for date and time formatting in region _region_.
        1. If _hourCycles_ is empty, set _hourCycles_ to « *"h23"* ».
        1. Return CreateArrayFromList(_hourCycles_).
```

gibson042 (commented):
> Open question is whether we should rewrite the other preference lookup operations to work like this. As far as I can tell, hour cycles is the only data source where it matters. (Aside from `<weekOfPreference>` data, which we don't expose in JS.)

It definitely makes sense for CalendarsOfLocale, which also already uses RegionPreference and is specifically covered by [rgScope.xml](https://github.com/unicode-org/cldr/blob/main/common/supplemental/rgScope.xml), and probably is not appropriate for TimeZonesOfLocale.

I'd like to just _completely_ defer to `<rgScope>` per [UTS #35 Part 6 Supplemental](https://unicode.org/reports/tr35/tr35-info.html#rgScope), but that'll be a bigger effort.

ptomato:
I've made the same change for CalendarsOfLocale, although there are not currently any language-region entries in CLDR's calendar preference data.

Note interactions with other open PRs:

- #1085 will merge-conflict with this.
- #1051 will also call RegionPreference in WeekInfoOfLocale.

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-07-09.md#pr-1086-take-language-subtag-into-account-in-locale-hour-cycle-lookup--scope-expansion-of-the-two-tiered-lookup-algorithm-to-include-calendarsoflocale

ptomato:
Tests in https://github.com/tc39/test262/pull/5091 (which V8 and JSC already pass)

gibson042 (approved):

PR #1087, opened 2026-07-07
Editorial: Make the Locale object abstract operation algorithms more explicit

gibson042:
Ref #1083

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -492,8 +492,9 @@
       </dl>
       <emu-alg>
         1. Let _baseName_ be GetLocaleBaseName(_locale_).
-        1. Assert: The first subtag of _baseName_ can be matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal.
-        1. Return the first subtag of _baseName_.
+        1. Let _subtags_ be StringSplitToList(_baseName_, *"-"*).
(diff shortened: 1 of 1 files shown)


spec/locale.html
```diff
@@ -492,8 +492,9 @@ <h1>
       </dl>
       <emu-alg>
         1. Let _baseName_ be GetLocaleBaseName(_locale_).
-        1. Assert: The first subtag of _baseName_ can be matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal.
-        1. Return the first subtag of _baseName_.
+        1. Let _subtags_ be StringSplitToList(_baseName_, *"-"*).
+        1. Assert: _subtags_[0] is matched by the <code>unicode_language_subtag</code> Unicode locale nonterminal.
```
ptomato, line 496: Note, "_subtags_ has at least one element" is implicit here and in similar language below, but I think that's OK.

ptomato (approved):
This is an improvement IMO.

spec/locale.html
gibson042: That's actually guaranteed by the final three steps of [StringSplitToList](https://tc39.es/ecma402/#sec-string-split-to-list).

PR #1093, opened 2026-08-07
Use language preferences in determining locales

### LINKED ISSUE #1010: Allow engines to ship data for locales in Accept-Language
As a follow-up to https://github.com/tc39/ecma402/issues/588 and https://github.com/tc39/ecma402/pull/780, we are now specifying that the available locales ought to be derived from the User Agent, and we left some room to explore how we can support additional locales.

I've floated the following idea, and it seems like it may be workable: allow the available locales list to derive not only from User-Agent but also from Accept-Language aka navigator.languages.

In order to avoid the fingerprint, it should be the case that, given any User-Agent/Accept-Language combo, the set of available locales is always the same. So, any particular User-Agent should either _always_ or _never_ change the set of available locales based on the Accept-Language, and it should always do so in the same way.

I've gotten questions about whether this approach is impacted by efforts to reduce the fingerprintability of Accept-Language. I think it is not, because we are allowing engines to add additional locale data based on the Accept-Language, and if the Accept-Language is reduced, then the engine is reduced in the amount of data that it can add. So, these are two disjoint problems.

jessealama:
As discussed at the 402 monthly call on 2026-08-06, allow the set of initial locales in browser contexts to be determined by (1) the user agent as well as (2) language preferences (in other words, `navigator.languages`).

closes #1010


eemeli:
This introduces a normative reference on the HTML spec, which should be noted in the relevant section.

jessealama:
> This introduces a normative reference on the HTML spec, which should be noted in the relevant section.

The Overview section where we're adding the new reference is non-normative, so I think it's OK to cite HTML here. We don't (currently) have something like Informative References, so the only place to add it would be the Normative References section, which I think wouldn't be right given the whole Overview section is to be understood as non-normative.

There's also one other reference to HTML in 402, in `spec/locales-currencies-tz.html`, about `DefaultLocale`. It's pretty much the same idea that the overview talks about:

> The returned value is is a potential fingerprinting vector. In browser environments, it should match `navigator.language` to avoid providing any additional distinguishing information.

eemeli:
> The Overview section where we're adding the new reference is non-normative, so I think it's OK to cite HTML here.

Huh, didn't notice that. Given that the relevant paragraph already now includes this:
> Furthermore, dynamic changes to these sets must not result in users becoming distinguishable from each other. 

we maybe should restructure things so that this does become normative. While I am aware that we don't use RFC 2119 keywords, an expression like "must not" does sound awfully normative. And it should be.

sffc (approved):

PR #1095, opened 2026-08-23
Normative: Define %TypedArray%.prototype.toLocaleString

### LINKED ISSUE #1094: ECMA-402 should define overriding of %TypedArray%.prototype.toLocaleString
[%TypedArray%.prototype.toLocaleString](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.prototype.tolocalestring) requires implementations that include ECMA-402 to override it analogously to the way they override Array.prototype.toLocaleString, but ECMA-402 makes no mention of that.

gibson042:
Fixes #1094
Depends upon JoinIndexedProperties and LocalizedListSeparator from https://github.com/tc39/ecma262/pull/3958


spec/locale-sensitive-functions.html
```diff
@@ -220,30 +220,34 @@ <h1>Array.prototype.toLocaleString ( [ _locales_ [ , _options_ ] ] )</h1>
...
-        1. Repeat, while _k_ &lt; _length_,
-          1. If _k_ > 0, then
-            1. Set _result_ to the string-concatenation of _result_ and _separator_.
-          1. Let _element_ be ? Get(_array_, ! ToString(𝔽(_k_))).
-          1. If _element_ is neither *undefined* nor *null*, then
-            1. Let _elementString_ be ? ToString(? Invoke(_element_, *"toLocaleString"*, « _locales_, _options_ »)).
-            1. Set _result_ to the string-concatenation of _result_ and _elementString_.
-          1. Set _k_ to _k_ + 1.
-        1. Return _result_.
+        1. Return ? JoinIndexedProperties(_array_, _length_, ~auto~, *"toLocaleString"*, « _locales_, _options_ »).
```
sffc: It seems like we should pass `_separator_` in here instead of relying on 262 to have the ILD behavior

sffc (approved):

spec/locale-sensitive-functions.html
gibson042: Updated to pass it explicitly, but using an operation defined in ECMA-262.

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-09-17.md#pr-1095-define-typedarrayprototypetolocalestring--forwarding-locales-and-options-arguments-to-elements-to-match-web-reality

PR #1096, opened 2026-08-31
Editorial: Generalize ResolveLocale to support pure BCP 47 lookup without localeData

gibson042:
There is a need to perform [BCP 47 lookup](https://www.rfc-editor.org/rfc/rfc4647.html#section-3) even in the absence of locale data, as noted in https://github.com/tc39/proposal-amount/issues/95 . This generalizes ResolveLocale to support that case by making the _localeData_ parameter optional (which avoids affecting any existing uses).

### DIFF
--- a/spec/negotiation.html
+++ b/spec/negotiation.html
@@ -222,7 +222,7 @@
           _requestedLocales_: a Language Priority List,
           _options_: a Record,
           _relevantExtensionKeys_: a List of Strings,
-          _localeData_: a Record,
+          optional _localeData_: a Record,
         ): a Record
       </h1>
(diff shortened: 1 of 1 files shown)


eemeli:
If this is only to help the Amount proposal, we should ensure there first that it is indeed heloful -- see https://github.com/tc39/proposal-amount/pull/110#discussion_r3895931758.

gibson042:
No, it is not only for the Amount proposal. I've thought for a while that ResolveLocale was too tightly bound to locale data, and would want this change even absent Amount.

eemeli (approved):

PR #1097, opened 2026-09-02
Normative: Determine getWeekInfo().firstDay using First Day Overrides

### LINKED ISSUE #1055: WeekInfoOfLocale: Handle "ca" Unicode extension
[WeekInfoOfLocale](https://tc39.es/ecma402/#sec-weekinfooflocale) should handle the `ca` Unicode extension per [First Day Overrides](https://unicode.org/reports/tr35/tr35-dates.html#First_Day_Overrides). 

See also #1051 which adds [RegionPreference](https://tc39.es/ecma402/#sec-regionpreference) to WeekInfoOfLocale. 

(There's currently only a note referencing UTS 35 "First Day Overrides", but notes aren't normative.)

ptomato:
UTS-35 specifies an algorithm, called "First Day Overrides" (https://unicode.org/reports/tr35/tr35-dates.html#First_Day_Overrides) that determines what subtags in a locale identifier should influence the determination of the first day of the week, and with what priority.

Apply the algorithm to Intl.Locale.prototype.getWeekInfo().firstDay. Before this change, the RegionPreference algorithm was applied, which ignored the `-u-ca` calendar value.

The RegionPreference algorithm continues to be applied to the determination of the weekend days, which are not affected by the choice of calendar.

Closes: #1055

### DIFF
--- a/spec/locale.html
+++ b/spec/locale.html
@@ -676,6 +676,36 @@
       </emu-alg>
     </emu-clause>
 
+    <emu-clause id="sec-firstdayoverrides" type="implementation-defined abstract operation">
+      <h1>
+        FirstDayOverrides (
+          _loc_: an Intl.Locale,
+        ): an integral Number in the inclusive interval from *1*<sub>𝔽</sub> to *7*<sub>𝔽</sub>
+      </h1>
+      <dl class="header">
+        <dt>description</dt>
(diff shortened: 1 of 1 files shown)


ptomato:
Tests are in https://github.com/tc39/test262/pull/5128.

Note the fallback priorities in that test; if you review this, it'd be good to review the test and see if you agree with the resulting priorities. You might be surprised by the influence of the options bag passed to the constructor. I think this result follows naturally from how the options are treated in the Intl.Locale constructor, but we could write this differently if we wanted a different result.

sffc (approved):

sffc:
TG2 approval: https://github.com/tc39/ecma402/blob/main/meetings/notes-2026-09-17.md#pr-1097-determine-getweekinfofirstday-using-first-day-overrides--fallback-priority-resolution-for-calendar-and-region-subtags

